PR #39, opened 2017-12-26
[DO NOT MERGE] {TensorDim->dimensions-0.3.2.0} in Tensor.Dynamic.Double and {Tensor.Raw->GenericOps typeclass}

### LINKED ISSUE #18: Convert shape types to use the dimensions package.
http://hackage.haskell.org/package/dimensions

stites:
Take a look @austinvhuang, @o1lo01ol1o, @mitchellwrosen and tell me what you think. Porting over all of core will take a sizable chunk of time so I want to get a second pair of eyes on this before continuing. I think this is going to need some more helper functions to make working with dynamic tensors approachable to newcomers (I'd want to limit the amount of existential quantification that a newbie would have to deal with) --alternatively, we could move back to a list-like `TensorDim` (although I noticed that `TH` only goes up to rank-4 on certain functions, hence `view4`) and use dimensions for any `Torch.Core.Tensor.Static.*`.

Other things that I've added include the `Torch.Core.Tensor.Generic.{Ops,Math,Random}` modules which are hand-written mappings of `raw`. I think these should be autogenerated.

Other comments:
- `Torch.Core.Tensor.Generic` is actually `Torch.Core.Tensor.Raw`, but generic and with some renamed functions.
- postfixed `_` notation is for inplace functions (in keeping with pytorch), so I used a postfixed `'` to indicate when a generic operator would use `SomeDims` (as opposed to `Dims (d::[k])`) --> not sure how this will play out when we move to static operators
- I started off a typeclass for `Torch.Core.Tensor.Dynamic.Double` -- I think there is a _lot_ of overlap with the lower-level generic functions.

The code doesn't compile -- there's quite a lot of code to port over, so let me know what you think.

closes #18

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: nightly-2017-11-24
+resolver: lts-10.0
 
 packages:
 - 'codegen'
@@ -11,7 +11,8 @@ packages:
     commit: e31a2cb894faf4c0cf4a3af3058152149f2b2ee3
   extra-dep: true
 
-extra-deps: []
+extra-deps:
+- dimensions-0.3.2.0
 
 flags: {}
 


austinvhuang:
We've had a couple discussions in slack and, although we'll iterate on the API, it's enough of an improvement to justify merging in my opinion. One minor change - I would reserve the nicely cleaned up function names (e.g. `normal`) for the typeclass functions that are polymorphic over core tensors and retain `c_`, `c_TH` or some other differentiating prefix for the polymorphic raw bindings.

Can you pull the latest upstream changes from master and submit a PR into a separate branch so we can stage the fixes before merging back into master?

stites:
@austinvhuang going to be AFK till January 2nd -- I did a tiny bit more work on the plane but have yet to migrate static modules and address the typeclass issue you mentioned.

RE: compile times -- I haven't heard too many complaints about compile times and large type classes. Usually, it's TemplateHaskell which is the primary offender. That said I'll sanity check on the next iteration.

stites:
the private-to-public switch blocked me from updating this PR. Closing this PR in favor of #42

PR #42, opened 2018-01-04
Typeclasses and dimensions-0.3.2.0

### LINKED ISSUE #18: Convert shape types to use the dimensions package.

stites:
See #39 for description. Closes #18.

The private-to-public switch did something to my fork and blocked me from updating #39, so I'm closing it in favor of this PR.


austinvhuang:
Let's get this merged @stites, can you submit the PR to a staging branch so we can resolve fixes without breaking the master branch?

stites:
Yep. So I just rebased this and need to review some changes. Issues I saw in the rebase:
- TensorRaw started porting code to do operations on RawTensorLong. This is unnecessary as the generic functions can handle all tensor types.
- `nn` got deleted in master???
- `examples` are still broken due to compile errors. Also `hasktorch-examples:exe:ad` changed significantly, so reverted back to master's version.

stites:
everything builds and `stack test` works, sans `standard_gamma` functions, which are new in ATen. I commented out these functions for now so that they would build on my machine which has an older `.so`, but I think it would also make a good first issue for someone to reintroduce this code -- @omaclaren, maybe?

@austinvhuang tell me what you think about the PR -- I think it's ready at this point, even though FloatMath has been commented out (it seems to be a c/p of DoubleMath, and I'd prefer to port DoubleMath to a generic version, like how Static.Generic is really a refactored Static.Double). The code as-is shouldn't be considered complete, there is still a lot of reorganizing that I think needs to happen before a release, but merging this moves us back to smaller PRs.

austinvhuang:
Merged. Great work @stites! The aten function update should be easily patched, and as I mentioned in slack, it's probably time for an upstream refresh since they've been adding some other probability density related functions to TensorMath.

stites:
👍 cool beans! I'm also getting a better sense of what the typeclass-based interface would look like (it's not super crazy) for the high-level `Dynamic` and `Static` modules.

PR #45, opened 2018-01-10
Changed the requirement of gcc-6 to gcc-7 when building on Mac

dataopt:
Match the compiler requirement in `build-aten.sh` and `build-error-handler.sh` when building on Mac.


stites:
:+1:

PR #48, opened 2018-01-18
Introduced dispRawRealFloat for pretty printing 1d and 2d DoubleTensor.

### LINKED ISSUE #53: Show instance for tensors
Show instance for tensors: better solution - use pytorch str functions as template | easy solution - convert to hmatrix and reuse hmatrix show instances

dataopt:
`printTensor` on 1d and 2d double tensors now print properly formatted values.

Closes #53.

### DIFF
--- a/core/src/Torch/Raw/Storage.hs
+++ b/core/src/Torch/Raw/Storage.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE TypeSynonymInstances #-}
+
 module Torch.Raw.Storage
   ( THStorage(..)
   , module X
@@ -5,13 +7,20 @@ module Torch.Raw.Storage
 
 import Torch.Raw.Internal as X
 
+import qualified THByteStorage as S
+import qualified THDoubleStorage as S
+import qualified THFloatStorage as S
+import qualified THIntStorage as S 
+import qualified THLongStorage as S
+import qualified THShortStorage as S
+
 -- CTHDoubleStorage -> CDouble
 class THStorage t where
   c_data :: Ptr t -> IO (Ptr CDouble)
   c_size :: Ptr t -> CPtrdiff
   -- c_elementSize :: CSize
   c_set :: Ptr t -> CPtrdiff -> CDouble -> IO ()
-  c_get :: Ptr t -> CPtrdiff -> CDouble
+  c_get :: Ptr t -> CPtrdiff -> (HaskReal t)
   c_new :: IO (Ptr t)
   c_newWithSize :: CPtrdiff -> IO (Ptr t)
   c_newWithSize1 :: CDouble -> IO (Ptr t)
@@ -51,3 +60,23 @@ class THStorage t where
   p_free :: FunPtr (Ptr t -> IO ())
   p_resize :: FunPtr (Ptr t -> CPtrdiff -> IO ())
   p_fill :: FunPtr (Ptr t -> CDouble -> IO ())
+
+
+-- TODO: Complete the implementations
+instance THStorage CTHByteStorage where
+  c_get = S.c_THByteStorage_get
+
+instance THStorage CTHDoubleStorage where
+  c_get = S.c_THDoubleStorage_get
+
+instance THStorage CTHFloatStorage where
+  c_get = S.c_THFloatStorage_get
+
+instance THStorage CTHIntStorage where
+  c_get = S.c_THIntStorage_get
+
+instance THStorage CTHLongStorage where
+  c_get = S.c_THLongStorage_get
+
+instance THStorage CTHShortStorage where
+  c_get = S.c_THShortStorage_get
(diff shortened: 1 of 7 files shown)


core/src/Torch/Raw/Storage.hs
```diff
@@ -51,3 +60,23 @@ class THStorage t where
   p_free :: FunPtr (Ptr t -> IO ())
   p_resize :: FunPtr (Ptr t -> CPtrdiff -> IO ())
   p_fill :: FunPtr (Ptr t -> CDouble -> IO ())
+
+
+-- TODO: Complete the implementations
```
stites: Can you either complete these in this PR or file a ticket and remove this TODO? TODOs tend to get lost and stale (even though I am probably a purveyor as well).

core/src/Torch/Raw/Tensor/Generic.hs
```diff
@@ -199,53 +205,176 @@ getDynamicDim
...
+
+numberFormat :: RealFloat a => Int -> Int -> [a] -> (String, a, Int)
+numberFormat precision minSz xs = (format, scale, sz)
+  where
+    listParams :: (RealFloat a) => [a] -> (a, a, Bool, Bool, Bool)
+    listParams z =
+      f (NL.infinity, -NL.infinity, True, False, False) z
+      where
+        f z [] = z
+        f (minAcc, maxAcc, intModeAcc, hasInvalidAcc, hasNegInfAcc) (x:xs) =
```
stites: I think this can be cleaned up with a `foldMap` over some newtype deriving Monoid which encodes the rules you are introducing. The cascading if statements take a while to process which triggers a memory of degoes' [_Destroy All Ifs_](http://degoes.net/articles/destroy-all-ifs) post.

stites: Scanning the rest of this function and re-scanning _Destroy All Ifs_, it also looks like you could replace some of those bools with Sum-type ADTs. I think this would also warrant a new module for printing which can be exported by this one. Maybe `Torch.Raw.Tensor.Printing` for now and we can discuss later if we need a `Torch.Generic` parent.

core/src/Torch/Raw/Tensor/Generic.hs
```diff
@@ -199,53 +205,176 @@ getDynamicDim
...
+    defaultScale = 1.0
+    (minAbs, maxAbs, intMode, hasInvalid, hasNegInf) = listParams xs
+    expMin = if (minAbs > 0.0) then 1 + (floor $ logBase 10 minAbs) else 1
+    expMax = if (maxAbs > 0.0) then 1 + (floor $ logBase 10 maxAbs) else 1
+    myMinSz = if hasInvalid then max minSz $ 3 + (if hasNegInf then 1 else 0)
+              else max minSz 2
+    (format, scale, sz) =
+      let expSz = (prec + 4 + (max (length (show expMax)) 
+                                   (length (show expMin)))) in
+      if (intMode) then
```
stites: Looks like a good place for guards and case-statements. Also, this function is getting huge and should probably be split up into top-level helper functions (this way it can be documented and exported for reuse if someone else needs it).

core/src/Torch/Raw/Tensor/Generic.hs
```diff
@@ -199,53 +205,176 @@ getDynamicDim
...
+            let tmpSz = max myMinSz 7 in
+            let format = "%" ++ show tmpSz ++ "." ++ show prec ++ "f" in
+            (format, 10.0**fromIntegral(max minLogScale (expMax - 1)), tmpSz)
+          else
+            let tmpSz = if (expMax > 0) then expMax + 6 else 7 in
+            let format = "%" ++ show tmpSz ++ "." ++ show prec ++ "f" in
+            (format, 1.0, max myMinSz tmpSz)
+
+
+dispRawIntegral tensor = undefined
```
stites: delete and file an issue.

core/src/Torch/Raw/Tensor/Generic.hs
```diff
@@ -199,53 +205,176 @@ getDynamicDim
...
+
+dispRawRealFloat :: (THTensor t, RealFloat (HaskReal t)) => Ptr t -> IO ()
+dispRawRealFloat tensor
+ | length sz == 0 = putStrLn "Empty Tensor"
+ | length sz == 1 = do
+     putStrLn ""
+     let indexes = [ fromIntegral idx :: CLLong
```
stites: I'm almost thinking that we should split out these index traversal functions somewhere but I'm not sure how useful it would be since we delegate so much to TH. Something like `mapM1d`, `mapM2d`, etc, and then this function just becomes:

```haskell
dispRawRealFloat tensor =
  case whateverTheRankFunctionIs tensor of
    0 -> ...
    1 -> map1d fancyPrint tensor
    2 -> map2d fancyPrint tensor
    ...
```

You would preference `mapM` versions so that you can write a quick helper over `Identity` to extract a pure mapping function. Also, in an ideal world mapM1d, mapM2d, etc, are just using a generalized "construct the index traversal order" from the tensor dimensions and we can do away with the rank-specific helpers.

stites:
Looking good @dataopt! I've added some comments which we can pair through, if you'd like.

PR #65, opened 2018-01-23
[WIP] THNN generic bindings

AlejandroCatalina:
So this is a work in progress PR to add THNN bindings to the general codebase. I tried to parse the concrete code and there are not functions so I don't know how to proceed on that.

Anyway, we should probably look at this code and start with next steps, which should probably include writing a new type implementation on raw/src/types/THTypes.hs and others?


AlejandroCatalina:
Okay, I have to make some fixes because THNN has some particularities, for instance their function names are not ```TH<tye>NN_<funname>``` but instead ```THNN_<type><funname>```.

We also have to deal with ```THNNState *``` which is a fancy way of saying ```void *```. I've looked at the pytorch python code and it seems to be a sort of string. So far I've included some tests and, for functions that don't need it, I am passing a ```nullPtr```.

Will update soon!

PR #67, opened 2018-01-28
[WIP] Fill typeclasses raw

AlejandroCatalina:
This is a WIP PR that addresses #50. It's missing some type classes and tests, but I wanted to submit the PR so you can take a look at the code in case there are some strange things.

I think that I make two unconventional changes that will need be reviewed

1. I had to add a second type for some typeclasses to generalise some function arguments, for which I needed some language extensions.
2. I had te add types definitions for `StorageCopy`, which are just synonyms to the `Storage` counterparts.

I also added bindings to the `standard_gamma` functions recently reintroduced by @austinvhuang , which I think is okay.


AlejandroCatalina:
We should now write the `Torch.Core` modules on top of these `Torch.Raw` instances, but I think this one is ready to ship (tests should be included for the `Torch.Core` implementations?).

stites:
👍 I think there are some extra pragmas in there -- eventually, we'll have to clean up the warnings and I think we can get rid of them then.

stites:
Oops! I just merged this (and missed the "[WIP]") -- a bit of a blip but there aren't many of us in the repo, atm, and this is all on dev.

Your assumptions seemed good -- the typeclasses were built out of a reference for `CTHDoubleTensor` and was not written to be fully generic.

AlejandroCatalina:
I think we can actually merge it, I should’ve removed the wip tag! The next step would be to keep working towards higher abstractions!.

PR #73, opened 2018-02-03
some notes on the api

stites:
I've worked through a couple of prototypes and these are my thoughts around the best way to proceed. Feel free to inline comments, ask questions, and point out things that I'm missing.


stites:
going to merge this in to start off our developer docs. Another thing we might consider doing is treating https://github.com/hasktorch/backpack-example/ like a dummy example of how the codebase looks if you strip out the hundreds of codegen'd ffi bindings.

PR #74, opened 2018-02-06
Lasso example

AlejandroCatalina:
This must be merged after #70 .

It would be nice if anyone with knowledge on Coordinate Descent methods would review that code, I think there's something missing for it to be fully working. 

FISTA is reliable and works as expected. 

All in all, this is a very successful example of the implementation of two state of the art methods in machine learning in haskell, which I think is a very nice precedent for hasktorch.

### DIFF
--- a/examples/lasso/Lasso.hs
+++ b/examples/lasso/Lasso.hs
@@ -45,7 +45,7 @@ l1 :: TDS '[M, 1] -> Double
 l1 = tds_sumAll . tds_abs
 
 l2 :: TDS '[M, 1] -> Double
-l2 = tds_sumAll . tds_square
+l2 = sqrt . tds_sumAll . tds_square
 
 prox_l1 :: TDS '[M, 1] -> Double -> TDS '[M, 1]
 prox_l1 w l        = tds_sign w ^*^ max_plus (w ^- l)
--- a/examples/lasso/Lasso.hs
+++ b/examples/lasso/Lasso.hs
@@ -1,0 +1,189 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies        #-}
+
+module Main where
+
+import           Data.Function                         ((&))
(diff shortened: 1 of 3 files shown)


examples/lasso/Lasso.hs
```diff
@@ -0,0 +1,189 @@
...
+prox_l1_single :: Double -> Double -> Double
+prox_l1_single w_i l = signum w_i * max (w_i - l) 0.0
+
+tds_getDim2d :: TDS '[N, M] -> Int -> Double
+tds_getDim2d _ i
+  | i == 0 = ((realToFrac $ natVal (Proxy :: Proxy N)) :: Double)
+  | i == 1 = ((realToFrac $ natVal (Proxy :: Proxy M)) :: Double)
+  | otherwise = error "Only works on 2d tensors; `i` must be either 0 or 1"
+
+coordinate_descent ::
```
ocramz: @AlejandroCatalina could you share a reference where this and the next CD methods are described?

examples/lasso/Lasso.hs
```diff
@@ -0,0 +1,189 @@
...
+
+loss :: (TDS '[N, M], TDS '[N, 1]) -> TDS '[M, 1] -> Double
+loss (x, y) w =
+  ( tds_sumAll . tds_square $ y ^-^ (x !*! w) ) + l1 w
+
+l1 :: TDS '[M, 1] -> Double
+l1 = tds_sumAll . tds_abs
+
+l2 :: TDS '[M, 1] -> Double
+l2 = tds_sumAll . tds_square
```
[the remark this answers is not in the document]
AlejandroCatalina: Ouh, that's right, I'll add it, thanks!

examples/lasso/Lasso.hs
AlejandroCatalina: The coordinate descent algorithm for Lasso is proposed by Hastie and Tibshirani and widely covered in their book https://web.stanford.edu/~hastie/StatLearnSparsity_files/SLS_corrected_1.4.16.pdf in section 2.4

Issue #71, opened 2018-02-02
write `ptr` combinator to clean up codegen functions


stites:

stites:
Optionally to add to this ticket is a clean-up of `THType`

Issue #69, opened 2018-01-31
add raw bindings for THCUNN


stites:
Next after #63. Note that this will require a cabal flag and CUDA bindings might have to be moved to a seperate subpackage (hasktorch-cuda) so that people can choose not to compile if they don't have the right cuda library files.

Issue #44, opened 2018-01-09
Document core/src/Torch/Raw module


stites:
It's pretty straightforward, but all changes should be documented

Issue #21, opened 2017-11-28
Codegen THHalf* function to use CTHHalf  arguments


stites:
I'm not sure if this is actually the correct thing to do (I haven't dug too deep into codegen, yet) but some sed replacements let this compile and, semantically, using CTHHalf at this stage might be correct.

This would also let us use hpack in `raw` without too much pain

PR #79, opened 2018-03-09
THC support, NN/CuNN support with restructured backpack modules

stites:
Codegen looks good. Changes which still have to happen:
- [x] backpack signatures need to be re-organized to account for how THC breaks down header files
- [x] typeclasses should be reorganized
- [x] indefinite code should be reorganized
- [x] core code should be reorganized.
- [x] finish e2e for unsigned types
- [x] finish e2e for signed types
- [x] finish e2e for floating types
- [x] get all examples working again
- [x] Generated THC structs still need to be lined up with codegen 
- [x] remove the verbose flag from the new codegen CLI tool
- [x] bump FFI tests
- [x] update circleci to use `cabal new-build`
- [x] update README.md: cover architecture restructuring, changes to codegen, overview to build process. update Makefile
- [x] compile on machine without cuda to make sure we don't break the build for CPU-bound users.

### Scope creep:
- ~~`Torch.Core` has a lot of constraint-based code which can live in `Torch.Classes`. This would make `Torch.Core` a lot cleaner.~~ done!
- ~~THNN and THCUNN need to be reintroduced.~~ This is happening.

### Issues to file:
- **Compile THC Float Tensors**: There's some compiler issue with THC Float Tensors. They don't seem to compile on any machine I am running (both in ATen and in the vanilla TH lib). Need a do-over on this one, but we can merge with just Double support.
- **Need to look into specializing the typeclasses**
- **Add notes to Development guidelines**: Typeclasses have been split into staticly and dynamic variants. All contributions should go there and anytime someone has a significantly tricky implementation requiring staticly-typed tensors, they can add some core functions to the typeclass and we can `sudo` in the indefinite instances (see `Torch.Indef.Static.Tensor`).
- **Static-NN needs work**: The Static typeclass for the neural network library isn't finished. The dynamic library alone was a terrible ordeal and I think the entire API (even the typeclass itself) is sucky. I think this requires some more examples and a bit more work -- for now we have dynamic tensors to keep us going.
- **Add to "development tips"**: backpack doesn't recompile well. If you change a signature, it's often the case that you have to blow away your entire dist-newstyle and rebuild before succeeding the build.
- **Double-check C-functions we want to include**: THRandom.h looks to be incomplete -- I've ignored the THGeneratorState in the refactor of codegen and it's caused a few functions to be elided.
- **Clean up boilerplate with Template Haskell**: There's a _lot_ of code duplication. While I think it was required to shape the code up to this point, a future refactor needs to happen to scrap the boilerplate with template haskell.
- **Add notes to README about API**: I've noticed that the c-function signatures aren't true to the pytorch signatures. I've started migrating our functions calls to reference prefixed-`_`s as "direct c-call to mutating function", suffixed-`_`s as "pytorch-style mutation which returns the mutated tensor".
- **Refactor with Template Haskell to remove boilerplate.** This issue has been filed.
- **Better management of dynamic-/static- tensor schizm.** I think this depends on making Dynamic Tensors higher-kinded, but there might be a better way 
- Lasso is hitting an infinite loop (I think?) -- it's a pretty complex implementation which might benefit from unit tests and recursion schemes. I'll take a stab at it after this PR.
- Linear types could be used for inplace functions.
- needs autodiff

### Former issues to file that have been resolved:
- ~~Also, because of how THC breaks up TensorMath, we _could_ break up the "mega-math" module that has come about.~~ This is required for the backpack signatures and needs to be addressed in the first point.
- ~~add CPP flags for compiling with MAGMA, LAPACK, and BLAS and CuBLAS (also revamp the "Half" CPP flag and think about how the "lite" flag might work). -> This can either be done here or bumped to a new issue.~~ This might be a non-issue. hasktorch will compile but GHC won't run the code if it can't find the symbols (we just need libATen present).
- ~~TH and THC differ in all of their Random code. One library takes an argument of a generator, the other doesn't. Furthermore, there are cuda- and non-cuda based signatures that could exist, but we unify everything. So there are a few more branching factors to take into account and *it will require some more metaprogramming keep everything sane in the same way we manage state* (need to discuss this in slack later).~~ The Random code is required to get examples to work (ie -- is required for pragmatic code). I've gone ahead and included these branching factors. _more branching factors exist and will need to be addressed_.
- ~~ATen is messes with the dependencies in TH and THC. One example of this is the removal of the C-based `expand` functions for the adoption of their CPP variaent. Expand would be vital for us to do broadcasting and, while it may not be the best option at the moment, it seems like it is our only option. It looks like we need to move back to building all 4 libraries from scratch, or continue without full functionality until ATen releases a C api.~~ This has been addressed by forking and branching at the last commit which includes expand functions. See hasktorch/ATen#hasktorch-expand.


stites:
Wow! The end to a 1m PR. This is now code complete -- although, we still need to update docs, file the listed issues, and bump the README.md and DEVELOPMENT.md

stites:
PS, this closes: #71 #69 #60 #59 #52 #51 #50 #44 #21

PR #83, opened 2018-04-23
Start bumping documentation

stites:
A quick side-track to document most of `-core` with `cabal new-haddocks`. Merging this straight into master (so long as  CI passes). I wasn't sure what to do about the copyrights -- I've been labeling these as `Hasktorch dev team` in all new cabal files, but I missed a bunch here when using the default in my snippets (which copyrights to Sam Stites). Let me know what you think I should do about that.


austinvhuang:
Hey Sam, thanks for adding this great documentation. For copyright - would be good to edit that so it's not confusing, but no big deal if it's patched in a later commit.

PR #90, opened 2018-05-02
Rewrite DEVELOPERS.md a bit

mitchellwrosen:
This patch:

* Removes `build.sh` files. I think these are awkward to maintain, and not really a part of the definition of this codebase (or are they?).

  The basic ask of developers is to install `libATen.so` in whatever way they are comfortable with. `nix`, `brew`, etc... we happen to include an `ATen` submodule for convenience, so I think it's sufficient just to document how to install manually (which I did in `DEVELOPERS.md`).

* Adds some documentation to `DEVELOPERS.md` about `cabal.project.local`, which is useful for specifying the location of the `ATen` installation, and disabling CUDA builds, if desired.


stites:
`build-aten.sh` is required for CI to work

stites:
Ugh. I fixed the merge conflicts but started the fix against my current HEAD so this commit was incredibly dirty. Pushing a rebase wound up closing this issue -- sorry about that @mitchellwrosen. The commit was at b63d977 if you would like to re-submit a PR.

Issue #103, opened 2018-07-11
Build with nix


stites:
Ideally, this would simplify the build process and would allow us some better CI and development workflows. Normally this would be wishful thinking but since @tstat is a native nix user, it might actually get picked up.

For guidelines, check out how [CircleCI builds ATen](https://github.com/hasktorch/hasktorch/blob/master/.circleci/config.yml#L99) and the [build-aten.sh](https://github.com/hasktorch/hasktorch/blob/master/vendor/build-aten.sh#L46) script.

stites:
Status update from #105 :

> On nix support, after consulting with @tstat: this wont be possible until either nix supports `cabal new-build`, or `cabal new-build` becomes `cabal build`. I have a feeling the latter will happen first. In the meantime, you can see https://github.com/tstat/hasktorch/pull/1 which contains a nix formula for ATen -- I'm actually supposed to remove the redundant `build-vendor.nix`.

PR #104, opened 2018-07-31
LeNet + CIFAR-10 example

stites:
Description should be updated later:
- adds `hasktorch-zoo` -- this will, hopefully, kick off the "call for collaboration."
- `hasktorch-zoo` includes a really dumb dataloader. Dataloading is sort of the bottleneck here.
- cabal2nix all the packages (even though this doesn't really work until we can build ATen with nix) (re: #103 , @tstat )

I think, when this is done, we are ready for hackage and make a subreddit announcement!

TODO:
- [ ] Fix exploding loss function
- [ ] stack minibatches into tensors.
- [ ] Shuffle training and testing data


austinvhuang:
Going ahead and merging so I'll work off this, note to add missing gd dependencies to fix ci.

PR #106, opened 2018-08-05
Add GD dependency to CI, address common build issues in readme

austinvhuang:

PR #107, opened 2018-08-13
[Depends on #106] Continued work on the hasktorch-zoo

stites:
Followup to #104, which broke a few things.

In addition to a working LeNet example, this PR does the following:
- add 8.4.3 freeze file
- start a dependently-typed catArray function
- Add an annotated CPU-first example of a `cabal.project.local` file.
- backport hasktorch/ATen to find Intel-MKL


stites:
This branch is going to need a little more love. Closing it for now.

PR #111, opened 2018-09-13
Update TH Reference in codegen/Readme.md

adamwespiser:
Fixed a dead link to TH in the codegen/Readme.md so it now points to the TH library within the PyTorch project.


austinvhuang:
Thanks @adamwespiser!

PR #113, opened 2018-10-03
add backprop regression example (WIP - don't merge yet)

austinvhuang:
Backprop version of a regression example. PR'ing for visibility, don't merge yet.

### DIFF
--- a/examples/experimental/backprop-regression/BackpropRegression.hs
+++ b/examples/experimental/backprop-regression/BackpropRegression.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -10,6 +11,7 @@ module Main where
 import System.IO (hFlush, stdout)
 import Text.Printf
 
+import Control.Monad (foldM)
 import Data.Generics.Product.Fields (field)
 import GHC.Generics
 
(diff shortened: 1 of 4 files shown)


examples/experimental/backprop-regression/BackpropRegression.hs
```diff
@@ -0,0 +1,86 @@
...
+type P = '[1, 2]
+type Precision = Double
+
+seedVal :: RNG.Seed
+seedVal = 3141592653579
+
+data Regression = Regression {
+    _linearLayer :: Linear 2 1
+}
+makeLenses ''Regression
```
stites: You (and I, both) should get into the habit of replacing template haskell with [generic-lens](https://hackage.haskell.org/package/generic-lens) which has equal performance, greater flexibility, and no TH (which can nuke compile times if you aren't careful).

austinvhuang: Haven't used generic-lens, do you have an example?

stites: You should be able to just replace the generated lens, `linearLayer`, with `field @"linearLayer"` (the `@` comes from the TypeApplications extension). I _think_ that's all! Since it's a lens, it should also be fully compatible with Justin Le's `^^.` combinator.

austinvhuang: Okay, uses generic-lens now.

austinvhuang:
Am going to close this PR to merge with the dev branch and continue work there.

stites:
sounds good to me!

PR #120, opened 2018-11-24
dimension-safe `diag`, simple autoencoder example, fix overwriting values sampler bug

austinvhuang:
- Dimension-safe `diag` - previously it was possible to pass value-level `k` parameters that would lead to underlying tensor dimensions that were different from the advertised type. Replaced default `diag` with a dimension-safe version. Kept the older version as well as `diagUnsafe` as well as the `diag1d` special case.

- Simple autoencoder - linear/relu layers with a bottleneck. Tested on a high-dimensional gaussian mixture (for now, will replace with something more interesting like mnist once the data loaders are ready).

- Started defining a data representation for Relu activations (analogous to the Linear data representation), with the idea of a more automated declarative API that automatically defines forward/backward prop without the current verbosity/redundancy. This isn't used yet though and the design could change down the line.

- Add `NOINLINE` to `withInplace`, resolves bug where values produced by samplers like `normal` were getting overwritten by subsequent calls to `normal`.

- Tests to check for regression of the `withInplace` sampler bug were added to the hasktorch test spec.

- There was a minor bug in the lapack test spec where one of the elements of the test matrix wasn't getting initialized. It was passing (erroneously) in CI because the random initialization value produced an invertible matrix, but this will be system-dependent. This is resolved.


stites:
:tada: :boom: :tada:
LGTM! I like the quick simple example and I apologize for not catching the O0 v. O1 bug sooner.

If you could migrate `rng-bug/RNGMinimal.hs` to `hasktorch/tests/Torch/Core/RandomSpec.hs` that would be excellent! 

Notes which are optional at this stage:
- Style-wise, I think we should start using the brittany formatter but there's no need to do that just yet (it's the default haskell-ide-engine and a personal preference).
- DRY-related things (for future reference) we should probably start pulling some of the common functions into `hasktorch-zoo`

PR #123, opened 2018-12-01
Bump up ffi submodule

ayush1999:


stites:
:+1:

stites:
Looks like the CI freeze file was flexible enough to handle _both_ 8.4.3 and 8.4.4 -- a bit disappointing on my part to only discover this now. That said, we should be running CI on a user's freeze file since we work outside of stack. I'm checking a commit which I'll push here in a bit.

Issue #131, opened 2018-12-15
Add precedence to infix operators


austinvhuang:
Need to add infix operator precedence. For example, currently `!*!` does not take precedence over `^-^` or `-`

austinvhuang:
Resolved by latest PR https://github.com/hasktorch/hasktorch/pull/132

PR #132, opened 2018-12-16
Fix library bugs seen in the GP example

austinvhuang:
- lapack getri algorithm type signature was too broad: `getri :: Tensor d -> Tensor d'` to `getri :: Tensor [d, d] -> Tensor [d, d]` - i.e. input and outputs should be matrices of equal dimensions w/ the output being a matrix inverse equal to the input.
- `csub` can't be inlined, otherwise evaluation ordering for a composite expression can lead to runtime incorrect dimensions.  In the posterior covariance evaluation:
```
postCov = covXX ^-^ covXY !*! invY !*! covYX
```
Was getting:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  matrices expected, got 2D, 1D tensors at /Users/huanga13/projects-personal/hasktorch/vendor/aten/src/TH/generic/THTensorMath.c:1944
bash returned exit code 0
```
after prohibiting `csub` inlining, this was resolved (EDIT: behavior changes with -O0, for which the error remains. Temporarily resolved now by forcing strict evaluation of the covXX as a workaround, but will have to flag this as an issue).
- infix operators didn't have precedence assigned, leading to matrix-matrix/matrix-vector multiplication being the same precedence level as pointwise subtraction. This was fixed and resolves https://github.com/hasktorch/hasktorch/issues/131


austinvhuang:
Still have some CI issues to resolve, don't merge yet.

PR #135, opened 2018-12-29
Add Dockerfiles

junjihashimoto:
This PR provides a prototype to build docker-images hasktorch and debian-packages of ATen.
To make the docker-image supporting GPU, we need nvidia-docker.
debian-packages of ATen are made from docker-image using deb-cpu/gpu.Dockerfile.
What do you think of this?


austinvhuang:
This is great, thanks @junjihashimoto! Am merging, have a few minor follow-ups - will chat on slack.

PR #141, opened 2019-01-28
Build nix-shell for gpu support

### LINKED ISSUE #143: Pin nix pkgs to a hash
#103 

Is really not complete. 
https://github.com/hasktorch/hasktorch/blob/master/shell.nix#L5
https://github.com/hasktorch/hasktorch/blob/master/build.nix#L49

Both use `<nixpkgs>` and as a result, the project does not rebuild reproducibility. I spent some time trying to get a hash that will build Aten, but failed. Please pin to a specific rev of nixpkgs, such that the project can build.

stites:
Some quirks which came up:
- our backend, `libATen-0.4.0`, doesn't seem to build with cuda-10 or cuda-9.2
- cudatoolkit_9_0 can't build with the default gcc (the version must be less than 6).
- magma-2.3.0, which is magma's hosted documentation, can't build with gcc5 -- I've bumped it up to magma-2.4.0.

While [cuda 9.0 is supposedly compatible with nvidia 410](https://devtalk.nvidia.com/default/topic/1042638/is-nvidia-driver-410-57-incompatible-with-cuda-9-0/?offset=5), I get a runtime error when building hasktorch-thc:
```
THCudaCheck FAIL file=/tmp/nix-build-hasktorch-aten.drv-0/aten/src/THC/THCGeneral.cpp line=70 error=35 : CUDA driver version is insufficient for CUDA runtime version
terminate called after throwing an instance of 'std::runtime_error'
  what():  cuda runtime error (35) : CUDA driver version is insufficient for CUDA runtime version at /tmp/nix-build-hasktorch-aten.drv-0/aten/src/THC/THCGeneral.cpp:70
Test suite spec: FAIL
Test suite logged to: dist/test/hasktorch-ffi-thc-0.0.1.0-spec.log
```
So I have a feeling this will build if you run nvidia_390 -- ~I'll confirm this on my laptop and merge it in then.~ looks like I can't compile nvidia390 on my nixos box. I'll have to take another look at this later.~

## Cuda packages now build on NixOS and Ubuntu (sort of).

I've worked out all of the issues with building this on my various local boxes (Ubuntu and NixOS). The caveat is that the `nvidia_x11` drivers _should_ line up with the system drivers. For Ubuntu, this means that you must have libcuda.so in your LD_LIBRARY_PATH (potentially also libcuda.so.1 -- on my Ubuntu box they are at different locations). For NixOS, this means that you link with the `nvidia_x11` version that you also load in your kernel modules.

The above is sort of codified in https://github.com/hasktorch/hasktorch/pull/141/commits/61ee71548f0adb8847a1c6bc3e1bb1041570a75e#diff-5204cbe23b218cae9de0b1aedd4edcd2R8, `nvidia_x11-pinned` is, selfishly, pinned to my nvidia drivers here: https://github.com/hasktorch/hasktorch/pull/141/commits/bf114db1a7c5104a7293a0c85d011cebda83a013#diff-4b0dc7588202ee4f4fe52c5af065f526R93.

I would highly encourage future users to run `nix-build ./build.nix --arg cudaSupport true` for the first time before dropping into a `nix-shell --arg cudaSupport true` so that the shared object files are not garbage collected by accident. Because of the dependency of our backend on cuda-9 and gcc-5 (as well as magma-2.4 and fortran from gcc-5) there is a _lot_ of building and tests that must be run. **Building dependencies for the first time takes a long time (like an hour or two). Most of this it due to the custom magma and fortran libraries.** Building hasktorch for the first time can also take up to 30m, but that's more acceptable since most new users don't use `cabal v2-build` and you have to rebuild the world. 

CC: @MMesch

Closes #143

### DIFF
--- a/build.nix
+++ b/build.nix
@@ -117,9 +117,13 @@ in {
     # hasktorch-examples
     # hasktorch-indef
     hasktorch-codegen
-    hasktorch-ffi-th
+
+    # hasktorch-ffi-thc will have a mismatched cuda versions depending
+    # on if you build on nixos or not, haskell's cuda-9.0 is also acting up.
     # hasktorch-ffi-thc
+    hasktorch-types-thc
+
     hasktorch-ffi-tests
+    hasktorch-ffi-th
     hasktorch-types-th;
-    # hasktorch-types-thc;
 }
--- a/build.nix
+++ b/build.nix
@@ -1,4 +1,4 @@
-{ compilerVersion ? "ghc844", cudaSupport ? false, mklSupport ? true }:
+{ compilerVersion ? "ghc844", cudaSupport ? false, mklSupport ? true, atenDev ? false }:
 let
   config = {
     allowUnfree = true;
@@ -30,6 +30,7 @@ let
         pkgs.callPackage ./ffi/deps/hasktorch-aten.nix {
           inherit (pkgs.python36Packages) typing pyaml;
           inherit cudaSupport mklSupport;
+          dev = atenDev;
           magma_cudatoolkit_9_0 = magma;
           stdenv = stdenv5;
           gfortran = gfortran-gcc5;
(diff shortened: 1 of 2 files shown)


stites:
Updated the PR comment to reflect current status (which is "works on my machine"). Since I think I'm the only dev using nix I'm just going to merge this in. Primarily I'll be using the NixOS build (I'm trying not to futz with my primary research box) -- if you want to build Ubuntu+Nix and run into trouble, ping me on gitter and I can help debug.

PR #144, opened 2019-02-05
Put solution to aten library linkage issue

AdLucem:
I put the solution for the Aten library linking issue in DEVELOPERS.md. 

It would perhaps be useful to put this command somewhere in the build script itself? I'm mildly scared of making a change to the build script and somehow breaking the build system even further, however, so I wrote it down as a "here's the duct tape to fix this issue" thing.


stites:
Thanks @AdLucem!

PR #145, opened 2019-04-04
Split indef into core, gpu and cpu

stites:
Perhaps this PR is indicative of making the wrong decision to denormalize the backpack architecture.

All dependent code which is not `storagedata`, `tensordata` or `Storage.newWithData` but which depends on one of these functions (mostly on Storage's IsList instance), is copy and pasted twice: into `indef/gpu/` and `indef/cpu`. Perhaps a better solution would have been to write the `IsList` instance by hand in C, pushing this down the heirarchy -- but would cause a recursive dependence between `hasktorch-ffi-th*` and `hasktorch-types-th*`. Perhaps another option, which was unexplored, would be a typeclass in `indef/core/` which we can specialize in `indef/{cpu,gpu}/`. 

This PR splits up indef into the following three packages and fixes the memory allocation problem on cuda devices.

### `indef/core/`

Contains what used to be `indef/` and holds the `hasktorch-indef.cabal` file

### `indef/cpu/` 

This has a __new__ `Torch.Indef.Storage`, `Torch.Indef.Dynamic.Tensor`, and `Torch.Indef.Static.Tensor`. The old corresponding files now live in `Torch.Indef.MODULE.Internal` under `indef/core/` and any code that is, or depends on, `storagedata` or `tensordata` or `Storage.newWithData` have been moved to original package namespace, `Torch.Indef.MODULE`, in `indef/cpu/`. Any `Torch.Indef.MODULE` will re-exports all of its `Internal` functions.

This package also cleanly seperates out the random modules and removes the `TH`/`THC` suffix.

### `indef/gpu/` 

Identical to `indef/cpu/` but uses marshalling functions from hackage's [cuda-0.9.0.3](https://hackage.haskell.org/package/cuda-0.9.0.3/docs/Foreign-CUDA-Runtime-Marshal.html) to correctly allocate memory on a cuda device.
 
---

I'm working off of an old ATen.so which is CPU-only (I didn't realize this until just now). @TomLippincott feel free to pull this down and check it out. The fix for your bug report can be found in the second-to-last commit: https://github.com/hasktorch/hasktorch/commit/08b3bd5a72e300b8601257a9d5ad102f9e5c078b

The last commit removes the unnecessary copies you'll see in that commit (which is an artifact of a memory-management bug-hunt).


stites:
ah! looks like I forgot to exclude cuda by default somewhere.

TomLippincott:
Hmmm, perhaps related, I'm seeing a new CUDA compilation error (note this is from a complete switch to your gpu-split branch, I don't know if this was the correct way to test, but couldn't find a lighter-weight way to pull in changes from a pull request):

```
/expscratch/tlippincott/hasktorch/dist-newstyle/tmp/src-183764/cuda-0.9.0.3/dist/setup/setup.hs:301:36:
error:
• Couldn't match type ‘Platform -> String’ with ‘[Char]’
Expected type: String
Actual type: Platform -> String
• Probable cause: ‘dllExtension’ is applied to too few arguments
In the second argument of ‘(<.>)’, namely ‘dllExtension’
In the first argument of ‘isInfixOf’, namely
‘("" <.> dllExtension)’
In the first argument of ‘find’, namely
‘(isInfixOf ("" <.> dllExtension))’
|
301 | return $ find (isInfixOf ("" <.> dllExtension)) (lines nmOutput)
| ^^^^^^^^^^^^
)
Failed to build hasktorch-indef-0.0.1.0 (which is required by
exe:isdefinite-gpu from hasktorch-0.0.1.0). The failure occurred during the
final install step. The exception was:
/expscratch/tlippincott/hasktorch/dist-newstyle/tmp/package-registration--183764:
removeDirectoryRecursive:removeContentsRecursive:removeDirectory: unsatisfied
constraints (Directory not empty)
Failed to build hasktorch-indef-cpu-0.0.1.0 (which is required by
exe:isdefinite-cpu from hasktorch-0.0.1.0, exe:xor from
hasktorch-examples-0.0.1.0 and others).
```

stites:
(A recap from slack) This was a ghc-8.6.1 and cuda-0.9.0.3 bug. I've bumped the cuda dependency to `-0.10.0.0` (which apparently works with CUDA-9.0!). This compilation bug should be gone, but the CI error will persist because we are building `cabal new-build all` in our `circle.yaml`, but we also have a transitive CUDA requirement in `hasktorch-indef-gpu`.

Because this PR is an intermediate representation of all of this work, I'm adding a fix at the end of #146 (basically running `cabal new-build hasktorch-examples` so that cabal ignores hasktorch-indef-gpu).

PR #147, opened 2019-05-06
Mnist example

alexander-myltsev:


alexander-myltsev:
@stites @austinvhuang I've implemented all the parts. Before I make the final launch and debug, would you tell me if everything is OK basically?

austinvhuang:
Thank you very much for the PR @alexander-myltsev ! Sorry for the slow reply, been tracking work on the next release lately. One thing you may be interested in for stuff like this is that @stites implemented a generic dataloader that should work with hasktorch over as part of datahaskell https://github.com/DataHaskell/dh-core/blob/10cbf7ea35b06bc0dd96fd8b5cc2fc6cbfc0ba51/datasets/src/Numeric/Dataloader.hs

Join us on the hasktorch slack if you want to discuss!

PR #153, opened 2019-08-05
Rename the package-name from ffi to libtorch-ffi

junjihashimoto:
Before uploading this package to hackage, this pr renames `ffi`-package to `libtorch-ffi`-package.

### DIFF
--- a/codegen/codegen.cabal
+++ b/codegen/codegen.cabal
@@ -2,7 +2,7 @@ name:                codegen
 version:             0.1.0.0
 synopsis:            parse torch yaml spec files, generate code
 -- description:
-homepage:            https://github.com/githubuser/ffi-experimental#readme
+homepage:            https://github.com/hasktorch/hasktorch#readme
 license:             BSD3
 author:              Austin Huang
 maintainer:          hasktorch@gmail.com
--- a/examples/examples.cabal
+++ b/examples/examples.cabal
@@ -2,7 +2,7 @@ name:                examples
 version:             0.2.0.0
 synopsis:            examples for the new version of hasktorch
 -- description:
-homepage:            https://github.com/githubuser/ffi-experimental#readme
+homepage:            https://github.com/hasktorch/hasktorch#readme
 license:             BSD3
 author:              Austin Huang
 maintainer:          hasktorch@gmail.com
@@ -30,7 +30,7 @@ executable elman
   build-depends:       base >= 4.7 && < 5
                      , hasktorch
                      , mtl
-                     , ffi
+                     , libtorch-ffi
 
 
 executable regression
@@ -40,7 +40,7 @@ executable regression
   build-depends:       base >= 4.7 && < 5
                      , hasktorch
                      , mtl
-                     , ffi
+                     , libtorch-ffi
 
 executable cnn
   hs-source-dirs:      cnn
--- a/ffi/libtorch-ffi.cabal
+++ b/ffi/libtorch-ffi.cabal
@@ -1,8 +1,9 @@
-name:                ffi
-version:             0.1.0.0
-synopsis:            test out alternative options for ffi interface to libtorch 1.0
+name:                libtorch-ffi
+version:             1.1.0.0
+-- The prefix(1.1) of this version("1.1.0.0") is the same as libtorch's one.
+synopsis:            test out alternative options for ffi interface to libtorch 1.x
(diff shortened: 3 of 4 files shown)


stites:
Is there a strong desire to prefix with `libtorch-*`? `hasktorch-ffi` adds a namespace which is easy to find, is more flexible if there are any adjustments to make to the backend of hasktorch (ie: if we want to bundle something else), and will last for the lifetime of hackage (hasktorch-ffi will become abandonware if we move to libtorch-ffi).

My vote would be to keep hasktorch-ffi and split it into libtorch-ffi (or just "libtorch") at a later point. Later points being adjustments to the backend (see above), or even just a future 1.0 release so that we aren't polluting hackage (which is my primary concern here).

stites:
Is there a strong desire to prefix with `libtorch-*`? `hasktorch-ffi` adds a namespace which is easy to find, is more flexible if there are any adjustments to make to the backend of hasktorch (ie: if we want to bundle something else), and will last for the lifetime of hackage (hasktorch-ffi will become abandonware if we move to libtorch-ffi).

My vote would be to keep hasktorch-ffi and split it into libtorch-ffi (or just "libtorch") at a later point. Later points being adjustments to the backend (see above), or even just a future 1.0 release so that we aren't polluting hackage (which is my primary concern here).

stites:
I added a comment about reusing the `hasktorch-ffi` namespace to be good hackage citizens and then realized that we will already be dealing with abandonware since backpack required hasktorch-ffi be split between TH and THC. Backpack did cause a ton of package fracturing -- I'll have to reach out to the hackage maintainers to get them to depreciate these.

Since the namespacing issue isn't a huge deal, I see three options: use `libtorch-ffi`, use `hasktorch-ffi`, or use `libtorch` and let the FFI bindings be implicit in the name. I sort of see them all as having pros and cons but I would lean towards the latter for simplicity.

junjihashimoto:
You are right. We have to think about pros and cons.
I am not confident in how to divide pros and cons.

-libtorch-ffi
  - pros
    - Probably, we can not insert the other c++libraries except for libtorch.
  - cons 
    - We can not know  libtorch-ffi is related to hasktorch.
hasktorch-ffi
  - pros
    - We can see  hasktorch-ffi is related to hasktorch.
  - cons 
   - We can insert the other c++libraries except for libtorch.(I do not know whether this is pros or cons.)
libtorch
  - pros
    - The name is simple.
    - Probably, we can not insert the other c++libraries except for libtorch.
  - cons 
    - We can not know  libtorch-ffi is related to hasktorch.

stites:
I think there should be some planning of what the final packages will become. I like how `libtorch*` is direct and also breaks away from the overly-normalized (in the SQL-sense) namespace I've created in v0.0.1 : (

In the project right now I'd expect the following:

- `codegen/` pushed to hackage as `hasktorch-codegen`. Hasktorch currently ships with assumptions around mkl and fbgemm. I'm not convinced this is necessary, but someone may want to use this to generate a custom-built hasktorch around their deployment.
- `ffi/` pushed to hackage as ???
- `hasktorch/` as pushed to hackage as `hasktorch`.
- I don't think `examples/` needs to be pushed up, right?

The less we push up, the more I think it's fine to be direct with ffi as libtorch (which is more direct). That said, I think `hasktorch` is currently only a wrapper around the FFI. There's a couple of questions, then, about where you and @austinvhuang want future code contributions going:
- Where do you expect the static interface to go? 
- Where do you expect common functions to go (optimizers, data loaders, etc)?

austinvhuang:
Will go ahead and merge these changes.

As far as hackage goes, my current thinking is to initially push a single hasktorch library and not a separate ffi package at all until we start getting requests to do so. My suspicion is that contributors may are probably the only ones to consume the ffi.

Codegen even more so, I think we'll be the only consumers there. We could always change our mind later if someone asks for it.

i.e. keep the package ecosystem footprint as simple as possible until there's a demand for things beyond the core library.

If we do split things off, I like the convention of using the main package name hasktorch-* follows convention of other mini-ecosystems (pipes-* conduit-*) makes sense so the linkage is clear.

For now I would add optimizers etc. to the main library. In cases where there's clean composable interfaces with other parts of the ecosystem (e.g. datahaskell or plotting libraries) we can take advantage of that. But for basic functionality, would just keep adding to the main library (ala pytorch).

PR #163, opened 2019-08-31
Add template codes to develop static-typed-functions

junjihashimoto:
I want to think adding static typed tensor practically.
At first, add template codes of static typed functions, but the definitions of some/many functions are wrong.
Next, add test for the functions and think next steps.

### DIFF
--- a/hasktorch/src/Torch/Static/Native.hs
+++ b/hasktorch/src/Torch/Static/Native.hs
@@ -37,7 +37,7 @@ import qualified ATen.Managed.Cast
 import ATen.Cast
 
 
-import Prelude hiding (sin, abs, max, min)
+import Prelude hiding (sin, sinh, cos, cosh, tan, tanh, asin, asinh, acos, acosh, atan, atanh, abs, max, min, exp, log, round)
 import qualified Torch.Tensor as D
 import qualified Torch.TensorFactories as D
 import Torch.Functions (Reduction(..), Tri(..), isUpper, kOne)
@@ -48,6 +48,7 @@ import Torch.Scalar
 import qualified Torch.TensorOptions as D
 import Data.Reflection
 import ATen.Cast
+import Control.Arrow ((&&&))
 
 ---
 
@@ -67,8 +68,7 @@ sumAll :: Tensor dtype shape -> Tensor dtype shape
 sumAll t = unsafePerformIO $ (cast1 ATen.sum_t) t
 
 -- |
--- >>> t = abs (ones :: Tensor Float '[2,2])
--- >>> (dtype t,shape t)
+-- >>> dtype &&& shape $ abs (ones :: Tensor Float '[2,2])
 -- (Float,[2,2])
 abs :: Tensor dtype shape -> Tensor dtype shape
 abs t = unsafePerformIO $ (cast1 ATen.abs_t) t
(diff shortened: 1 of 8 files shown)


tscholak:
are you sure this should go into `master`?

austinvhuang:
@tscholak we've had kind of an informal convention to mark PRs as work-in-progress (WIP) for visibility/discussion. These aren't intended to be merged yet.

junjihashimoto:
@austinvhuang Thank you for explaining the convention. 
This is just a trial and there is no intention of merging.
It makes a template for static typing automatically, so it is easy to track upstream(libtorch) changes.
For now, I am checking how hard it is to fix and test by hand.

hasktorch/src/Torch/Static/Native.hs
```diff
@@ -23,19 +23,1474 @@ import Data.Finite
...
+-- >>> (dtype t,shape t)
+-- (Float,[2,2,2])
+type family SqueezeAll (shape :: [Nat]) :: [Nat] where
+    SqueezeAll '[] = '[]
+    SqueezeAll (1: xs) = SqueezeAll xs
+    SqueezeAll (x: xs) = x ': SqueezeAll xs
+squeezeAll :: Tensor dtype shape -> Tensor dtype (SqueezeAll shape)
+squeezeAll t = unsafePerformIO $ (cast1 ATen.squeeze_t) t
+
+binary_cross_entropy_loss :: Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> Reduction-> Tensor dtype shape
```
tscholak:
```suggestion
binary_cross_entropy :: Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> Reduction -> Tensor dtype '[]
```
see https://pytorch.org/docs/stable/nn.functional.html#binary-cross-entropy

tscholak: ah, it's more complicated than that. I just ran this, and this returns a `Tensor dtype shape` if the 4th argument is equal to `ReduceNone` and `Tensor dtype '[]` otherwise.

junjihashimoto: I have implemented a shape chaging due to Reduction on type-level.
How about this one?
https://github.com/hasktorch/hasktorch/pull/163/commits/bc583bfc1c574e303b4fb8912716ed8f11a64429#diff-ed5b0b77220a9f439e61aa413e33bdfdR227-R266

tscholak: This is awesome!
it's a bit odd that you have to use a type application instead of a value for `reduction` now, but I don't see how to do it otherwise.

hasktorch/src/Torch/Static/Native.hs
```diff
@@ -22,20 +22,1224 @@ import Data.Proxy
...
+-- >>> (dtype t,shape t)
+-- (Float,[])
+min :: Tensor dtype shape -> Tensor dtype '[]
+min t = unsafePerformIO $ (cast1 ATen.min_t) t
+
+-- |
+-- >>> t = max (ones :: Tensor Float '[2,2])
+-- >>> (dtype t,shape t)
+-- (Float,[])
+max :: Tensor dtype shape -> Tensor dtype '[]
```
apaszke: nit: I'd keep `max` for the element-wise operation and call the reduction operators `maxAll` and `maxDim`

junjihashimoto: Thank you for your feedback.

Add {max,min,median}Dim 
https://github.com/hasktorch/hasktorch/pull/163/files#diff-ed5b0b77220a9f439e61aa413e33bdfdR1257-R1270

PR #162, opened 2019-08-31
WIP: shell.nix that provides all the dependencies

wavewave:
To make hasktorch buildable simply.
```
$ nix-shell shell.nix
$ cabal new-build hasktorch
```

Note: this is not yet finalized since I disabled `mkl` and adjusted C++11 ABI flag. I would like to have some discussion about them.

PR #166, opened 2019-09-02
Drop -D_GLIBCXX_USE_CXX11_ABI=0

junjihashimoto:
libtorch begins to provide binaries using C++11ABI.
(Old C++ABI with  -D_GLIBCXX_USE_CXX11_ABI=0 is to keep backward compatibility for nvcc using gcc-4.)
https://discuss.pytorch.org/t/issues-linking-with-libtorch-c-11-abi/29510/2
https://github.com/hasktorch/hasktorch/pull/162

This PR changes the C++ABI to C++11ABI.


tscholak:
ah, now I understand why @wavewave did that ;)

PR #170, opened 2019-09-10
[draft] some more static tensor experiments

tscholak:


tscholak:
not needed anymore, closing...

PR #171, opened 2019-09-10
add nll_loss implementation

tscholak:
see https://pytorch.org/docs/stable/nn.functional.html?highlight=nll_loss#torch.nn.functional.nll_loss for the version in pytorch and https://pytorch.org/docs/stable/_modules/torch/nn/functional.html#nll_loss for the implementation.

there's no check so far on `target` or on `ignoreIndex`.
each element value in `target` has to fulfill `0 <= value < c`, where `c > 0` is the number of classes.
the purpose of `ignoreIndex` and it's constraints are poorly understood by me at this point.


junjihashimoto:
LGTM!

PR #172, opened 2019-09-10
improve broadcasting, support broadcasting for sub and mul

tscholak:


hasktorch/test/StaticSpec.hs
```diff
@@ -33,32 +33,68 @@ import Torch.Static.Native
...
+      let b = ones :: Tensor Float '[2, 3]
+      let c = add a b
+      checkDynamicTensorAttributes c
+      D.asValue (toDynamic c) `shouldBe` ([[2, 2, 2], [2, 2, 2]] :: [[Float]])
+    it "works on broadcastable tensors of different shapes" $ do
+      let a = ones :: Tensor Float '[3, 1, 4, 1]
+      let b = ones :: Tensor Float '[2, 1, 1]
+      let c = add a b
+      checkDynamicTensorAttributes c
+      D.asValue (toDynamic c) `shouldBe` ([[2, 2], [2, 2], [2, 2]] :: [[Float]])
```
junjihashimoto: I feel this is something storage.
> D.asValue (toDynamic c) `shouldBe` ([[2, 2], [2, 2], [2, 2]] :: [[Float]])
c's shape is  [3,2,4,1].

```
> add (ones' [3,1,4,1]) (ones' [2,1,1])
Tensor Float [3,2,4,1]
```

I think this becomes as follows.

```
D.asValue (toDynamic c) `shouldBe` ([[[[2.0],[2.0],[2.0],[2.0]],[[2.0],[2.0],[2.0],[2.0]]],[[[2.0],[2.0],[2.0],[2.0]],[[2.0],[2.0],[2.0],[2.0]]],[[[2.0],[2.0],[2.0],[2.0]],[[2.0],[2.0],[2.0],[2.0]]]] :: [[[[Float]]]])
 
```

tscholak: @junjihashimoto I agree, and I don't understand how this test could have passed. I just added the correct expectations while leaving the old (wrong ones). Tests are passing, so I don't understand what's going on.

it seems like `D.asValue (toDynamic c) :: [[Float]]` is different from `D.asValue (toDynamic c) :: [[[[Float]]]]` for the same `c`

junjihashimoto:
LGTM!
Thank you for updating the test.

hasktorch/test/StaticSpec.hs
junjihashimoto: We need the validation of asValue to check the shape.

PR #180, opened 2019-09-12
fix nix-build and nix-shell for libtorch-ffi

tscholak:


junjihashimoto:
nix-build works on linux.
Thx!

PR #183, opened 2019-09-14
type-level programming with DType

tscholak:
~this is becoming a bit of a larger experiment.~ **This is ready for review now!**

I needed to do some type-level programming with `DType`s for `logsumexp` and others.
This required the following change for statically typed tensors:
```
data Tensor (dtype :: DType.DType) (shape :: [Nat]) where
  UnsafeMkTensor :: forall dtype shape . { toDynamic :: D.Tensor } -> Tensor dtype shape
```
(here in GADT syntax).

However, this remains very much an experiment. This so far does not support reification of Haskell types as it has been accomplished with the `Reifies` type class.

~I also pushed version 0.7 of `ghc-typelits-knownnat` for the `KnownBool` typeclass.~ (this can wait)

~there is an unresolved issue: the return values for `all`, `any`, `all'`, `any'` are wrong, and this has nothing to do with the static typing I added.~ this has been resolved as of #184, thanks @junjihashimoto!

### DIFF
--- a/examples/static_xor_mlp/Main.hs
+++ b/examples/static_xor_mlp/Main.hs
@@ -20,6 +20,7 @@
 
 module Main where
 
+import Prelude hiding (tanh)
 import Torch.Static
 import Torch.Static.Native hiding (linear)
 import Torch.Static.Factories
@@ -51,10 +52,10 @@ instance A.Parameterized (Parameter d s) where
   flattenParameters (Parameter x) = [x]
   replaceOwnParameters _ = Parameter <$> A.nextParameter
 
-data LinearSpec d (i::Nat) (o::Nat) = LinearSpec
+data LinearSpec (d::DType) (i::Nat) (o::Nat) = LinearSpec
   deriving (Show, Eq)
 
-data Linear d (in_features::Nat) (out_features::Nat) =
+data Linear (d::DType) (in_features::Nat) (out_features::Nat) =
   Linear { weight :: Parameter d '[in_features,out_features]
          , bias :: Parameter d '[out_features]
          } deriving (Show, Generic)
@@ -67,15 +68,15 @@ makeIndependent t = Parameter <$> A.makeIndependent (toDynamic t)
 
 instance A.Parameterized (Linear d n m)
 
-instance (KnownNat n,KnownNat m,Reifies d DType) => A.Randomizable (LinearSpec d n m) (Linear d n m) where
+instance (KnownDType d, KnownNat n, KnownNat m) => A.Randomizable (LinearSpec d n m) (Linear d n m) where
   sample LinearSpec = do
       w <- makeIndependent =<< (randn :: IO (Tensor d '[n,m]))
       b <- makeIndependent =<< (randn :: IO (Tensor d '[m]))
       return $ Linear w b
 
-data MLPSpec d (i::Nat) (o::Nat) = MLPSpec
+data MLPSpec (d::DType) (i::Nat) (o::Nat) = MLPSpec
 
-data MLP d (i::Nat) (o::Nat) =
+data MLP (d::DType) (i::Nat) (o::Nat) =
   MLP { l0 :: Linear d i 20
       , l1 :: Linear d 20 20
       , l2 :: Linear d 20 o
@@ -83,7 +84,7 @@ data MLP d (i::Nat) (o::Nat) =
 
 instance A.Parameterized (MLP d i o)
 
-instance (KnownNat n, KnownNat m, Reifies d DType) => A.Randomizable (MLPSpec d n m) (MLP d n m) where
+instance (KnownDType d, KnownNat n, KnownNat m) => A.Randomizable (MLPSpec d n m) (MLP d n m) where
   sample MLPSpec = do
       l0 <- A.sample LinearSpec :: IO (Linear d n 20)
       l1 <- A.sample LinearSpec :: IO (Linear d 20 20)
@@ -91,21 +92,21 @@ instance (KnownNat n, KnownNat m, Reifies d DType) => A.Randomizable (MLPSpec d 
       return $ MLP {..}
 
 mlp :: MLP d i o -> Tensor d '[b,i] -> Tensor d '[b,o]
-mlp MLP{..} input = linear l2 $ linear l1 $ linear l0 input 
+mlp MLP{..} input = linear l2 . tanh . linear l1 . tanh . linear l0 $ input
 
 batch_size = 32
 num_iters = 10000
 
-model :: MLP Float 2 1 -> Tensor Float '[n,2] -> Tensor Float '[n,1]
+model :: MLP 'Float 2 1 -> Tensor 'Float '[n,2] -> Tensor 'Float '[n,1]
 model params t = sigmoid (mlp params t)
 
 main = do
-    init <- A.sample $ (MLPSpec :: MLPSpec Float 2 1)
+    init <- A.sample $ (MLPSpec :: MLPSpec 'Float 2 1)
     trained <- foldLoop init num_iters $ \state i -> do
         input <- D.rand' [batch_size, 2] >>= return . (D.toDType Float) . (D.gt 0.5)
         let expected_output = tensorXOR input
 
-        let output = D.squeezeAll $ toDynamic $ model state (UnsafeMkTensor input :: Tensor Float '[batch_size,2])
+        let output = D.squeezeAll $ toDynamic $ model state (UnsafeMkTensor input :: Tensor 'Float '[batch_size,2])
         let loss = D.mse_loss output expected_output
 
         let flat_parameters = A.flattenParameters state
--- a/hasktorch/src/Torch/Functions.hs
+++ b/hasktorch/src/Torch/Functions.hs
@@ -188,11 +188,11 @@ mse_loss a b = unsafePerformIO $ (cast3 ATen.mse_loss_ttl) a b ATen.kMean
 conv2d :: Tensor -> Tensor -> Tensor -> (Int, Int) -> (Int, Int) -> Tensor
 conv2d input weight bias (dh, dw) (ph, pw) = unsafePerformIO $
     (cast7 ATen.conv2d_tttllll) input weight bias
-                                [dh, dw] [ph, pw] ([1, 1] :: [Int]) (0 :: Int)
+                                ([dh, dw] :: [Int]) ([ph, pw] :: [Int]) ([1, 1] :: [Int]) (0 :: Int)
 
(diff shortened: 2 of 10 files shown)


hasktorch/src/Torch/Static/Native.hs
```diff
@@ -849,8 +961,73 @@ log _self = unsafePerformIO $ (cast1 ATen.log_t) _self
...
+                                      Text " not supported")
+
+type family IsIntegral (dtype :: DType) :: Constraint where
+  IsFloatingPoint 'Bool = ()
+  IsFloatingPoint 'UInt8 = ()
+  IsFloatingPoint 'Int8 = ()
+  IsFloatingPoint 'Int16 = ()
+  IsFloatingPoint 'Int32 = ()
+  IsFloatingPoint 'Int64 = ()
+  IsFloatingPoint dtype  = TypeError (Text "Data type " :<>:
```
junjihashimoto: Is this `IsIntegral`?

tscholak: thanks, I fixed this oversight!

junjihashimoto:
LGTM!

examples/static_xor_mlp/Main.hs
```diff
@@ -67,45 +68,45 @@ makeIndependent t = Parameter <$> A.makeIndependent (toDynamic t)
...
+instance (KnownDType d, KnownNat n, KnownNat m) => A.Randomizable (MLPSpec d n m) (MLP d n m) where
   sample MLPSpec = do
       l0 <- A.sample LinearSpec :: IO (Linear d n 20)
       l1 <- A.sample LinearSpec :: IO (Linear d 20 20)
       l2 <- A.sample LinearSpec :: IO (Linear d 20 m)
       return $ MLP {..}
 
 mlp :: MLP d i o -> Tensor d '[b,i] -> Tensor d '[b,o]
-mlp MLP{..} input = linear l2 $ linear l1 $ linear l0 input 
+mlp MLP{..} input = linear l2 . tanh . linear l1 . tanh . linear l0 $ input
```
tscholak: the composition of linear maps without nonlinearity is moot, hence this change

hasktorch/src/Torch/Functions.hs
```diff
@@ -188,11 +188,11 @@ mse_loss a b = unsafePerformIO $ (cast3 ATen.mse_loss_ttl) a b ATen.kMean
 conv2d :: Tensor -> Tensor -> Tensor -> (Int, Int) -> (Int, Int) -> Tensor
 conv2d input weight bias (dh, dw) (ph, pw) = unsafePerformIO $
     (cast7 ATen.conv2d_tttllll) input weight bias
-                                [dh, dw] [ph, pw] ([1, 1] :: [Int]) (0 :: Int)
+                                ([dh, dw] :: [Int]) ([ph, pw] :: [Int]) ([1, 1] :: [Int]) (0 :: Int)
```
tscholak: needed to make OverloadedLists extension happy in doctests

hasktorch/src/Torch/Static.hs
```diff
@@ -36,7 +38,7 @@ import qualified Torch.DType as DType
 natValI :: forall n. KnownNat n => Int
 natValI = fromIntegral $ natVal $ Proxy @n
 
-class (All KnownNat shape) => KnownShape shape where
+class KnownShape shape where
```
tscholak: not needed, does nothing

hasktorch/src/Torch/Static.hs
```diff
@@ -48,8 +50,68 @@ instance (KnownNat h, KnownShape t) => KnownShape (h ': t) where
...
+  ComputeDType DType.Int32 = DType.Int32
+  ComputeDType Int = DType.Int64
+  ComputeDType DType.Int64 = DType.Int64
+  ComputeDType Float = DType.Float
+  ComputeDType DType.Float = DType.Float
+  ComputeDType Double = DType.Double
+  ComputeDType DType.Double = DType.Double
+  ComputeDType dtype' = TypeError (Text "Unsupported tensor type " :<>: ShowType dtype')
+
+data Tensor (dtype :: DType.DType) (shape :: [Nat]) where
```
tscholak: the most important change: `dtype` has now kind `DType`!

hasktorch/src/Torch/Static.hs
```diff
@@ -48,8 +50,68 @@ instance (KnownNat h, KnownShape t) => KnownShape (h ': t) where
...
+  ComputeHaskellType DType.Float = Float
+  ComputeHaskellType DType.Double = Double
+  ComputeHaskellType dtype = TypeError (Text "Unsupported tensor type " :<>: ShowType dtype)
+  
+type family ComputeItemType (ty :: Type) (shape :: [Nat]) :: Type where
+  ComputeItemType _ '[] = TypeError (Text "Scalars are not supported")
+  ComputeItemType ty (_ ': '[]) = ty
+  ComputeItemType ty (_ ': h ': t) = [ComputeItemType ty (h ': t)]
+
+instance (D.TensorLike [ComputeItemType (ComputeHaskellType dtype) shape], KnownShape shape) => IsList (Maybe (Tensor dtype shape)) where
```
tscholak: this instance is for convenience: `[0,1] :: Maybe (Tensor ‘Int64 [2])`

hasktorch/src/Torch/Static/Native.hs
```diff
@@ -36,19 +38,14 @@ import qualified ATen.Type as ATen
...
+import qualified Torch.Scalar as D
 import Torch.Functions (Reduction(..), Tri(..), isUpper, kOne)
-import Torch.DType
 import Torch.Static
 import Torch.Static.Factories
-import Torch.Scalar
-import qualified Torch.TensorOptions as D
-import Data.Reflection
-import ATen.Cast
-import Control.Arrow ((&&&))
```
tscholak: I grouped the imports a bit to improve readability

hasktorch/src/Torch/Static/Native.hs
```diff
@@ -836,21 +955,58 @@ is_signed _self = unsafePerformIO $ (cast1 ATen.is_signed_t) _self
...
+--
+-- >>> dtype &&& shape &&& (\t' -> D.asValue (toDynamic t') :: [[Float]]) $ (logsumexp @1 @KeepDim t :: Tensor 'D.Float '[4, 1])
+-- (Float,([4,1],[[5.01815],[3.3132617],[4.0485873],[7.0067153]]))
+--
+-- >>> dtype &&& shape &&& (\t' -> D.asValue (toDynamic t') :: [Float]) $ (logsumexp @0 @DropDim t :: Tensor 'D.Float '[2])
+-- (Float,([2],[5.44019,7.0116277]))
+--
+-- >>> dtype &&& shape &&& (\t' -> D.asValue (toDynamic t') :: [[Float]]) $ (logsumexp @0 @KeepDim t :: Tensor 'D.Float '[1, 2])
+-- (Float,([1,2],[[5.44019,7.0116277]]))
+logsumexp
```
tscholak: logsumexp throws for integral tensors, hence the type-level constraint to floating point tensors

hasktorch/src/Torch/Static/Native.hs
```diff
@@ -1334,26 +1506,37 @@ type family DropValue (shape :: [Nat]) (i :: Nat) :: [Nat] where
...
 
 medianDim :: forall d dtype shape. (KnownNat d) => Tensor dtype shape -> Tensor dtype (DropValue shape d)
-medianDim _self = fst $ (unsafePerformIO $ (cast2 ATen.median_tl) _self (natValI @d) :: (Tensor dtype  (DropValue shape d),Tensor Int (DropValue shape d)))
+medianDim _self = fst $ (unsafePerformIO $ (cast2 ATen.median_tl) _self (natValI @d) :: (Tensor dtype  (DropValue shape d), Tensor 'D.Int64 (DropValue shape d)))
+
+-- | See https://pytorch.org/docs/stable/torch.html#torch.median.
+-- >>> t = fromJust [[5, 1], [3, 2], [4, 1], [2, 7]] :: Tensor 'D.Float '[4, 2]
+-- >>> median' @0 @KeepDim t :: (Tensor 'D.Float '[1, 2], Tensor 'D.Int64 '[1, 2])
+-- (Tensor Float [1,2] [[ 3.0000   ,  1.0000   ]],Tensor Int64 [1,2] [[ 1,  0]])
+median'
```
tscholak: this function was missing, so I added it

hasktorch/src/Torch/Tensor.hs
```diff
@@ -7,6 +7,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE IncoherentInstances #-}
```
tscholak: note to self: I should probably get rid of this

hasktorch/src/Torch/Tensor.hs
```diff
@@ -136,6 +137,7 @@ class TensorLike a where
   -- Internal functions(like "_xxx") are below. Do not use them directly.
   _dtype :: DType
   _dims :: a -> [Int]
+  _deepDims :: a -> Maybe [Int]
```
tscholak: safer check. makes sure that all lists have the same lengths

nix/ghc-typelits-knownnat.nix
```diff
@@ -0,0 +1,21 @@
+{ mkDerivation, base, ghc, ghc-prim, ghc-tcplugins-extra
+, ghc-typelits-natnormalise, stdenv, tasty, tasty-hunit, tasty-quickcheck
+, template-haskell, transformers
+}:
+mkDerivation {
+  pname = "ghc-typelits-knownnat";
+  version = "0.7";
```
tscholak: bump to 0.7. used only by nix build sa far

tscholak:
Thank you for reviewing and accepting this PR @junjihashimoto !

hasktorch/src/Torch/Static.hs
apaszke: I may be wrong, but IIUC this is necessary, because Haskell will happily accept types like `(1 - 5)` (`-` is simply not defined for those arguments) and will only raise an error if you ask that type to additionally satisfy `KnownNat`

tscholak: oh, I wasn’t aware of that. I just saw the instance hierarchy for `KnownShape`, and didn’t see a reason for the `All KnownNat` bound here. Since the everything built just fine without this bound, maybe we should add a test case that probes for simple `Nat` arithmetic at the type level and add this bound back. I’ll gladly make a PR for that! thanks for bringing this up :+1:

PR #184, opened 2019-09-16
Fix the mismatch of Bool-type between C++ and Haskell

junjihashimoto:
The size of Bool-type is different between Haskell and C++.
Haskell's bool is 4bytes.
```
> :m + Foreign.Storable
> sizeOf (True :: Bool)
4
```

C++'s bool is 1byte.
```
$ cat test.cpp
#include <iostream>
using namespace std;
int
main (){
  cout << sizeof(bool)  << endl;
  return 0;
}
$ g++ test.cpp ; ./a.out
```
Tensorlike-class uses the data-size of Haskell to store the value in libtorch's memory.

This PR fixes the mismatch.

### DIFF
--- a/hasktorch/src/Torch/Tensor.hs
+++ b/hasktorch/src/Torch/Tensor.hs
@@ -18,6 +18,7 @@ import Foreign.Storable
 import Foreign.C.Types
 import System.IO.Unsafe
 import Data.Int (Int64)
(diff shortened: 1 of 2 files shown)


austinvhuang:
Thanks for the fix @junjihashimoto !

PR #185, opened 2019-09-17
Optimizers as a standalone example

austinvhuang:
An example sketch with optimizer implementations, gd, gd + momentum, adam, run against some simple test functions (2d rosenbrock, convex quadratic).

Runs as a standalone example, but the idea is to start working on an optimizers interface. This is a step towards getting that going though there are remaining questions regarding the interface / api design that need to be explored further in transitioning from toy example to library functions.

Might also add some hvega plots in the future.


examples/optimizers/Main.hs
```diff
@@ -0,0 +1,226 @@
...
+-- | show output after n iterations
+showLog :: (Show a) => Int -> Int -> Tensor -> a -> IO ()
+showLog n i lossValue state = 
+    when (mod i n == 0) do
+        putStrLn ("Iter: " ++ printf "%4d" i 
+            ++ " | Loss:" ++ printf "%.4f" (asValue lossValue :: Float)
+            ++ " | Parameters: " ++ show state)
+
+-- | produce flattened parameters and gradient for a single iteration
+runIter state loss i = do
```
tscholak: top level definition without type annotations, suggestion:
```suggestion
runIter
  :: (Show f, Parameterized f) =>
     f -> (f -> Tensor) -> Int -> IO ([Parameter], [Tensor])
runIter state loss i = do
```

examples/optimizers/Main.hs
```diff
@@ -0,0 +1,226 @@
...
+-- gradient descent step
+gd :: LearningRate -> [Parameter] -> Gradient -> [Tensor]
+gd lr parameters gradients = zipWith step depParameters gradients
+  where
+    step p dp = p - (lr * dp)
+    depParameters = fmap toDependent parameters
+
+-- gradient descent with momentum step
+--     lr              beta      memory      parameters     gradients
+gdm :: LearningRate -> Float -> [Tensor] -> [Parameter] -> Gradient -> [(Tensor, Tensor)]
```
tscholak: this seems to be a better way to document the arguments:
```suggestion
gdm
  :: LearningRate -- ^ learning rate
  -> Float -- ^ beta
  -> [Tensor] -- ^ memory
  -> [Parameter] -- ^ parameters
  -> Gradient -- ^ gradients
  -> [(Tensor, Tensor)]
```
cf. https://haskell-haddock.readthedocs.io/en/latest/markup.html#constructors-and-record-fields

examples/optimizers/Main.hs
```diff
@@ -0,0 +1,226 @@
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE BlockArguments #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Main where
```
tscholak: any particular reason why this isn't `Torch.Optimizers`?

examples/optimizers/Main.hs
```diff
@@ -0,0 +1,226 @@
...
+        lossValue = loss state
+        flatParameters = flattenParameters state
+        gradients = grad lossValue flatParameters
+
+-- | foldM as a loop with action block as the last argument
+foldLoop :: Monad m => a -> Int -> (a -> Int -> m a) -> m a
+foldLoop x count block = foldM block x [1..count]
+
+-- | Rosenbrock function gradient descent
+testRosenGD :: Int -> IO ()
```
tscholak: could this be translated into a unit test?
this doesn't seem to be deterministic:
```
*TestFunctions Main> testRosenGD 10
Initial :(0.32480815,1.086431)
*TestFunctions Main> testRosenGD 10
Initial :(0.6006327,0.91784304)
```

examples/optimizers/Main.hs
tscholak: I read the description of your PR again, and I think I understand why you want to have it separate at this point

examples/optimizers/Main.hs
austinvhuang: Definitely can. The non-determinism comes from the initialization being random. Would be fairly straightforward to incorporate these into a unit test (though we might hold off on extensive testing of this sort for the library version of the API to take shape).

hasktorch/src/Torch/Functions.hs
```diff
@@ -90,9 +90,18 @@ sub a b = unsafePerformIO $ (cast3 ATen.sub_tts) a b kOne
 mul :: Tensor -> Tensor -> Tensor
 mul a b = unsafePerformIO $ (cast2 ATen.mul_tt) a b
 
+cadd :: Scalar a => Tensor -> a -> Tensor
+cadd t a = unsafePerformIO $ (cast2 ATen.add_ts) t a
```
apaszke: I'd rather avoid carrying over those weird names from TH. Why not `addScalar`?

apaszke: Alternatively why don't we make the signature of `add` this:
```hs
class TensorLike a where
  asTensorLike :: Tensor -> a -> Tensor -- first tensor used only for dtype/device

add :: TensorLike a => Tensor -> a -> Tensor
```

examples/optimizers/Main.hs
apaszke: It might be also interested to explore some [avenues that JAX took](https://github.com/google/jax/blob/master/design_notes/prng.md). Their approach to non-determinism is heavily functional.

examples/optimizers/Main.hs
austinvhuang: Using this convention now in `Optimizers.hs`

examples/optimizers/Main.hs
austinvhuang: Added

hasktorch/src/Torch/Functions.hs
austinvhuang: Changed the library functions to use descriptive names, also backported some of the other examples to use descriptive function names.

junjihashimoto:
Thank you for great work.

PR #189, opened 2019-09-23
clean up and improve static xor example

tscholak:
- [x] add broadcast support to the comparison ops with tests
- [x] clean up the static xor example code
- [x] use more type-safe ops in the example
- [x] make the model much smaller and tune hyper-parameters like batch size and learning rate

### DIFF
--- a/examples/static_xor_mlp/Main.hs
+++ b/examples/static_xor_mlp/Main.hs
@@ -16,6 +16,7 @@
 {-# LANGUAGE NoStarIsType #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE OverloadedLists #-}
 
 module Main where
 
@@ -85,7 +86,7 @@ data MLP (d::D.DType) (i::Nat) (o::Nat) (h::Nat) =
   MLP { l0 :: Linear d i h
       , l1 :: Linear d h h
       , l2 :: Linear d h o
-      } deriving (Generic)
+      } deriving (Show, Generic)
 
 instance A.Parameterized (MLP d i o h)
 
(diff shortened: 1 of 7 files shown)


hasktorch/src/Torch/Static/Native.hs
```diff
@@ -194,47 +189,11 @@ sqrt t = unsafePerformIO $ (cast1 ATen.sqrt_t) t
...
-ne a b = unsafePerformIO $ (cast2 ATen.ne_tt) a b
-
-(/=.) = ne
-
 -- |
 -- >>> dtype &&& shape $ (toDType (ones :: Tensor 'D.Float '[2,2]) :: Tensor 'D.Double '[2,2])
 -- (Double,[2,2])
 toDType
-  :: forall dtype dtype' shape
+  :: forall dtype' dtype shape
```
tscholak: it's true for type arguments, too:
https://twitter.com/rob_rix/status/1171495487267991555?s=20

examples/static_xor_mlp/Main.hs
```diff
@@ -70,58 +76,66 @@ instance A.Parameterized (Linear d n m)
...
-    foldLoop x count block = foldM block x [1..count]
-    tensorXOR t = (1 - (1 - a) * (1 - b)) * (1 - (a * b))
-      where
-        a = D.select t 1 0
-        b = D.select t 1 1
+  let numIters = 100000
+  init    <- A.sample (MLPSpec :: MLPSpec 'D.Float 2 1 4)
+  trained <- foldLoop init numIters $ \state i -> do
+    input <-
+      toDType @D.Float
```
tscholak: note the change in order of the type arguments of `toDType` to make this work

junjihashimoto:
LGTM!

tscholak:
@junjihashimoto thanks for your review. I improved the code even further such that it can live up to its purpose of being an example for newcomers :)

PR #193, opened 2019-10-01
Transformer Language Model! Part Deux

tscholak:
- [x] infix notation for HLists
- [x] layer norm
- [x] embeddings
- [x] tensor expand
- [x] unsqueeze
- [x] masked fill
- [x] triu and tril
- [ ] tbd

### DIFF
--- a/examples/static-transformer/Main.hs
+++ b/examples/static-transformer/Main.hs
@@ -189,7 +189,7 @@ getHidden
   -> IO (Tensor dtype '[batchSize, seqLen])
 getHidden embedding posEmbedding input = do
   let srcTokens = transpose @0 @1 input
-      positions = expand @'[batchSize, seqLen, embedDim] True . embed posEmbedding $ (_ :: Tensor 'D.Int64 '[seqLen])
+      positions = expand @'[seqLen, batchSize, embedDim] True . unsqueeze @1 . embed posEmbedding $ (_ :: Tensor 'D.Int64 '[seqLen])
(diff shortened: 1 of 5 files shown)


tscholak:
@junjihashimoto @austinvhuang I wanted to split this massive file up first, but I’m busy with MNIST right now, so how about we merge this in for the ops I implemented?
the transformer is still incomplete: model is done (in two different ways, with and without arrows), but there’s no way of training it yet. hence, there will be a third chunk of work to be done on this. I’ll probably get to it on the weekend. I just don’t want to delay any further.

examples/static-transformer/Main.hs
```diff
@@ -116,23 +187,361 @@ data Dropout where
...
+       , Tensor dtype '[seqLen, batchSize, embedDim]
+       )
+       (Tensor dtype '[seqLen, batchSize, embedDim])
+transformerLMLayer' TransformerLMLayer {..} =
+  (arr snd &&& attn)
+    >>> uncurry add
+    ^>> Main.layerNorm tLN0
+    ^>> (id &&& transformerLMMLP' tMLP)
+    >>^ uncurry add
+    >>^ Main.layerNorm tLN1
```
apaszke: While this is a cute notation, I think that if we really want this library to become successfull then we should be very cautious about using a hundred different operators and `Kleisli`, because it's certainly not super beginner friendly.

tscholak: I agree! this was meant only as an illustration

PR #196, opened 2019-10-07
Feature/mnist

junjihashimoto:
+ Recover mnist-example


examples/mnist/Main.hs
```diff
@@ -0,0 +1,159 @@
...
+    _ -> toCPU t
+
+
+
+main = do
+  backend' <- try (getEnv "BACKEND") :: IO (Either SomeException String)
+  let backend = 
+        case backend' of
+          Right "CUDA" -> "CUDA"
+          _ -> "CPU"
```
tscholak: In Pytorch, the best way to deal with this is through devices rather than backends. That lets you deal with multi-cpu and multi-gpu sitatuations.

examples/mnist/Main.hs
```diff
@@ -0,0 +1,159 @@
...
+  -> Tensor dtype '[batchSize, inputFeatures]
+  -> Tensor dtype '[batchSize, outputFeatures]
+model = (sigmoid .) . mlp
+
+foldLoop
+  :: forall a b m . (Num a, Enum a, Monad m) => b -> a -> (b -> a -> m b) -> m b
+foldLoop x count block = foldM block x ([1 .. count] :: [a])
+
+type BatchSize = 64
+type TestSize = 10000
```
tscholak: nice! we could also use `someNat` to read this in from the command line.

examples/mnist/Image.hs
```diff
@@ -0,0 +1,76 @@
...
+       1 -> [1,0,0,0,0,0,0,0,0,0]
+       2 -> [0,1,0,0,0,0,0,0,0,0]
+       3 -> [0,0,1,0,0,0,0,0,0,0]
+       4 -> [0,0,0,1,0,0,0,0,0,0]
+       5 -> [0,0,0,0,1,0,0,0,0,0]
+       6 -> [0,0,0,0,0,1,0,0,0,0]
+       7 -> [0,0,0,0,0,0,1,0,0,0]
+       8 -> [0,0,0,0,0,0,0,1,0,0]
+       9 -> [0,0,0,0,0,0,0,0,1,0]
+       _ -> [0,0,0,0,0,0,0,0,0,1]
```
tscholak: yuck ;)

austinvhuang:
Purely superfluous bikeshedding in case it's of interest (don't feel obligated to incorporate) - for my own use in pytorch kludged this function for displaying data in the terminal a while back :)

```
def render_tensor(img: torch.Tensor, scale=1):
    # render images in a terminal
    # http://paulbourke.net/dataformats/asciiart/
    intensity_map = " .:-=+*#%@"
    new_size = (torch.tensor(img.shape).double() * scale).int()
    img = F.interpolate(img.view(1, 1, img.shape[0], img.shape[1]),
                        scale_factor=scale) \
                                .view(new_size[0], new_size[1])
    norm = torch.floor((img - img.min()) / (img.max() - img.min()) * 9)
    render = ""
    for row in range(norm.shape[0]):
        for col in range(norm.shape[1]):
            render += intensity_map[int(norm[row, col])]
        render += "\n" 
    print(render)
```

examples/mnist/Main.hs
```diff
@@ -0,0 +1,159 @@
...
+  :: MLP dtype inputFeatures outputFeatures hiddenFeatures
+  -> Tensor dtype '[batchSize, inputFeatures]
+  -> Tensor dtype '[batchSize, outputFeatures]
+mlp MLP {..} = linear layer2 . relu . linear layer1 . relu . linear layer0
+
+model
+  :: MLP dtype inputFeatures outputFeatures hiddenFeatures
+  -> Tensor dtype '[batchSize, inputFeatures]
+  -> Tensor dtype '[batchSize, outputFeatures]
+model = (sigmoid .) . mlp
```
austinvhuang: Given how trivial this is, I'd be in favor of merging mlp into model.

examples/mnist/Main.hs
```diff
@@ -0,0 +1,159 @@
...
+
+makeIndependent :: Tensor dtype shape -> IO (Parameter dtype shape)
+makeIndependent t = Parameter <$> A.makeIndependent (toDynamic t)
+
+instance A.Parameterized (Linear dtype inputFeatures outputFeatures)
+
+instance (KnownDType dtype, KnownNat inputFeatures, KnownNat outputFeatures) => A.Randomizable (LinearSpec dtype inputFeatures outputFeatures) (Linear dtype inputFeatures outputFeatures) where
+  sample LinearSpec =
+    Linear <$> (makeIndependent =<< randn) <*> (makeIndependent =<< randn)
+
```
austinvhuang: Lines 47-79 seem like the should go into the static version of NN in the library instead of being here, though I'd be fine with merging first and saving the library merge  for a subsequent PR.

examples/mnist/Main.hs
```diff
@@ -0,0 +1,159 @@
...
+{-# LANGUAGE UndecidableSuperClasses #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE NoStarIsType #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE OverloadedLists #-}
+
```
austinvhuang: Wow that's a lot, are all of these used?

examples/mnist/Main.hs
```diff
@@ -0,0 +1,159 @@
...
+    let (indexes,nextIndexes) = (take (natValI @I.DataDim) idxs, drop (natValI @I.DataDim) idxs)
+    let input           = toBackend backend $ I.getImages @BatchSize trainingData indexes 
+    let expected_output = toBackend backend $ I.getLabels @BatchSize trainingData indexes
+    let actual_output   = model state $ input
+    let loss            = mse_loss actual_output expected_output
+
+    let flat_parameters = A.flattenParameters state
+    let gradients       = A.grad (toDynamic loss) flat_parameters
+
+    when (i `mod` 250 == 0) (print loss)
```
austinvhuang: I like `when`, we should do the same for other examples (not required for this PR though) where we use `if/then/else` with else being a dummy return ().

examples/mnist/Main.hs
austinvhuang: standard thing to do is to use log softmax rather than sigmoid for multiclass classification, see the pytorch reference example https://github.com/pytorch/examples/blob/master/mnist/main.py

examples/mnist/Main.hs
```diff
@@ -0,0 +1,159 @@
...
+
+    new_flat_parameters <- mapM A.makeIndependent
+      $ A.sgd 1e-1 flat_parameters gradients
+    return $ (A.replaceParameters state new_flat_parameters, nextIndexes)
+
+  putStrLn $ "Learning loss:"
+  let test_data = toBackend backend $ I.getImages @TestSize testData [0..]
+      test_label = toBackend backend $ I.getLabels @TestSize testData [0..]
+  print $ mse_loss (model trained test_data) test_label
```
austinvhuang: while it'll probably still work with mse, nll_loss is probably more appropriate for the classification task https://github.com/pytorch/examples/blob/ee964a2eeb41e1712fe719b83645c79bcbd0ba1a/mnist/main.py#L34

examples/mnist/Main.hs
tscholak: @austinvhuang @junjihashimoto you are completely right, I’m not sure why I didn’t see this issue when I looked at the code. Sigmoid won’t work here, it has to be a softmax.

examples/mnist/Main.hs
tscholak: Likewise, mse is not a good choice, either. Luckily, typed negative log likelihood is already implemented 😅

junjihashimoto:
@tscholak @austinvhuang 
Thank you for reviewing.

This code occurs  a runtime error as follows.

```
[nix-shell:~/git/hasktorch/examples]$  ./dist/build/mnist/mnist
mnist: CppStdException "Exception: Differentiated tensor has more than a single element; type: std::runtime_error"
```

austinvhuang:
> @tscholak @austinvhuang
> Thank you for reviewing.
> 
> This code occurs a runtime error as follows.
> 
> ```
> [nix-shell:~/git/hasktorch/examples]$  ./dist/build/mnist/mnist
> mnist: CppStdException "Exception: Differentiated tensor has more than a single element; type: std::runtime_error"
> ```

Headed out, but I would double check what's the underlying shape of the `loss` variable on line 137. The type of the `nll_loss'` function looks right, but maybe there's a bug underlying tensor doesn't reflect the signature.

examples/mnist/Main.hs
tscholak: let's postpone that to later. these lines appear in a number of examples. We will need to revisit this again for type-safe gradients, anyway.

examples/mnist/Main.hs
tscholak: this appears as such also in the static-mlp example

tscholak:
@junjihashimoto one more thing: can you rename this example to `static-mnist`?

tscholak:
hmmm, so I just ran the training on my laptop. it's still going (slowly), but I can already tell that the results won't be good:
```
$ ../dist/build/mnist/mnist
Tensor Float [] -3.1250e-2
Tensor Float [] -9.3750e-2
Tensor Float [] -0.1094
Tensor Float []  0.0000
Tensor Float [] -4.6875e-2
Tensor Float [] -0.1094
Tensor Float [] -3.1250e-2
...
```
the loss is jumping around and doesn't converge.

few things to tweak:
- learning rate (probably too high)
- model capacity (probably too low)
- model architecture (cnn would be a much better fit than ffn)
- I don't know if this network is learning anything because we don't report validation loss.
- I don't know if this model is overfitting or not because we don't report validation loss.

tscholak:
there's also this table on the wikipedia: https://en.wikipedia.org/wiki/MNIST_database#Classifiers

if you don't mind, I'll iterate a bit today on the model to make the machine learning part perform better. I'll post a PR against your branch later.

tscholak:
oops, I meant to type `git push origin`, but I typed `git push`, and somehow git let me push into your branch.

tscholak:
I tried to move the model to CUDA, but the naive way in which I did it doesn't work:
```

**Comment by junjihashimoto** [comment]
@tscholak 
The push is fine.
Thank you for great help.
I will check the model on CUDA.

**Comment by junjihashimoto** [comment]
https://colab.research.google.com/drive/1T7jSAIEQliOL-2CgX9sx0IYbvSRoyAQN

```
/bin/bash: line 1: 21110 Segmentation fault      (core dumped) BACKEND=CUDA /content/hasktorch/dist-newstyle/build/x86_64-linux/ghc-8.6.5/examples-0.2.0.0/x/static-mnist/build/static-mnist/static-mnist
139
```

tscholak:
@junjihashimoto rebase my local branch from yours, currently staring at the segfault. no idea what's going on ¯\_(ツ)_/¯

junjihashimoto:
By printf debug, test loss seems to be a problem.

tscholak:
In debugging this I found that the type signature of `Torch.Static.Native.sumAll` is totally wrong, I'm going to push a fix in a few minutes.

junjihashimoto:
Actually I noticed it yesterday.

junjihashimoto:
thx!

tscholak:
@junjihashimoto no problem, was an easy fix! Pushed it just now!

tscholak:
@junjihashimoto I think the last commit fixed it!

tscholak:
running training now normally on the GPU. however, the GPU utilization is terrible. the whole thing seems to be CPU-bound.

junjihashimoto:
Cloud you run ```perf top```? (I have to get actual machine with GPU to debug it.)

junjihashimoto:
@tscholak 
Awesome work!
I think it can be merged, and we can improve performance later .

PR #200, opened 2019-10-10
Merge static typed Parameter

junjihashimoto:
* Remove duplicated static typed parameter.


junjihashimoto:
thx!

hasktorch/src/Torch/Static/NN.hs
```diff
@@ -0,0 +1,75 @@
...
+                      (ComputeBroadcast
+                         (ReverseImpl
+                            (CheckMatMul
+                               shape
+                               '[inputFeatures, outputFeatures]
+                               (ComputeMatMul
+                                  (ReverseImpl shape '[]) '[outputFeatures, inputFeatures]))
+                            '[])
+                         '[outputFeatures])
+                    ~ shape')
```
apaszke: All of this should be possible to infer from the body. Why not use partial type signatures??

PR #202, opened 2019-10-12
Support for libtorch-1.3

junjihashimoto:
- [x] Add Dimname and DimnameList of new data-types 
- [x] Fix compilation-error
- [x] Add Symbol of new data-type
- [x] Upload binaries of libtorch-1.3
- [x] Refine auto generated functions.


libtorch-ffi/libtorch-ffi.cabal
```diff
@@ -74,11 +78,11 @@ library
  extra-ghci-libraries: stdc++
  if os(darwin)
   ld-options: -Wl,-keep_dwarf_unwind
-  ghc-options:       -optc-std=c++11 -optc-xc++
+  ghc-options:       -optc-std=c++17 -optc-xc++
```
junjihashimoto: I should revert this option.

austinvhuang:
Thanks for getting this going @junjihashimoto! Let us know if there’s aspects others can help with. We may need to tweak some of the examples and higher level apis depending on how the ffi gets updated.

junjihashimoto:
@austinvhuang 
Thank you for your kind offer.  Surprisingly  some of the examples and higher level api could compile and test without change.  I think I should tweak binding.yaml deciding which ffi-functions for dynamic tensors to make.

tscholak:
hi @junjihashimoto! `Dimname` and `DimnameList` are libtorch primitives for Named Tensors, are they? I'm actually surprised that Named Tensors have been implemented in libtorch. I was almost sure that it was implemented as a wrapper in Python. Cool!

junjihashimoto:
@tscholak 
Yes. The usage is here. 
https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/test/NamedTensor_test.cpp
I'm looking forward to seeing it on high level api.

austinvhuang:
@junjihashimoto @tscholak from what I understand the trajectory for libtorch is basically feature parity with the python api, which is great in terms of having as much functionality as possible.

There might be some features where we'll have to make design decisions in terms of how they're interfaced though (or even to have high level interfaces at all). For namedtensor, it's not obvious to me what it looks like yet. But we can go ahead with the codegen bindings first and consider options for the functionality.

tscholak:
@austinvhuang Oh, yes, for sure we can go ahead without figuring out named tensors! I was just surprised to find any evidence of named tensors in libtorch.

junjihashimoto:
@tscholak @austinvhuang 
All tasks for upgrading libtorch are done.
I've found tensor-factory of named-tensor does not work. I don't know the cause.
But  existing examples are working, for now I think this pr can be merge.

tscholak:
sweet! thanks @junjihashimoto, what an achievement!
I'm not the right person to review the codegen part, but I trust in the tests we wrote in Torch.Static.

hasktorch/test/SparseSpec.hs
```diff
@@ -25,13 +25,13 @@ spec = do
     (shape (asTensor v)) `shouldBe` [3]
     print (toDense x)
     -- When we call print for sparse tensor, it throws a exception.
-    (print x) `shouldThrow` anyException
+    print x -- `shouldThrow` anyException
```
tscholak: doesn't throw anymore? did they implement something sparse-related that wasn't implemented before?

junjihashimoto: No. Our understanding is right. They changed the implementation, but I do not find the change.
We can print out sparse-tensor.

austinvhuang:
Wow @junjihashimoto this is amazing. Considering how long it took to catch up to the last PTDC release and now it's down to a few days.

Any idea what's going on with the OSX Circle CI build? Should CircleCI just be deprecated entirely?

If CI passes, I'm good with going ahead and merging.

junjihashimoto:
@austinvhuang 
Thank you for reviewing.
We should deprecate circleci.
CircleCI allows ssh (it is useful to debug.), but the resource is no so much. 
It is 1x concurrency,  4GB memory and short timeout on macos.
Sometimes, we waste time to wait for CI on macos.

PR #211, opened 2019-10-17
Add toType and fast image-loader

junjihashimoto:
Current image loader using asTensor is slow, because it creates objects of every pixels values.
The profile of static-mnist-mlp is below.
image
The bottleneck is peekElemOff.

This PR uses memcpy to load image directly.
The profile using memcpy is below.  CPU usage improves.
image


tscholak:
Awesome!!! Thanks for figuring this out, I couldn’t have done it. The solution isn’t obvious to me, could you please add a test that the tensors returned by getImages and getImages’ are identical in value?

junjihashimoto:
Sure!

tscholak:
I just realize that it could be hard to implement this consistency check because it would depend on the downloaded mnist data. could we fake the data somehow for the purpose of the test?

junjihashimoto:
@tscholak Thank you for your reviewing.

PR #215, opened 2019-10-18
Add hvega for static-mnist-mlp

junjihashimoto:
This pr shows  a graph of loss-function.
I think I should use new module to show it.

image

### DIFF
--- a/examples/examples.cabal
+++ b/examples/examples.cabal
@@ -43,10 +43,12 @@ library static-mnist
   hs-source-dirs:      static-mnist
   exposed-modules:     Image
                      , Monitoring
+                     , Common
   build-depends:       bytestring >= 0.10.8
                      , zlib >= 0.6
                      , libtorch-ffi
                      , hvega
+                     , random >= 1.1
 
 test-suite spec
   import:             config
--- a/examples/examples.cabal
+++ b/examples/examples.cabal
@@ -78,6 +78,7 @@ executable static-mnist-mlp
                      , random >= 1.1
                      , safe-exceptions
                      , static-mnist
+                     , hvega
 
 executable static-transformer
   import:              config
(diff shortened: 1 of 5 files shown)


austinvhuang:
Yes! Am a fan of vega-lite in general. Few notes thoughts on visualization integrations (don't have to cover these in this PR).

For small image data like mnist, can use something like rect to visualize examples https://vega.github.io/vega-lite/docs/rect.html

For tracking losses and other common visualizations that don't require a full-blown plotting framework, it's common to use tensorboard (and maybe visdom which is not quite as popular). The nice thing about these is that they're run as decoupled services that run realtime. I haven't looked into this space in a while, but there have been various ways to get API integration with tensorboard from outside tensorflow.

Once upon a time I prototyped a haskell client to crayon https://github.com/austinvhuang/crayon-hs/blob/master/src/API.hs which was a restful wrapper around tensorboard for such interop. Crayon is now completely deprecated, but I'm sure there are other API integration points now to do something similar.

tscholak:
I also recommend integration with tensorboard.
However, in the absence of such integration, vega lite plots are great

junjihashimoto:
@tscholak @austinvhuang
Thank you for your reviewing.
 
 tensorboard is nice.
It seems pytorch and tensorboardx uses protocol-buffers directly.
https://github.com/pytorch/pytorch/blob/master/torch/utils/tensorboard/writer.py
https://github.com/pytorch/pytorch/blob/master/torch/utils/tensorboard/summary.py
https://github.com/lanpa/tensorboardX/blob/master/tensorboardX/proto/summary_pb2.py

If we use tensorboard, 
we may have to develop tensorboard' api for haskell, or use tensorflow-logging.

austinvhuang:
Tensorboard integration is only a “nice to have” and not essential at this stage considering there’s more foundational things we’re pushing on. I think the more flexible, lighter weight use of Vega is definitely the right approach for now.

tscholak:
I concur. Btw, google may have already worked on that for tensorflow-haskell, see https://github.com/tensorflow/haskell/blob/master/tensorflow-logging/src/TensorFlow/Logging.hs

PR #216, opened 2019-10-18
Enable profile by default

junjihashimoto:
hasktorch is development phase, so we are going to take profile many times.
This pr enables profile by default.


nix/shared.nix
```diff
@@ -28,6 +28,7 @@ let
                 overrideCabal = pkgsNew.haskell.lib.overrideCabal;
                 optionalString = pkgsNew.stdenv.lib.optionalString;
                 isDarwin = pkgsNew.stdenv.isDarwin;
+                enableLibraryProfiling = true;
```
tscholak: I don't think this has any effect. use this:
```suggestion
                enableLibraryProfiling = pkgsNew.stdenv.lib.enableLibraryProfiling;
```
(see https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/lib.nix#L168) and then wrap the derivations you want to enable profiling for in that function like I did with `failOnAllWarnings`.

junjihashimoto: Thx! That helps a lot.

tscholak:
Hi @junjihashimoto, your last commit, https://github.com/hasktorch/hasktorch/pull/216/commits/b725b9f868df79cde8e18109ccd8e599bc1567d7, only added the `enableLibraryProfiling` to the scope, but you still have to use it.

tscholak:
for instance, instead of 
```nix
                    "hasktorch_${postfix}" =
                      overrideCabal
                        (haskellPackagesOld.callCabal2nix
                          "hasktorch"
                          ../hasktorch
                          { libtorch-ffi = haskellPackagesNew."libtorch-ffi_${postfix}"; }
                        )
                        (old: {
                              preConfigure = (old.preConfigure or "") + optionalString (!isDarwin) ''
                                export LD_PRELOAD=${pkgs.mkl}/lib/libmkl_rt.so
                              '';
                            }
                        );
```
write
```nix
                    "hasktorch_${postfix}" =
                      enableLibraryProfiling
                        (overrideCabal
                          (haskellPackagesOld.callCabal2nix
                            "hasktorch"
                            ../hasktorch
                            { libtorch-ffi = haskellPackagesNew."libtorch-ffi_${postfix}"; }
                          )
                          (old: {
                                preConfigure = (old.preConfigure or "") + optionalString (!isDarwin) ''
                                  export LD_PRELOAD=${pkgs.mkl}/lib/libmkl_rt.so
                                '';
                              }
                          )
                        );
```

junjihashimoto:
@tscholak 
Thank you for pointing it out.

BTW, the failure of CI on stack-macos is the same as this issue(https://github.com/fpco/inline-c/issues/102).

tscholak:
@junjihashimoto, in order to fix this nasty macOS bug, I introduced this hot-fix: https://github.com/hasktorch/hasktorch/blob/b725b9f868df79cde8e18109ccd8e599bc1567d7/nix/shared.nix#L46-L48
I'm surprised it's not working any longer

tscholak:
@junjihashimoto it’s super hard to make both cabal and stack happy about this cx flag on all platforms. That’s why I went for this hot-fix solution in shared.nix

junjihashimoto:
@tscholak 
I can not find options to pass stack-macos.
This commit is the same as nix's one. But stack fails.
https://github.com/hasktorch/hasktorch/pull/216/commits/0480daba3315bb8a606590ccff032fb4c8ba8b78

junjihashimoto:
@tscholak 
I've realized setting profile on macos is super hard.
ghc passes options of "-DPROFILING -x c" to clang in front of user's options(-xc++ -std=c++11).
clang uses the option of "clang -xc++ -std=c++11 -DPROFILING -x c <generated code>."
So clang recognizes generated codes of inline-c-cpp as c-sources.
I am finding the source-codes adding the options in ghc-repo.

junjihashimoto:
We can get the result of benchmark from artifact of github-action.
image

tscholak:
sweet, @junjihashimoto !

tscholak:
I can try to fix the nix-macos build later today if it should still fail

junjihashimoto:
nix-macos fails during setup of nix.
I share the error with cachix-action.
https://github.com/cachix/cachix-action/issues/14

tscholak:
O.o, ok. I'll follow this one.

PR #217, opened 2019-10-18
Rename Static to Typed

junjihashimoto:
When people hear static on deep learning context, they think of static graph.
hasktorch does not create static graph. 
static of hastorch means statically typed data.
This pr renames static to typed, because it is confusing.


junjihashimoto:
@tscholak Thank you for this proposal and reviewing.

PR #218, opened 2019-10-19
type-safe device

tscholak:
The tests pass, time to make this a proper PR.

I'm very sorry that this is thing ended up being so massive.
I will assemble a PR description with links to interesting code bits soon.


hasktorch/src/Torch/Typed/Device.hs
```diff
@@ -0,0 +1,14 @@
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
+
+module Torch.Typed.Device where
+
+import           GHC.TypeLits
+
+data DeviceType = CPU | CUDA
```
apaszke: One thing we should consider is that at some point in the future PyTorch will support registration of new devices at run-time, and so having a closed ADT might not be a good solution. I don't know how to do it differently though, since you cannot have kind-classes (type-level type-classes).

tscholak: hmm, so I thought of that. And I think that defining all possible device types in line 9 is fine and necessary. same with device indices. I'm using `Nat` for that, and hence the number of devices is not restricted at compile time.
that means that functions like `toCUDA` will have to return an option of a moved tensor.

tscholak: oh, wait, I think I misunderstood you.
are you saying that future torch versions will support adding device runtimes at runtime?
in that case, defining all possible device types in line 9 would be impossible.

apaszke: Yes, that's what I meant. I don't think hot-swapping GPUs is that important, but being open to new backends is (you're e.g. already missing TPU).

examples/static-transformer/Main.hs
```diff
@@ -62,53 +65,64 @@ import qualified Torch.TensorFactories         as D
...
      , Mod (embedDim * 3) 3 ~ 0
      , Div (embedDim * 3) 3 ~ embedDim
-     , KnownDType dtype
      , All KnownNat [embedDim, numHeads, seqLen, batchSize, headDim]
+     , KnownDType dtype
+     , dtype ~ DTypePromotion dtype (SumDType dtype)
+     , StandardFloatingPointDTypeValidation device dtype
+     , MatMulDTypeIsValid device dtype
+     , BasicArithmeticDTypeIsValid device dtype
+     , BasicArithmeticDTypeIsValid device (SumDType dtype)
```
junjihashimoto: Where is BasicArithmeticDTypeIsValid 's definition?

tscholak: there's a new module called `Torch.Typed.Aux` where I put most of these definitions.
This particular one is defined in `Torch.Typed.Tensor`, though: https://github.com/tscholak/hasktorch/blob/feature/type-safe-device/hasktorch/src/Torch/Typed/Tensor.hs#L421-L426

hasktorch/src/Torch/Typed/NN.hs
```diff
@@ -260,45 +286,47 @@ data Conv2d (dtype :: D.DType)
 conv2d
   :: forall stride padding
    . _
-  => Conv2d _ _ _ _ _
-  -> Tensor _ _
-  -> Tensor _ _
+  => Conv2d _ _ _ _ _ _
+  -> Tensor _ _ _
+  -> Tensor _ _ _
```
junjihashimoto: PartialTypeSignatures is good on some examples, but I don't know the type definition of API.
If it is not easy  you add the type-definition, could you add some comments for '_' of  'Conv2d _ _ _ _ _ _'?
There is no need to fix it now. 
It ’s good to consider what to do with other PRs.

junjihashimoto:
Thank you for your great work!

hasktorch/src/Torch/Typed/NN.hs
tscholak: I got strong positive feedback from @apaszke for doing it like this, but we should discuss this approach together as a team, too.
I think the advantage is that the compiler doesn't need the type annotations, and therefore it's an opportunity to reduce boilerplate and noise, to make things less daunting, and to improve readability.
The downside is that one now has to look elsewhere for type hints and type documentation, and reasoning about misbehaviour and compile time errors could be harder.

tscholak:
@junjihashimoto thanks!
I just realize I still have to check that the static examples run on GPU... ;)
I just followed the types, but I still have to run the compiled code on CUDA

examples/static-transformer/Main.hs
junjihashimoto: thx!

junjihashimoto:
The operation on the GPU is an important function of this PR.
I wait for the result of GPU to approve.

PR #220, opened 2019-10-21
Xavier and Kaiming initializer implementations as examples

austinvhuang:
Work on untyped NN initializers and 1:1 comparison with pytorch counterparts (WIP)


tscholak:
could we put this in `Torch.NN.Initializer` or `Torch.Typed.NN.Initializers`, respectively?
Seems like every neural network initialization could make use of these :)

tscholak:
also, awesome work!

austinvhuang:
> could we put this in `Torch.NN.Initializer` or `Torch.Typed.NN.Initializers`, respectively?
> Seems like every neural network initialization could make use of these :)

Yes, in some way they seem to even slot into TensorFactories as a drop-in replacement for randn' since afaict the architecture context besides dimensionality doesn't get used unless the end-user explicitly implements something. But NN init is where pytorch has these.

As an aside, I found out these functions are actually already implemented in libtorch as well (linked in the readme), but I don't think the NN ops are part of declarations.yaml at this point so at least for now we may have to deal with the ffi by hand or write our own placeholder spec until the upstream specs get expanded.

austinvhuang:
Will go ahead and merge this as is if CI passes, next iterations should cover:
- add test spec to check summary stats for different tensor dimensions against equation defns
- write out python outputs and comparisons
- move to library

PR #221, opened 2019-10-22
Add image-processing

junjihashimoto:
- Read/Write both bitmap and png by JuicyPixels
- Add sharpness-filter
- Add lowpass-filter


austinvhuang:
Nice! Hadn’t played around with juicy pixels, wish the file read /write were a bit higher level and less verbose. Anyway I’m sure we’ll reuse these operations for vision examples down the line.

PR #223, opened 2019-10-23
static LSTM cell

o1lo01ol1o:
The LSTM cell is working and runs in the current grid lstm example (though this is currently missing the "grid" part).  

I can split this as a standalone PR if it's preferable to handle finishing the example separately.


hasktorch/src/Torch/Typed/Native.hs
```diff
@@ -2338,8 +2338,16 @@ q_zero_point _input = unsafePerformIO $ (cast1 ATen.q_zero_point_t) _input
...
--- lstm_cell _input _hx _w_ih _w_hh _b_ih _b_hh = unsafePerformIO $ (cast6 ATen.lstm_cell_tltttt) _input _hx _w_ih _w_hh _b_ih _b_hh
+lstm_cell :: Tensor dtype '[batchSize, inputDim] 
+  -> [Tensor dtype '[batchSize, hiddenSize]]
+  -> Tensor dtype '[4 * hiddenSize, inputDim] 
+  -> Tensor dtype '[4 * hiddenSize, hiddenSize] 
+  -> Tensor dtype '[4 * hiddenSize] 
+  -> Tensor dtype '[4 * hiddenSize]  
+  -> (Tensor dtype '[batchSize, hiddenSize],Tensor dtype '[batchSize, hiddenSize])
+lstm_cell _input _hx _w_ih _w_hh _b_ih _b_hh = unsafePerformIO $ (cast6 ATen.lstm_cell_tltttt) _input _hx _w_ih _w_hh _b_ih _b_hh
+  where
```
junjihashimoto: Thank you for your contribution!
Could you add a test of lstm_cell?

hasktorch/src/Torch/Typed/Native.hs
```diff
@@ -2338,8 +2338,16 @@ q_zero_point _input = unsafePerformIO $ (cast1 ATen.q_zero_point_t) _input
 -- combinations :: Tensor dtype shape -> Int -> Bool -> Tensor dtype shape
 -- combinations _input _r _with_replacement = unsafePerformIO $ (cast3 ATen.combinations_tlb) _input _r _with_replacement
 
--- lstm_cell :: Tensor dtype shape -> [Tensor dtype shape] -> Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> (Tensor dtype shape,Tensor dtype shape)
--- lstm_cell _input _hx _w_ih _w_hh _b_ih _b_hh = unsafePerformIO $ (cast6 ATen.lstm_cell_tltttt) _input _hx _w_ih _w_hh _b_ih _b_hh
+lstm_cell :: Tensor dtype '[batchSize, inputDim] 
+  -> [Tensor dtype '[batchSize, hiddenSize]]
```
tscholak: I believe it would be better if this could be a `Tensor dtype '[seqLen, batchSize, hiddenSize]` instead of a list of tensors, what do you think?

examples/static-gridLSTM/LSTM.hs
```diff
@@ -0,0 +1,112 @@
...
+    let backend = case backend' of
+            Right "CUDA" -> "CUDA"
+            _            -> "CPU"
+    let foldLoop x count block = foldM @[] block x [1 .. count]
+
+    -- randomly initializing training values
+    (input_tensor :: [Tensor D.Float '[1, 2]]) <- mapM
+        (const (randn :: IO (Tensor D.Float '[1, 2])))
+        [0 .. 4]
+    expected_output                 <- randn :: IO (Tensor D.Float '[1, 2])
```
tscholak: hmm, any idea what the expected behaviour of an lstm cell is when you train it on normally distributed input and output data?

tscholak:
awesome!

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,145 @@
...
+import qualified Torch.TensorFactories         as D
+import           Torch.Typed
+import           Torch.Typed.Factories
+import           Torch.Typed.Native      hiding ( linear )
+import           Torch.Typed.NN
+
+
+
+data LSTMCellSpec (dtype :: D.DType) (inputDim :: Nat) (hiddenDim :: Nat) =
+    LSTMCellRandomInitState -- ^ Random initial memory cells and hiddenstate values
```
apaszke: "Random state" is very uninformative. We should either mention what's the distribution, or better e.g. let people provide a tensor constructor (something like `forall shape. Tensor shape`).

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,145 @@
...
+    | LSTMCellZeroInitState -- ^ Zero-valued initial memory cells and hiddenstate values
+    deriving (Show, Eq, Ord, Generic, Enum, Bounded)
+
+data LSTMCell (dtype :: D.DType) (inputDim :: Nat) (hiddenDim :: Nat) =  LSTMCell {
+    lSTMCell_init_h :: Parameter dtype '[hiddenDim] -- ^ initial hidden state
```
apaszke: nit: let's make it `lstm` not `lSTM`?

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,145 @@
...
+    , lSTMCell_init_c :: Parameter dtype '[hiddenDim] -- ^ initial cell state
```
apaszke: Why do we keep the initial states in the data structure? We should probably reallocate them at each step.

tscholak: I concur. The initial states are not usually not trainable, but are supplied at the beginning of each recurrence.

hasktorch/src/Torch/Typed/Native.hs
o1lo01ol1o: What kind of test would you like to see and where would you like to see it?

hasktorch/src/Torch/Typed/Native.hs
o1lo01ol1o: The `lstm_cell` does not manage recurrent application, only state transitions.  As such, it only knows about `batchSize`, `inputDim`, and `hiddenSize`.  The list due to the underlaying c++ api which expects to receive the current state value as a list.

hasktorch/src/Torch/Typed/Native.hs
tscholak: we have used doctests in the past. the whole `Native.hs` module is full of them. feel free to replicate the structure of one of them.
what I'm interested in is a test that the statically calculated types match the runtime shape and dtype of the tensor.

examples/static-gridLSTM/LSTM.hs
o1lo01ol1o: This is just a copy of the memorization task that exists for the current recurrent models in example.  It was only used to make sure `autograd` was sensible.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
o1lo01ol1o: I concur, but I didn't see any established convention for initialization states.  My Preference would be something like a reference to a canonical initialization in the literature, but that might raise the bar for contributions.  I defer the maintainers on the best convention.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
o1lo01ol1o: Since the lstmCell does not manage recurrence, it expects the initial state to be provided and whatever the provided values are should be kept as part of the state of the cell; pytorch [defaults to zeros](https://pytorch.org/docs/stable/nn.html#lstmcell).  

The `foldl'`-based `forward` function threads the accumulated state through the container and passes the current state through each cell transition.

I had thought somewhere in the docs I saw that these were trainable, but it looks like although sensible [this is far from canonical](https://r2rt.com/non-zero-initial-states-for-recurrent-neural-networks.html).  I'll expose this case as a separate constructor.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,145 @@
...
+import           Torch.Typed.Factories
+import           Torch.Typed.Native      hiding ( linear )
+import           Torch.Typed.NN
+
+
+
+data LSTMCellSpec (dtype :: D.DType) (inputDim :: Nat) (hiddenDim :: Nat) =
+    LSTMCellRandomInitState -- ^ Random initial memory cells and hiddenstate values
+    | LSTMCellZeroInitState -- ^ Zero-valued initial memory cells and hiddenstate values
+    deriving (Show, Eq, Ord, Generic, Enum, Bounded)
```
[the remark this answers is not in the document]
o1lo01ol1o: Up to you you guys here; if you don't want to make any choices regarding initialization, I'll just let the user provide values.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
tscholak: I'm very confused now. I think we have to be careful not to mix up rnns and rnn cells.
In https://pytorch.org/docs/stable/_modules/torch/nn/modules/rnn.html, you can see two groups of implementations, one for rnns (base rnn, gru, lstm) and one for rnn cells (base rnn cell, gru cell, lstm cell).
In the latter case, for the rnn cells, not only the tensor input to the cell but also a hidden state is provided.
In the former case, for the rnns the module makes an assumption of what the first hidden state should be.
I think you made this assumption for the rnn cells where you should have made that assumption only for the rnns.

Thus, either don't make any assumptions for your cell implementation, or add new classes for lstm rnns.

o1lo01ol1o: > I think you made this assumption for the rnn cells where you should have made that assumption only for the rnns.

That's a fair assessment;  It seems like it makes sense forgo any default (hidden,cell) state for the `_cell` functions.  However, this wouldn't conform to the expectations of a pytorch user.

tscholak: I think we agree that the hidden states of cells are meant to be fed into the same cell again, over and over again, as one is folding the input sequence. It therefore does not make sense to feed in a default value at every step.
This begs the question what the first input hidden state should be. Pytorch does not have an opinion on that at the level of cells. The forward function of a cell is just one step of the recurrence and hence always takes both, input step features and hidden state.

tscholak: My suggestion would be that you follow the approach of Pytorch add another class, LSTMRNN. This one would indeed be opinionated on the first input hidden state, because its forward function is calling the lstm cell recurrently

o1lo01ol1o: > add another class, LSTMRNN [...] would indeed be opinionated [...]

Yes, that was my desired end-goal.  After the feedback, I think that this PR should only scope over the `LSTMCell` implementation and the other components should be separate.

> Pytorch does not have an opinion on that at the level of cells.

 [It does](https://pytorch.org/docs/stable/nn.html#lstmcell): if the current hidden and cell states are not provided to `forward`, they are set to zero values. But Aten makes no such assumptions and fails at runtime if two states are not passed.  This is the ambiguity that led to the current implementation.  

To be clear:  Your preference is that `lstm_cell` should always be provided `h_0` and `c_0` (and defaults should only be the prerogative of the code calling `lstm_cell`), contrary to the above PyTorch api?  This seems to me to be the cleaner api boundary, but, again, totally your call.

tscholak: > It does: ...

Thanks for pointing that out, I wasn't aware of that.

> Your preference is that lstm_cell should always be provided h_0 and c_0 (and defaults should only be the prerogative of the code calling lstm_cell), contrary to the above PyTorch api?

Yes, it would be my personal preference.

o1lo01ol1o:
The scope has been reduced to only cover the implementation of the `LSTMCell` module. 
 Assuming CI remains green on doctests, this is ready for review (again).  

I think I addressed everyone's comments, but feel free to point out anything that still needs tweaks.

hasktorch/src/Torch/Typed/Native.hs
```diff
@@ -2338,8 +2338,18 @@ q_zero_point _input = unsafePerformIO $ (cast1 ATen.q_zero_point_t) _input
 -- combinations :: Tensor dtype shape -> Int -> Bool -> Tensor dtype shape
 -- combinations _input _r _with_replacement = unsafePerformIO $ (cast3 ATen.combinations_tlb) _input _r _with_replacement
 
--- lstm_cell :: Tensor dtype shape -> [Tensor dtype shape] -> Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> (Tensor dtype shape,Tensor dtype shape)
--- lstm_cell _input _hx _w_ih _w_hh _b_ih _b_hh = unsafePerformIO $ (cast6 ATen.lstm_cell_tltttt) _input _hx _w_ih _w_hh _b_ih _b_hh
+-- | lstm_cell
+-- >>> dtype &&& shape $ fst $ lstm_cell (ones :: Tensor 'D.Float '[2,2]) [(ones :: Tensor 'D.Float '[2,3]), (ones :: Tensor 'D.Float '[2,3])] (ones :: Tensor 'D.Float '[12,2]) (ones :: Tensor 'D.Float '[12,3]) (ones :: Tensor 'D.Float '[12]) (ones :: Tensor 'D.Float '[12])
+-- (Float,[2,3])
+lstm_cell :: Tensor dtype '[batchSize, inputDim] 
+  -> [Tensor dtype '[batchSize, hiddenSize]]
```
tscholak: could you please make this a tuple instead like you did in the `forwardStep` function?

junjihashimoto:
LGTM!
I will check CI-result, when I get home.

o1lo01ol1o:
Hm.  Looks like this might be a real problem.  Let me update again.

(doctests appear to be broken for me locally on `darwin`)

hasktorch/src/Torch/Typed/Native.hs
```diff
@@ -2350,7 +2351,7 @@ lstm_cell :: Tensor dtype '[batchSize, inputDim]
   -> (Tensor dtype '[batchSize, hiddenSize],Tensor dtype '[batchSize, hiddenSize])
 lstm_cell _input (_cc, _hc) _w_ih _w_hh _b_ih _b_hh = unsafePerformIO $ (cast6 ATen.lstm_cell_tltttt) _input _hx _w_ih _w_hh _b_ih _b_hh
   where 
-    _hx = [toDynamic _cc, toDynamic _hc]
+    _hx = [_cc, _hc] :: [Tensor dtype '[batchSize, hiddenSize]]
```
tscholak: yes, toDynamic shouldn't be necessary here. I'm surprised that you have to annotate the type. does it type-check but not run?

o1lo01ol1o: The test fails with an ambiguous type. I think due to the `IsList` instance of tensor.  We'll see if the annotation clears it up.

PR #230, opened 2019-10-28
Add serialize-function for tensor

junjihashimoto:
This pr supports serialize-function using checkpoint-file of pytorch.

### DIFF
--- a/hasktorch/hasktorch.cabal
+++ b/hasktorch/hasktorch.cabal
@@ -30,6 +30,7 @@ library
                     , Torch.Layout
                     , Torch.Cast
                     , Torch.Dimname
+                    , Torch.Serialize
  hs-source-dirs:      src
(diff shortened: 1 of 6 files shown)


tscholak:
OMG that's awesome, I had no idea it could be done with libtorch!

junjihashimoto:
I read today and realized that I can do it.

tscholak:
question: do you know if this tensorlist you're dumping/loading has to be on cpu?

junjihashimoto:
That's a good question.
I have to test it.
Before the test, I checked some samples using save and load.
Probably, save supports both cpu and gpu, on the other hand load outputs the data to cpu only.

junjihashimoto:
Yes, I will. 
Because torch's serialize supports deserialization with the correct type.

PR #232, opened 2019-11-03
Static lstm

o1lo01ol1o:
Doctests were green, let me know what else anyone would like to see from this.


hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,478 @@
...
+
+-- | helper to do xavier uniform initializations on weight matrices and
+-- orthagonal initializations for the gates.
+xavierUniormLSTM
+    :: forall device dtype hiddenSize d
+     . (KnownDType dtype, KnownNat d, KnownNat hiddenSize, KnownDevice device, RandDTypeIsValid device dtype)
+    => IO (Tensor device dtype '[4 * hiddenSize, d])
+xavierUniormLSTM = do
+    fixme <- randn
+    pure $ fixme -- TODO: what's the story with the initializers code?
```
tscholak: @austinvhuang

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,478 @@
...
+                                              (cc', hc')
+                                              (params p)
+                                              d
+                                              doDropout
+  where
+    cc' =
+        reshape @'[numLayers * (NumberOfDirections dir), batchSize, hiddenDim]
+            $ expand
+                  @'[batchSize, numLayers * (NumberOfDirections dir), hiddenDim]
+                  False -- TODO: What does the bool do?
```
tscholak: nobody knows ¯\ _(ツ)_/¯
cf. https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/Native.hs#L1788

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,478 @@
...
+import           Torch.Typed.Factories
+import           Torch.Typed.Native       (expand, lstm)
+import           Torch.Typed.NN
+import           Torch.Typed.Tensor
+
+
+-- | A specification for a long, short-term memory layer.
+--
+data LSTMSpec device (dir :: Directionality)
+        (dtype :: D.DType)
```
tscholak: In my `nn` code, I decided to put `dtype` and `device` at the end, cf. https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/NN.hs#L89, for instance.
the reasoning was that, with type applications enabled, one would almost never need to apply these, but almost always have to annotate the others.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,478 @@
...
+
+    | LSTMSpecLearnedInit             -- ^ Weights drawn from Xavier-Uniform
+                                      --   with zeros-value initialized biases
+                                      --   and learned cell states.
+        (Tensor device dtype '[numLayers * (NumberOfDirections dir), hiddenDim]) -- ^ The initial (learnable)
+                                      -- values of the memory cell
+        (Tensor device dtype '[numLayers * (NumberOfDirections dir), hiddenDim]) -- ^ The initial (learnable)
+                                      -- values of the hidden state
+        DropoutSpec
+    deriving (Show, Generic)
```
tscholak: the documentation is beautiful, thanks for that!

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,478 @@
...
+                    -> LSTMParams device dtype numDirections inputDim hiddenSize 1
+    LSTMLayerK ::(1 <= numLayers)
+                    => LSTMParams device dtype numDirections inputDim hiddenSize numLayers
+                    -> (Parameter device dtype '[4 * hiddenSize, numDirections * hiddenSize])
+                    -> (Parameter device dtype '[4 * hiddenSize, hiddenSize])
+                    -> (Parameter device dtype '[4 * hiddenSize ])
+                    -> (Parameter device dtype '[4 * hiddenSize ])
+                    -> LSTMParams device dtype numDirections inputDim hiddenSize (numLayers + 1)
+
+--  TODO: Generics? see https://gist.github.com/RyanGlScott/71d9f933e823b4a03f99de54d4b94d51
```
tscholak: I don't think you'll be able to derive a generics instance for `LSTMParams` because of the constraint in the data constructor for `LSTMLayerK`

tscholak: oh, https://gist.github.com/RyanGlScott/71d9f933e823b4a03f99de54d4b94d51 is neat

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,478 @@
...
+-- (and optionally) bidirectional LSTM.
+--
+data LSTMParams (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (numDirections :: Nat) (inputDim :: Nat) (hiddenSize :: Nat) (numLayers :: Nat) where
+    LSTMLayer1 ::Parameter device dtype '[4 * hiddenSize, inputDim]
+                    -> (Parameter device dtype '[4 * hiddenSize, hiddenSize])
+                    -> (Parameter device dtype '[4 * hiddenSize ])
+                    -> (Parameter device dtype '[4 * hiddenSize ])
+                    -> LSTMParams device dtype numDirections inputDim hiddenSize 1
+    LSTMLayerK ::(1 <= numLayers)
+                    => LSTMParams device dtype numDirections inputDim hiddenSize numLayers
```
tscholak: instead of nesting layers like this, we could also use an `HList`. what do you think?

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,478 @@
...
+      , lstm_init_c     :: Tensor device dtype '[numLayers * (NumberOfDirections dir), hiddenDim]
+      , lstm_init_h     :: Tensor device dtype '[numLayers * (NumberOfDirections dir), hiddenDim]
+    }
+    | LSTMLearnedInit {
+        lstmLearnedInit_params :: ParamsPerDirection device dtype inputDim hiddenDim numLayers dir
+        , lstmLearnedInit_dropout :: Dropout
+        , lstmLearnedInit_init_c     :: Parameter device dtype '[numLayers * (NumberOfDirections dir), hiddenDim]
+        , lstmLearnedInit_init_h     :: Parameter device dtype '[numLayers * (NumberOfDirections dir), hiddenDim]
+    }
+    deriving Generic
```
tscholak: that's cool and gives a lot of flexibility. thanks for including these two options here

hasktorch/src/Torch/Typed/Native.hs
```diff
@@ -2943,9 +2943,72 @@ qZeroPoint input = unsafePerformIO $ cast1 ATen.q_zero_point_t input
...
+          '[numLayers * numDirections, batchSize, hiddenSize]
+      ]
+  (_num_layers :: I.Int64) =
+    fromIntegral $ natVal (undefined :: Proxy numLayers)
+  _bidirectional = case natVal (undefined :: Proxy numDirections) of
+    1 -> False
+    2 -> True
+    _ -> error "lstm: numDirections must be 1 or 2!"
+
+-- gru_cell :: Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape -> Tensor dtype shape
```
tscholak: could you add the `lstmCell` doc and doctests back here please?

hasktorch/src/Torch/Typed/Native.hs
```diff
@@ -2943,9 +2943,72 @@ qZeroPoint input = unsafePerformIO $ cast1 ATen.q_zero_point_t input
...
--- >>> dtype &&& shape $ fst $ lstmCell (ones :: CPUTensor 'D.Float '[12,2]) (ones :: CPUTensor 'D.Float '[12,3]) (ones :: CPUTensor 'D.Float '[12]) (ones :: CPUTensor 'D.Float '[12]) ((ones :: CPUTensor 'D.Float '[2,3]), (ones :: CPUTensor 'D.Float '[2,3])) (ones :: CPUTensor 'D.Float '[2,2])
--- (Float,[2,3])
+-- | lstm
+-- Parameters for this ATen function are non-trivially provided.  See the
+-- `Typed.NN.LSTM` module for doctests.
+--
+lstm
+  :: forall
+       numLayers
+       numDirections
```
tscholak: there are only two valid choices, 1 and 2, right? wouldn't it be better to use the ADT you defined in the other module here?

hasktorch/src/Torch/Typed/Native.hs
```diff
@@ -2943,9 +2943,72 @@ qZeroPoint input = unsafePerformIO $ cast1 ATen.q_zero_point_t input
...
+                                                False -- _batch_first
+ where
+  _hx =
+    [_cc, _hc] :: [ Tensor
+          device
+          dtype
+          '[numLayers * numDirections, batchSize, hiddenSize]
+      ]
+  (_num_layers :: I.Int64) =
+    fromIntegral $ natVal (undefined :: Proxy numLayers)
```
tscholak: we've got a helper for that, `natValI`, defined elsewhere

hasktorch/src/Torch/Typed/Native.hs
```diff
@@ -2943,9 +2943,72 @@ qZeroPoint input = unsafePerformIO $ cast1 ATen.q_zero_point_t input
...
+  -> ( Tensor
+         device
+         dtype
+         '[numLayers * numDirections, batchSize, hiddenSize]
+     , Tensor
+         device
+         dtype
+         '[numLayers * numDirections, batchSize, hiddenSize]
+     )
+  -> [D.Tensor]
```
tscholak: ideally, there would be no reference to `D.Tensor` in all signatures in `Torch.Typed.Native`.
if we used `HList`s here, we could make that happen. This could be done later, though.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,478 @@
...
+    :: forall device dtype inputDim hiddenDim numLayers dir
+     . ParamsPerDirection device dtype inputDim hiddenDim numLayers dir
+    -> [D.Tensor]
+params (BidirectionalParams fwd rvs) =
+    ziplstmLayers [] (lstmParamsToTlist' [] fwd) (lstmParamsToTlist' [] rvs)
+params (UniDirectionalParams fwd) = lstmParamsToTlist' [] fwd
+
+-- | Forward propagage the `LSTM` module and apply dropout on the outputs of each layer.
+--
+-- >>> input ::Tensor 'D.Float '[5,16,10]          <- randn
```
tscholak: that should be `CPUTensor 'D.Float '[5,16,10]`.
are you sure, doctests runs on this file?

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -0,0 +1,478 @@
...
+            <*> A.replaceOwnParameters d
+
+-- | The directional specification of the LSTM
+--
+data Directionality =
+    Bidirectional -- ^ Forward and backward along the sequential axis using independant parameters for each.
+    | Unidirectional -- ^ Forward along the sequential axis.
+    deriving (Show, Generic) -- TODO:  We could also have BidirectionalTied weights.
+
+type family NumberOfDirections a where
```
tscholak: could you annotate the kind here please, so that a `NumberOfDirections dir` constraint always forces `dir :: Directionality`?

tscholak: same for the return kind, i.e. `Nat`

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
o1lo01ol1o: I know, it's not documented in the ATen source either.  (Which seems to be poorly documented given that its such a vital codebase . . .  one wonders how many less-than-breaking bugs are lying around in there.)

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
o1lo01ol1o: Good call

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
o1lo01ol1o: Yes, but a finicky, error-prone investment at this point.  Esp if you want to move everything to `HList`s later.

hasktorch/src/Torch/Typed/Native.hs
o1lo01ol1o: I though I accepted both changes but it looks like the merge ate it.  It's been restored.

hasktorch/src/Torch/Typed/Native.hs
o1lo01ol1o: It depends on how you want the modules structured.  I saw the `Native` module as an unsafe, internal module.  You could beef up the safety but I think it would add a fair amount of code in that sourcefile when all is said and done.  

I had a lot of stuff along side the `lstm` function thinking that it would also be used in the other recurrent cells, but it got cluttered and I thought it would be better to just focus on the `lstm` before factoring our the shared machinery.

hasktorch/src/Torch/Typed/Native.hs
o1lo01ol1o: found and fixed

hasktorch/src/Torch/Typed/Native.hs
o1lo01ol1o: Right, that makes sense but probably better done when / if someone get's the itch to make the inductive `LSTM` bits `HList`s.

o1lo01ol1o: Though the latter would also make `Generics` easier, and `Generics` make life in production much nicer.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
o1lo01ol1o: Right, done.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
o1lo01ol1o: I thought the doctests were auto-discovered but since found the file where they're specified, let's see if it works now.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
o1lo01ol1o: I can see there's a good argument for the `HList` but I'm inclined to leave it for another time (like when `Generics` become more valuable).  

In general, I find the inductive GADTs to be simpler in developing the invariants but once you know what they are, I guess going to type families and constraints isn't that bad.  

(Though I have to say, figuring out the `Overlapping` pragmas and the needed constraints on that singleton helper class was a hard-won struggle I would have preferred to avoid . . . )

hasktorch/src/Torch/Typed/Native.hs
tscholak: initially, the `Typed.Native` module was supposed to be automatically generated and a thin wrapper, but, after manually implementing many ops, it turns out that it won't be possible to automatically generate the constraints for all ops in there for quite some time. `Typed.Native` has just become a place where we implemented type-safe ops and also put auxiliary type families and classes. Therefore, I think it does make sense to put helpers in there. If it is a generic helper, consider also `Typed.Aux`.

o1lo01ol1o: addressed

tscholak: thank you!

tscholak:
all green from my side, merging!
Thanks a lot, hope to see contributions from you again!

PR #235, opened 2019-11-05
fix svd and inverse on cuda

tscholak:
@junjihashimoto discovered some issues and additional constraints while testing on cuda. this is the fix.


junjihashimoto:
LGTM!

PR #239, opened 2019-11-08
LSTM Fix: flatten order needs to be the same as replace order

o1lo01ol1o:
:derp:


tscholak:
Oops

junjihashimoto:
I recommend to use generic to make the instance.
If generic does not work, I can improve it.

PR #240, opened 2019-11-10
Untyped MNIST Example + Updates to Untyped Library

austinvhuang:
Untyped MNIST example + library updates for functions needed to support functionality.

- [x] Found a library bug in initialization of FC layers where dimension order seems to be flipped. Please double check this before merging though since it affects all uses of FC layers
- [x] Adding a simplified nllLoss for untyped, added argmax for untyped with some helper types
- [x] Reviewed and updated API for untyped reshape operation, propose to reorder s.t. shape comes first, input comes last as per convention elsewhere. Fixed library and example code affected.
- [x] diagnostics + checks of model behavior
- [x] add some text-mode rendering of mnist images + show a few model outputs
- [x] add content to readme
- [x] clean up path location of data download - for now just replicate the download script in this example directory (duplicates dataset here, but at least that avoids having to confusingly run this example from within static-mnist)
- [ ] simplify data loading/chunking code (can we move the pointer arithmetic stuff into the beginnings of a torchvision-like module? it's a bit intimidating for a beginner example)
- [ ] clean up shared dependency with typed mnist modules

**The model works as is, but will look to address last 2 items above in future PRs.

Due to the change in the untyped reshape interface, and changes to the module organization in other PRs, I'd suggest merging to master to avoid continuing conflicts.

This is ready to review now.**


hasktorch/src/Torch/NN.hs
```diff
@@ -107,9 +112,16 @@ data LinearSpec = LinearSpec { in_features :: Int, out_features :: Int }
...
+linear' layer input = linear input w b
+    where
+        linear input weight bias = unsafePerformIO $ (cast3 ATen.linear_ttt) input weight bias
+        w = toDependent (weight layer)
+        b = toDependent (bias layer)
+
 instance Randomizable LinearSpec Linear where
   sample LinearSpec{..} = do
-      w <- makeIndependent =<< randn' [in_features, out_features]
+      w <- makeIndependent =<< randn' [out_features, in_features]
```
austinvhuang: This looks to be a dimension bug in the existing implementation, but please check this when we get around to reviews for the PR since it effects every use of untyped FC layers.

hasktorch/src/Torch/NN.hs
```diff
@@ -107,9 +113,17 @@ data LinearSpec = LinearSpec { in_features :: Int, out_features :: Int }
 
 data Linear = Linear { weight :: Parameter, bias :: Parameter } deriving (Show, Generic)
 
+linear' :: Linear -> Tensor -> Tensor
```
apaszke: Why the `'`?

hasktorch/src/Torch/NN.hs
```diff
@@ -107,9 +113,17 @@ data LinearSpec = LinearSpec { in_features :: Int, out_features :: Int }
...
+    where
+        linear input weight bias = unsafePerformIO $ (cast3 ATen.linear_ttt) input weight bias
+        w = toDependent (weight layer)
+        b = toDependent (bias layer)
+
 instance Randomizable LinearSpec Linear where
   sample LinearSpec{..} = do
-      w <- makeIndependent =<< randn' [in_features, out_features]
+      w <- makeIndependent =<< kaimingUniform' [out_features, in_features]
+      -- w <- makeIndependent =<< randn' [out_features, in_features]
```
apaszke: We should probably decide if we want to match upstream with initialization, or if we want to use some more modern choices, and apply those rules more consistently.

hasktorch/src/Torch/NN.hs
junjihashimoto: Our implemented linear function is `linear input depWeight depBias= matmul input depWeight + depBias`.

We begun to use Pytorch's linear.
Pytorch's linear behaves below.

```
> linear (ones' [10,2]) (ones' [3,2]) (ones' [])
Tensor Float [10,3] [..]
```

Pytorch's linear may be below.
```
linear input depWeight depBias= matmul input (transpose depWeight) + depBias`
```

So this depends on which linear we use.

austinvhuang: Yes that's what I saw as well. Fair it's more of a difference in convention than bug, but I think it's going to be very confusing if we don't follow the defacto standard. Also, checking into the typed code, it uses the same (pytorch) convention.

Note that there are several untyped examples that re-implement the above linear function, I've changed all occurrences of them as well (not just mnist). They have to be consistent since they're sharing the typeclass implementation.

hasktorch/src/Torch/NN.hs
austinvhuang: This is a holdover from initial exploration when a few existing `linear` implementations were in the examples and I was trying to avoid breaking them at first. 

However, now that all untyped implementations are standardized to the library implementation and use the pytorch dimension conventions (they had to be in order to share the Randomizable typeclass), it makes since to rename this and its occurrences in examples to `linear`.

hasktorch/src/Torch/NN.hs
austinvhuang: Could you point to some more modern choices you have in mind?

I don't feel too strongly atm, at least pytorch's defaults are well known (untyped hasktorch optimizations seem to be well behaved for me now, though haven't ventured into hairier architectures much). If there's something better I wouldn't mind trying as long as it's not much more complex to implement.

hasktorch/src/Torch/NN.hs
austinvhuang: I've replaced `linear'` with `linear`, leaving the name clash resolution up to the user to determine at import time for now.

Going forward we should try to land on a convention for the function version vs. the version that operates on `Parameter` state.

hasktorch/src/Torch/NN.hs
apaszke: Ouch, we should just decide on a single convention for linear transforms. Otherwise it's going to be too error prone.

hasktorch/src/Torch/NN.hs
apaszke: Well pure randn is simply a bad idea, PyTorch uses some variants of LeCun initialization, kaimingUniform is yet another choice (I think a more modern one?).

PR #241, opened 2019-11-11
Add prometheus

junjihashimoto:
This PR adds prometheus-metrics of GHC to check memory-management with GC.

static-mnist-mlp's result is below.
It seems program calls GC many times, but memory is not released.
image

### DIFF
--- a/examples/examples.cabal
+++ b/examples/examples.cabal
@@ -13,7 +13,7 @@ build-type:          Simple
 
 common config
   default-language:    Haskell2010
-  ghc-options:         -fplugin GHC.TypeLits.Normalise -fplugin GHC.TypeLits.KnownNat.Solver -fplugin GHC.TypeLits.Extra.Solver -fconstraint-solver-iterations=0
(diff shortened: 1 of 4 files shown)


austinvhuang:
I was recently looking over and stripping down the static mnist code for an untyped implementation https://github.com/hasktorch/hasktorch/pull/240 and was thinking it'd be good to at least refactor the static example to pull the instrumentation code out of the core training loop.

Also, it might have gotten a bit too modular for what started as a standalone example (there's 4 modules split across an app and a library). By contrast the pytorch example is one file https://github.com/pytorch/examples/blob/master/mnist/main.py

Maybe we could an go one step further than simplifying this example and consider there's two separate goals reflected here - an instructive MNIST example vs. a deeply-instrumented test case to troubleshoot GC.

Up to now we've been pretty lax about using examples as an experimental testing ground, but as newcomers are starting to add eyes to the project, we could improve the situation by putting our own experimental functionality and needs under a separate top level `experimental/` area while keeping the `examples/` oriented towards clean implementations aimed at helping new users.

junjihashimoto:
@apaszke @austinvhuang 
Thank you for your advice.
I will separate debug-codes(experimental codes) and examples.

austinvhuang:
BTW really impressed with the instrumentation you've implemented @junjihashimoto ! Makes me feel optimistic about being able to tackle the resource management implementation one piece at a time.

PR #243, opened 2019-11-13
Expose the `batch_first` flag from the ATen api

o1lo01ol1o:


junjihashimoto:
@o1lo01ol1o @tscholak 
Thank you for your great contribution.
I feel HList is great for this application.
but I concern about  the performance, if we use long HList.
Are there any performance issues?

tscholak:
@junjihashimoto HLists are linked lists, and as such they should be O(n) for selection and O(1) for extension at runtime. For very large HLists, like a hundred of elements and more, I expect compile times to be markedly longer than they would be with lists. This overhead should be gone at runtime, though.

PR #248, opened 2019-11-24
typed gradients and optimizers

tscholak:
I added typed optimizers to the core hasktorch lib and used them for the typed mnist mlp example. Please have a look in `Torch.Typed.Autograd` and `Torch.Typed.Optim`. I still want to implement a test suite based on the test functions @austinvhuang used for the `Optimizers` example module and following some of the tests in pytorch, cf. https://github.com/pytorch/pytorch/blob/master/test/test_optim.py.

The mlp code compiles fine and runs without error, but learning isn't working yet:
```
[nix-shell:~/Projects/thirdParty/hasktorch/examples/static-mnist]$ DEVICE="cpu" ../dist/build/static-mnist-mlp/static-mnist-mlp
Iteration: 250. Training batch loss: 2.7391508. Test loss: 1.8653502. Test error-rate: 0.56799316
Iteration: 500. Training batch loss: 3.7517178. Test loss: 2.6845038. Test error-rate: 0.5695801
Iteration: 750. Training batch loss: 2.5789738. Test loss: 1.9683098. Test error-rate: 0.66186523
Iteration: 1000. Training batch loss: 3.3718398. Test loss: 2.5216846. Test error-rate: 0.633667
Iteration: 1250. Training batch loss: 4.0687294. Test loss: 3.5834208. Test error-rate: 0.5814209
Iteration: 1500. Training batch loss: 3.7929404. Test loss: 3.1979089. Test error-rate: 0.70129395
Iteration: 1750. Training batch loss: 2.2915678. Test loss: 1.7558963. Test error-rate: 0.63964844
Iteration: 2000. Training batch loss: 2.4241033. Test loss: 1.9236282. Test error-rate: 0.60131836
Iteration: 2250. Training batch loss: 3.0669777. Test loss: 2.4795806. Test error-rate: 0.80407715
```
The reason is unknown. Maybe there's a bug in the implementation of `Adam` and `GDM`. `GD` works fine. Additional tests should help clarifying.

### DIFF
--- a/hasktorch/src/Data/HList.hs
+++ b/hasktorch/src/Data/HList.hs
@@ -69,6 +69,22 @@ instance (Monoid a, Monoid (HList as)) => Monoid (HList (a ': as)) where
 class Apply f a b where
   apply :: f -> a -> b
 
+-- | Stronger version of `Apply` that allows for better inference of the return type
+class Apply' f a b | f a -> b where
+  apply' :: f -> a -> b
+
+data Fst = Fst
+
+instance Apply' Fst (a, b) a
+  where
+    apply' _ (a, _) = a
+
+data Snd = Snd
+
+instance Apply' Snd (a, b) b
+  where
+    apply' _ (_, b) = b
+
 class HMap f xs ys where
   hmap :: f -> HList xs -> HList ys
 
@@ -78,6 +94,16 @@ instance HMap f '[] '[] where
 instance (Apply f x y, HMap f xs ys) => HMap f (x ': xs) (y ': ys) where
   hmap f (x :. xs) = apply f x :. hmap f xs
 
+-- | Alternative version of `HMap` with better type inference based on `Apply'`
+class HMap' f xs ys | f xs -> ys where
+  hmap' :: f -> HList xs -> HList ys
+
+instance HMap' f '[] '[] where
+  hmap' _ _ = HNil
+
+instance (Apply' f x y, HMap' f xs ys) => HMap' f (x ': xs) (y ': ys) where
+  hmap' f (x :. xs) = apply' f x :. hmap' f xs
+
 class HMapM m f xs ys where
   hmapM :: f -> HList xs -> m (HList ys)
 
@@ -85,7 +111,16 @@ instance (Monad m) => HMapM m f '[] '[] where
   hmapM _ _ = pure HNil
 
 instance (Monad m, Apply f x (m y), HMapM m f xs ys) => HMapM m f (x ': xs) (y ': ys) where
-  hmapM f (x :. xs) = (:.) <$> (apply f x) <*> (hmapM f xs)
+  hmapM f (x :. xs) = (:.) <$> apply f x <*> hmapM f xs
+
+class HMapM' m f xs ys | f xs -> ys where
+  hmapM' :: f -> HList xs -> m (HList ys)
+
+instance (Monad m) => HMapM' m f '[] '[] where
+  hmapM' _ _ = pure HNil
+
+instance (Monad m, Apply' f x (m y), HMapM' m f xs ys) => HMapM' m f (x ': xs) (y ': ys) where
+  hmapM' f (x :. xs) = (:.) <$> apply' f x <*> hmapM' f xs
 
 class Applicative f => HSequence f xs ys | xs -> ys, ys f -> xs where
   hsequence :: HList xs -> f (HList ys)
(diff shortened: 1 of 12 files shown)


hasktorch/src/Data/HList.hs
```diff
@@ -69,6 +69,22 @@ instance (Monoid a, Monoid (HList as)) => Monoid (HList (a ': as)) where
 class Apply f a b where
   apply :: f -> a -> b
 
+-- | Stronger version of `Apply` that allows for better inference of the return type
+class Apply' f a b | f a -> b where
```
o1lo01ol1o: Might also be nice to add a haddock to this module that briefly describes some of the uses of this machinery in the domain (ie, "HLists" let us carry around parameters of different static size as well as "stack" different "types" of things).

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -579,10 +578,8 @@ lstmWithDropout, lstmWithoutDropout
      , hxShape ~ '[numLayers * NumberOfDirections directionality, batchSize, hiddenSize]
      , Parameterized (LSTM inputSize hiddenSize numLayers directionality dtype device) parameters
      , tensorParameters ~ LSTMR inputSize hiddenSize numLayers directionality dtype device
-     , HFoldrM IO TensorListFold [D.ATenTensor] tensorParameters
-     , Apply TensorListUnfold [D.ATenTensor] (HUnfoldMRes IO [D.ATenTensor] tensorParameters)
-     , HUnfoldM IO TensorListUnfold (HUnfoldMRes IO [D.ATenTensor] tensorParameters) tensorParameters
-     , HMap ToDependent parameters tensorParameters
+     , ATen.Castable (HList tensorParameters) [D.ATenTensor]
+     , HMap' ToDependent parameters tensorParameters
```
o1lo01ol1o: The reduction in constraints here is nice.

hasktorch/src/Data/HList.hs
```diff
@@ -78,14 +94,33 @@ instance HMap f '[] '[] where
 instance (Apply f x y, HMap f xs ys) => HMap f (x ': xs) (y ': ys) where
   hmap f (x :. xs) = apply f x :. hmap f xs
 
+-- | Alternative version of `HMap` with better type inference based on `Apply'`
+class HMap' f xs ys | f xs -> ys where
```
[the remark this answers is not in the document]
tscholak: I wish. I can't deprecate `HMap` because not for all operations you can define an `Apply'` instance. There's an instance of `Apply` in `Torch.Typed.Tensor` that can't be made into an `Apply'` instance because -- to channel ghc -- the output type is not determined jointly by `f` and the input type...

hasktorch/src/Data/HList.hs
tscholak: yes, this module has been growing for a while now, and it is very opaque

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
tscholak: yes, this was possible before this PR as well, I just didn't see it ;)

tscholak:
@junjihashimoto it would be cool if you could finish your PR #253 first. then I can apply the changes to the simplified MNIST modules. would be even greater if I had to that only once in Common :)

tscholak:
@junjihashimoto or, alternatively, we merge in this here now, you resolve merge conflicts in your PR #253 , and I provide the tests later.

tscholak:
@junjihashimoto @o1lo01ol1o everything is typed now in the typed mnist examples, the gradients, the optimizers, even the serialization.

hasktorch/src/Torch/Typed/Autograd.hs
```diff
@@ -0,0 +1,42 @@
...
+import qualified ATen.Cast as ATen
+import qualified ATen.Class as ATen
+import qualified Torch.Managed.Autograd
+import qualified Torch.DType                   as D
+import qualified Torch.Device                  as D
+import qualified Torch.Tensor                  as D
+import           Torch.Typed.Tensor
+import           Torch.Typed.Parameter
+
+type family GradR (parameters :: [a]) (dtype :: D.DType) (device :: (D.DeviceType, Nat)) :: [a] where
```
apaszke: Why do we require all gradients to be on a single device and have the same dtype? This seems like an unnecessary requirement

tscholak: Hey @apaszke, thanks for the question. This is interesting, are you implying that `grad` supports dtype promotion and loss and/or tensor parameters on different devices? I never tried it on such inputs. I just assumed it would be as restricted as most other ops.

apaszke: Not sure what you mean by "dtype promotion", but the quantities you differentiate _can_ be on different devices and have distinct dtypes. The returned gradients will have the dtypes and devices matching the HList of quantities differentiated wrt., so they can be heterogenous too.

tscholak: dtype promotion as described here: https://pytorch.org/docs/stable/tensor_attributes.html

This is great news for me, that means that a severe limitation I had taken for granted actually doesn’t exist.

apaszke: No, data types are always preserved. The signature of grad really is:
```
grad :: outputTensors -> inputTensors -> inputTensors
```
meaning that the gradients are a HList of the same type as the second argument.

tscholak: Nice, I’ll lift the restrictions and add some tests in my next PR

PR #250, opened 2019-11-25
Add random generator for high level api

junjihashimoto:


hasktorch/src/Torch/Random.hs
```diff
@@ -0,0 +1,68 @@
...
+  unsafePerformIO $ do
+    gen <- case device of
+             Device CPU _ -> ATen.newCPUGenerator seed
+             Device CUDA idx -> do
+               gen <- ATen.newCUDAGenerator (fromIntegral idx)
+               ATen.generator_set_current_seed gen seed
+               return gen
+    tensor <- (cast3 func) size gen options
+    nextSeed <- ATen.generator_current_seed gen
+    return (tensor,Generator device nextSeed)
```
tscholak: would it be possible to add the `ATen.Generator` to the `Torch.Random.Generator` data type?

junjihashimoto: I think it it not possible, because `ATen.Generator` is mutable.
In following case, the result depends on whether t2 or t3 is evaluated first, if we add it.

```
let (t,next) = randn [2] (cpuGenerator 0)
     (t2,next2) = randn [2] next
     (t3,next2) = randn [2] next

print t2
print t3
-- or --
print t3
print t2
```

tscholak: I'm worried about allocating new generators all the time when we could reuse already existing ones.
what if you made the constructor for `Torch.Random.Generator` private, and forced people to use a few smart utility functions that always make sure that the internal `ATen.Generator` is set to the accompanying seed in `Torch.Random.Generator`. that way you would prevent the tracked seed and the internal seed to run out of sync. kind of like you are already doing in the `generatorFactory` function...

junjihashimoto: When the random-tensor's size is big enough, we can ignore the overhead.
However, I think  a implementation of the tracked seed by using following data.

```
data Generator = Generator {
  seed :: Word64
, device :: Device
, generator :: (TVar (Maybe (ForeignPtr Generator)))
```

junjihashimoto: As you said, privete constructor is a good choice anyway.
I will add it and hide original constructor.

tscholak: cool, looking forward to that!

hasktorch/src/Torch/Random.hs
```diff
@@ -0,0 +1,107 @@
...
+data Generator
+  = Generator
+  { device :: Device
+  , seed :: Word64
+  , generator :: TVar (Maybe (ForeignPtr ATen.Generator))
+  }
+  deriving (Eq,Show)
+
+mkGenerator :: Device -> Word64 -> IO Generator
+mkGenerator device seed =
```
junjihashimoto: This is the private constructor of generator.
If the generator has a pointer to the actual ATen generator, the pointer moves to the next generator.

tscholak:
thanks, this looks much better! thanks for adding the pointer to the libtorch generator to the data type

junjihashimoto:
Thx!

hasktorch/src/Torch/Random.hs
```diff
@@ -0,0 +1,107 @@
...
+                Nothing -> case device of
+                  Device CPU _ -> ATen.newCPUGenerator seed
+                  Device CUDA idx -> do
+                    gen <- ATen.newCUDAGenerator (fromIntegral idx)
+                    ATen.generator_set_current_seed gen seed
+                    return gen
+    tensor <- (cast3 func) size genPtr options
+    nextSeed <- ATen.generator_current_seed genPtr
+    nextGenenerator <- newTVarIO (Just genPtr)
+    return (tensor,Generator device nextSeed nextGenenerator)
```
apaszke: I don't think we should do this. PyTorch RNG infrastructure is not designed around efficient generator copying and it's very likely that the performance will be terrible. IIRC We might be using some PRNGs with huge states.

tscholak: @junjihashimoto hm, can you confirm that the ATen generator is actually copied? To me it looks like you are just mutating a TVar with maybe an ATen generator in it, allocating a new one only at first use

junjihashimoto: @apaszke @tscholak 

This data with TVar manages ownership of ATen generator.
When the data has ownership, copy does not occur.
When the data does not have ownership, new ATen generator is created by seed.

Let me show you two examples.

Case1:
```
rng0 <- mkGenerator ..
(t1,rng1) = rand' [] rng0
(t2,rng2) = rand' [] rng1
```
In this case, copy does not occur.
rng0's ownership moves to rng1, next rng1 moves to rng2.

Case2:
rng0 <- mkGenerator ..
(t1,rng1) = rand' [] rng0
(t2,rng2) = rand' [] rng0

In this case, copy occurs.
rng0's ownership moves to rng1, next rng0 does not have ownership, and rng2 is created from seed.

apaszke: Oh I see. However, I'm not convinced that `current_seed` gives you a seed that would yield an equivalent generator. IIRC it just returns the seed that it was initialized with, so it will produce a constant stream as a result? Would need to check the PyTorch code though.

junjihashimoto: > so it will produce a constant stream as a result?

Yes.
This randn's test reproduces the same stream with previous called stream.
https://github.com/hasktorch/hasktorch/pull/250/files#diff-1ad63067a684ef689650d73894fc7038R19

> Would need to check the PyTorch code though.

I will do it later.

PR #256, opened 2019-12-05
Move modules of  {ATen,Torch}.* to LibTorch.{ATen,Torch}.*

junjihashimoto:
In current tree of modules, libtorch(api wrapper of libtorch)'s modules and hasktorch(high level api)'s one are mixed.
This pr splits the tree into libtorch's tree and hasktorch's one.


tscholak:
the idea to use `Internal` for the low-level bits in libtorch-ffi wasn’t to your liking?

junjihashimoto:
Let me conform the `Internal` usage.
Current tree is here. (This is the same as https://hasktorch.github.io/)

```
ATen
L Cast
L Class
L Const
L GC
L Type
L Managed
L Unmanaged

Data

Torch
L Autograd
L Backend
L Cast
L DType
L Device
L Dimname
L Functions
L Layout
L Managed
L NN
L Random
L Scalar
L Serialize
L Tensor
L TensorFactories
L TensorOptions
L Typed
L Unmanaged
```
Is the case using `Internal` like this?

```
Internal.ATen
L Cast
L Class
L Const
L GC
L Type
L Managed
L Unmanaged

Internal.Torch
L Managed
L Unmanaged

Data

Torch
L Autograd
L Backend
L Cast
L DType
L Device
L Dimname
L Functions
L Layout
L NN
L Random
L Scalar
L Serialize
L Tensor
L TensorFactories
L TensorOptions
L Typed
```

or

```
Data

Torch
L Autograd
L Backend
L Cast
L DType
L Device
L Dimname
L Functions
L Layout
L NN
L Random
L Scalar
L Serialize
L Tensor
L TensorFactories
L TensorOptions
L Typed

Torch.Internal.ATen
L Cast
L Class
L Const
L GC
L Type
L Managed
L Unmanaged

Torch.Internal.Torch
L Managed
L Unmanaged
```

junjihashimoto:
When we use `Internal`, all members of libtorch-ffi's functions may become Internal for hasktorch-package.
When all members are internal and people see libtorch-ffi's module-tree, people may say what does `Internal` means. (If your thinking module-tree is different from my one, let me know.)
If we merge both package and delete libtorch-ffi' package, I feel the `Internal` is OK.

Using LibTorch's prefix is good because people can know the modules from libtorch-ffi(torch's ffi).

PR #258, opened 2019-12-12
Feature/typed optim tests

tscholak:
* add tests for the typed optimizers by turning the optimization test functions into a test suite
* add a more flexible treatment of the typed `grad` function that permits individual model parameters and the loss to have different dtypes and to live on different devices
* add a test suite for the `grad` function that tests the above feature


hasktorch/src/Torch/Typed/Autograd.hs
```diff
@@ -5,38 +5,60 @@
...
-     , ATen.Castable (HList gradients) [D.ATenTensor]
-     )
-  => Tensor device dtype '[]
-  -> HList parameters
-  -> HList gradients
-grad loss inputs = unsafePerformIO $ ATen.cast2
-  Torch.Managed.Autograd.grad
-  loss
-  (hmap' ToDependent inputs)
+class HasGrad a b | a -> b where
```
austinvhuang: This is going to be a pretty foundational typeclass. Do you have any other implementations in mind beyond Parameter and HList?

tscholak: I discussed with @apaszke about, in essence, having an instance for Tensor, too. He thinks having it is an easy way to shoot oneself in the foot, for me the benefits of not having to insist on Parameter outweigh that. libtorch's grad can definitely handle it. if we want it later, it's in this file too, just commented for now. other than that, I don't see any other instances for now.

tscholak:
@junjihashimoto 
re: autograd spec. there is a lot of boiler plate, unfortunately, that obfuscates what's going on. I decided to test automatic differentiation using the n-dimensional Rastrigin function, https://en.wikipedia.org/wiki/Rastrigin_function, for no particular reason other than it's nontrivial, but not too nontrivial. I define a model, `Rastrigin`, that wraps a "stack" of functionally identical Rastrigin "layers". We have a similar construct in the LSTM module @o1lo01ol1o made. Here, the difference is that each "layer" can have a different `dtype` and live on a different `device`. This is on purpose, because I want to test whether or not `grad` can handle that case. The "loss" tensor is just a sum of Rastrigin function values evaluated for each individual layer in the stack. This "loss" can live on yet another `device` and can have yet another `dtype`. There are several parts of the test. The first step is to use `grad` to calculate the gradient of the "loss" with respect to the parameter list in the `Rastrigin` layer stack. In addition, I compute the gradient manually. I picked the Rastrigin function because that is particularly easy. Once I have both the `grad` and the manual gradients, I can compare them: https://github.com/hasktorch/hasktorch/blob/884522afdf005b81d9bb509cf7dbddc2aca4a4f2/hasktorch/test/Torch/Typed/AutogradSpec.hs#L285-L287. At this occasion, I also check that the gradients live on the same `device`s and have the same `dtype`s as the original Rastrigin parameters, https://github.com/hasktorch/hasktorch/blob/884522afdf005b81d9bb509cf7dbddc2aca4a4f2/hasktorch/test/Torch/Typed/AutogradSpec.hs#L283-L284. The `spec` itself, https://github.com/hasktorch/hasktorch/blob/884522afdf005b81d9bb509cf7dbddc2aca4a4f2/hasktorch/test/Torch/Typed/AutogradSpec.hs#L314-L345, is the just checking everything for different Rastrigin model configurations as defined using the `RastriginSpec` helper type. If both cpu and cuda are available, I'm checking if `grad` "works if individual model layers and loss have different dtypes and live on different devices". That's about it.

tscholak:
@junjihashimoto I'm actually not entirely happy with OptimSpec. The problem is that the tests currently fail on CUDA. I am still investigating possible mitigations, but the core problem is that the CUDA random number generator behaves differently than the CPU generator. It produces different initial model parameters (different even for each `dtype`, that is, `Half`, `Float`, and `Double`), and the optimization fails because it is actually not that easy to find the global minimum of, say, the Ackley function (https://en.wikipedia.org/wiki/Ackley_function) from a random initial point. One also has to cherry-pick the number of optimization steps and the `rtol` and `atol` values for the final loss and parameter checks. All this prompts me to believe that it would be better to fix the initial values to something that has been shown to work in pytorch.
@austinvhuang, @apaszke, what do you think?

hasktorch/src/Torch/Typed/Native.hs
```diff
@@ -1987,8 +1987,22 @@ det input = unsafePerformIO $ ATen.cast1 ATen.Managed.det_t input
 -- diagonal :: Tensor device dtype shape -> Int -> Int -> Int -> Tensor device dtype shape
 -- diagonal _input _offset _dim1 _dim2 = unsafePerformIO $ (ATen.cast4 ATen.Managed.diagonal_tlll) _input _offset _dim1 _dim2
 
--- dot :: Tensor device dtype shape -> Tensor device dtype shape -> Tensor device dtype shape
--- dot _input _tensor = unsafePerformIO $ (ATen.cast2 ATen.Managed.dot_tt) _input _tensor
+type family DotDTypeIsValid (device :: (D.DeviceType, Nat)) (dtype :: D.DType) :: Constraint where
+  DotDTypeIsValid '( 'D.CPU, 0)            dtype = ( DTypeIsNotBool '( 'D.CPU, 0) dtype
+                                                   , DTypeIsNotHalf '( 'D.CPU, 0) dtype
+                                                   )
+  DotDTypeIsValid '( 'D.CUDA, deviceIndex) dtype = DTypeIsNotBool '( 'D.CUDA, deviceIndex) dtype
```
tscholak: this is NOT correct and needs to be further constrained to floating point types!

austinvhuang:
I wouldn't fret about not finding the global minimum for the Ackley function (or any test function with pathological local minima). When I added that case originally, failure was the expected behavior. I'd also expect it to fail in pytorch, regardless of initializer choice. 

I would either consider success to be "runs without crashing irregardless of the value" or exclude them from a test spec.

tscholak:
@austinvhuang In parallel, I actually implemented and monitored gradient descent in scipy following https://docs.scipy.org/doc/scipy/reference/tutorial/optimize.html, to see if our hasktorch implementation produces identical results (up to numerical accuracy). I see a lot of value in going beyond the "runs without crashing" test. Any optimizer we implement should be on par with what's already in pytorch, and I'd like our tests to confirm that.

tscholak:
Just to stress this again, the purpose of the test is not necessarily to verify that a given optimizer is able to find the global minimum, but instead to be sure that a given optimizer is performing at its best potential. I therefore think it is fine to cherry-pick the initial conditions, learning-rate, etc. Other implementations elsewhere should then be used to set the expectations for a given optimizer algorithm, and these should be probed in our test.

austinvhuang:
@tscholak I think I see what you mean. I hadn't thought as much about per-step validation, you're right in that case tractability doesn't matter as long as the behavior is identical.

If validating behavior at the step level, could we just take the RNG out of the implementation? Have a set of initial conditions and validate against a fixed known trajectory.

My comment about "runs without crashing" was specifically about the Ackley function and not optimizers in general, but I see I had something different in mind for validating behavior. Replacing the Ackley function with something tractable like a convex quadratic, you could have random initialization each time and still have the spec to check that you consistently arrive within some error of the correct result. This would be less precise than a numerical per-step validation though.

tscholak:
@junjihashimoto I resolved the issues with `dot`, and I will merge this as soon as the tests pass.

with regards to the optimizers, their tests, and the remaining issues for passing the tests on CUDA, I think I will postpone the fix. What I think right now is that I should make a `HasToDevice` type class that has implementations for `Tensor`, `Parameter`, and (model) data types wrapping these. We need this class anyway. once that's done, I will sample the models on CPU and then move them to GPU. that will make sure that the initial conditions are the same.

PR #260, opened 2019-12-16
Move libtorch-ffi to Torch.Internal

junjihashimoto:
- This pr moves some modules that user does not use directly into `Internal`. 
- To avoid existing module-name on hackage, `Data.HList` is renamed to `Torch.HList` . 
- To mach pytorch's `torch.nn.functional`, `Functions` is renamed to `Functional`.

This pr provides following module-tree.

```
Torch
L HList (from Data.HList)
L Autograd
L Backend
L Cast
L DType
L Device
L Dimname
L Functional  (from Torch.Functions)
   L Internal  (from Torch.Functions.Native)
L Layout
L NN
L Random
L Scalar
L Serialize
L Tensor
L TensorFactories
L TensorOptions
L Typed
   L Functional  (from Torch.Typed.Native)
Torch.Internal
L Cast
L Class
L Const
L GC
L Type
L Unmanaged
L Managed
  L Autograd (from Torch.Managed.Autograd)
  L TensorFactories (from Torch.Managed.Native)
  L Serialize (from Torch.Managed.Serialize)
  L Native (from ATen.Managed.Native)
  L Type (from ATen.Managed.Type)
```


hasktorch/src/Torch/Tensor.hs
```diff
@@ -25,19 +25,19 @@ import Data.Proxy
...
+import qualified Torch.Internal.Managed.Type.Context as ATen
+import qualified Torch.Internal.Managed.Type.Tensor as ATen
+import qualified Torch.Internal.Managed.Type.TensorOptions as ATen
+import qualified Torch.Internal.Managed.Type.StdArray as ATen
+import qualified Torch.Internal.Managed.Type.StdString as ATen
+import qualified Torch.Internal.Managed.Native as ATen
+import qualified Torch.Internal.Managed.Cast as ATen
+import qualified Torch.Internal.Type as ATen
+import qualified Torch.Internal.Const as ATen
+import qualified Torch.Internal.Managed.TonsorFactories as LibTorch
```
tscholak: typo:
```suggestion
import qualified Torch.Internal.Managed.TensorFactories as LibTorch
```

junjihashimoto: Thx!

junjihashimoto:
Thank you, too.

PR #262, opened 2019-12-26
add HasToDevice class and instances

tscholak:
~this is wip~
This is ready for review!

I implemented two mechanisms:
One for moving (parts of) model ADTs from one device to different one and another mechanism for changing the data type of (parts of) a model ADT.
These mechanisms have very similar implementations and generalize `toDevice` and `toDType`, respectively.
They work kind of like a lens, and they need a `Generic` instance for the model data type. Furthermore, the model data type must be parameterized over at least one device type or at least one dtype type, respectively. I used type families to walk the type parameters of the model data type until I find the device or dtype I want to change.
This idea came to me after reading the generic-lens paper by Csongor Kiss et al.

### DIFF
--- a/hasktorch/src/Torch/Typed/Parameter.hs
+++ b/hasktorch/src/Torch/Typed/Parameter.hs
@@ -30,7 +30,7 @@ import           GHC.Generics
 
 import qualified Torch.NN                      as A
 import qualified Torch.Autograd                as A
-import qualified Torch.Tensor                  as A
+import qualified Torch.Tensor                  as D
 import qualified Torch.DType                   as D
 import qualified Torch.Device                  as D
 import           Torch.Typed.Aux
@@ -38,14 +38,14 @@ import           Torch.Typed.Factories
 import           Torch.Typed.Functional
 import           Torch.Typed.Tensor
 
-newtype Parameter (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (shape :: [Nat]) = Parameter A.IndependentTensor
-  deriving (Show)
+newtype Parameter (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (shape :: [Nat]) = UnsafeMkParameter A.IndependentTensor
+  deriving Show
 
 toDependent
   :: forall shape dtype device
    . Parameter device dtype shape
   -> Tensor device dtype shape
-toDependent (Parameter t) = UnsafeMkTensor $ A.toDependent t
+toDependent (UnsafeMkParameter t) = UnsafeMkTensor $ A.toDependent t
 
 data ToDependent = ToDependent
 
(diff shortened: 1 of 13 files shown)


hasktorch/src/Torch/Typed/Parameter.hs
```diff
@@ -30,22 +30,22 @@ import           GHC.Generics
...
 import qualified Torch.DType                   as D
 import qualified Torch.Device                  as D
 import           Torch.Typed.Aux
 import           Torch.Typed.Factories
 import           Torch.Typed.Functional
 import           Torch.Typed.Tensor
 
-newtype Parameter (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (shape :: [Nat]) = Parameter A.IndependentTensor
-  deriving (Show)
+newtype Parameter (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (shape :: [Nat]) = UnsafeMkParameter A.IndependentTensor
```
austinvhuang: Just wondering - what's the reasoning for marking this constructor as unsafe?

hasktorch/test/Torch/Typed/OptimSpec.hs
```diff
@@ -240,44 +241,46 @@ instance
  where
   apply GDConvQuadSpec _ _ = do
     ATen.manual_seed_L 123
-    initModel <- A.sample (ConvQuadSpec @features @dtype @device)
-    let initOptim = mkGD
-        a         = eyeSquare @features @dtype @device
-        b         = zeros @'[features] @dtype @device
-        loss model = convQuad model a b
+    initModel <- A.sample (ConvQuadSpec @features @'D.Float @'( 'D.CPU, 0))
+    let initModel'   = Torch.Typed.Device.toDevice @device @'( 'D.CPU, 0) . Torch.Typed.DType.toDType @dtype @'D.Float $ initModel
```
junjihashimoto: Awesome.
This PR is really useful.

hasktorch/test/Torch/Typed/AutogradSpec.hs
```diff
@@ -234,7 +234,7 @@ instance
   , StandardFloatingPointDTypeValidation device dtype
   ) => Apply' (RastriginA a (Proxy device') (Proxy dtype')) (Parameter device dtype '[n]) (Tensor device' dtype' '[]) where
   apply' (RastriginA a _ _) parameter =
-    toDevice @device' . toDType @dtype' . rastriginLayer' (toDependent parameter) a $ natValI @n
+    Torch.Typed.Tensor.toDevice @device' . Torch.Typed.Tensor.toDType @dtype' . rastriginLayer' (toDependent parameter) a $ natValI @n
```
austinvhuang: Not a huge deal since this is buried in a test spec and not user facing, but the verbosity of the module+function name is starting to hurt readability. Any reason not to import so that the functions can be used directly?

hasktorch/src/Torch/Typed/Parameter.hs
tscholak: Thanks for the question, it’s roughly the same as for the typed `Tensor`: with this constructor you can lift any `IndependentTensor` to a typed Parameter with arbitrary `device`, `dtype`, and `shape`. thus, the runtime values of these and the static values do not have agree which can lead to exceptions and undefined behaviour.

hasktorch/test/Torch/Typed/AutogradSpec.hs
tscholak: we will need to take a better look at this. I did not want to make opinionated decisions about modules, function names, scopes, and imports at this point. the change here was necessary because both the `Tensor` and the `Parameter` module now export `toDevice` and `toDType` functions.

austinvhuang: Could `hiding` on the import of the Parameter module be used to achieve the same thing? I understand though - am fine with coming back to this later once we decide on modules.

tscholak: yes, hiding `toDevice` when importing `Typed.Parameter` would be a way to handle this.

tscholak:
Thank you all for reviewing, I’ll wait a couple of days or so for @stites’ and/or @o1lo01ol1o‘s reply.

tscholak:
I'm going to merge this now. @stites @o1lo01ol1o @apaszke feel free to reopen the discussion

PR #263, opened 2020-01-01
Incorporate example prototypes for untyped initializers + optimizers into the library

austinvhuang:
- [X] Move initializers module to library
- [X] Move optimizers module to library
- [X] Move sgd implementation to optim module, update dependencies
- [X] Make examples use library dependencies
- [x] Migrate prior sgd implementations to conform to the generic optimizer typeclass pattern
- [X] resolve conflicts

Notes for discussion:
- Simplifying randn so there aren't two separate implementations (the older version doesn't use the new representation of RNGs). Not entirely happy with forcing the user to seed / instantiate an RNG to be get going, ideally there's syntactic sugar for a top-level IO version for users that don't care about (or prefer) non-deterministic executions. (see changes to `examples/regression/Main.hs`
- Previously gradients were tensors or `type` aliases for tensors. Should gradients be newtype wrapped tensors so they are distinguished at the type level?
- Same question could be applied to `LearningRate` - should LearningRate be a newtype wrapped tensor?

TODO:
- [ ] Implement initializer testspec
- [ ] Implement optimizer testspec


hasktorch/src/Torch/Optim.hs
```diff
@@ -11,13 +11,26 @@ import Torch.Autograd
 import Torch.NN
 
 type LearningRate = Tensor
+type Loss = Tensor -- TODO - consider newtype wrapping?
 newtype Gradients = Gradients [Tensor] deriving Show
 
 grad' t p = Gradients (grad t p)
```
tscholak: add top level type signature?

hasktorch/src/Torch/Optim.hs
```diff
@@ -11,13 +11,26 @@ import Torch.Autograd
...
 
 class Optimizer o where
     step :: LearningRate -> Gradients -> [Tensor] -> o -> ([Tensor], o)
 
+-- | run a single iteration of an optimizer, returning new parameters and updated optimizer state
+runStep :: (Show p, Parameterized p, Optimizer o) =>
+        p -> o -> Loss -> LearningRate -> IO ([Parameter], o)
+runStep paramState optState lossValue lr = do
```
tscholak: ah, I defined the same function for the typed optimizers :)

hasktorch/src/Torch/Optim.hs
```diff
@@ -37,6 +50,15 @@ gd' lr gradients depParameters dummy = (gd lr gradients depParameters, dummy)
 instance Optimizer GD where
     step = gd'
 
+-- TODO - alias sgd to gd once dependencies are migrated
+-- sgd = gd
+
+sgd :: Tensor -> [Parameter] -> [Tensor] -> [Tensor]
```
tscholak:
```suggestion
sgd :: LearningRate -> [Parameter] -> [Tensor] -> [Tensor]
```

hasktorch/src/Torch/Optim.hs
austinvhuang: This was originally `runIter` in the optimizers example and was a bit messier w/ its responsibilities (had the loss function passed in as a HoF instead of the value, did logging which required iteration information be passed in), rejiggered it to have the same shape as your module's function.

hasktorch/src/Torch/Optim.hs
austinvhuang: Updated, although I'm hoping to deprecate sgd in this form (with Parameters as input) entirely.

Similar to the typed module - optimizer functions in Optim operate on tensors (instead of [Parameter]) and runStep bridges the marshalling between independent/dependent tensors. So for the examples, the typical pattern would be to call runStep with the selected optimizer instead of sgd directly.

I like this decoupling a little better because the optimizers can be used in other contexts outside backprop / computation graphs. As a side benefit this makes the examples less verbose and easier to grok for beginners since it hides the rituals around flattening / dependent tensors from the end user a little bit.

austinvhuang:
If this looks alright - propose going ahead with a merge as a few upstream changes in master are starting to drift. Will submit a separate PR for tests.

austinvhuang:
Thanks @tscholak! That’d be great, I think a lot of this exists in the typed optim module from what I saw.

tscholak:
some of it, yes. `runStep` is there as well.
I was more thinking of the initializer functions. we have some ad-hoc definitions in the typed `LSTM` and `GRU` modules that I'd like to retire, see https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs#L270-L315

PR #267, opened 2020-01-10
Add a sample using dataloader

junjihashimoto:
- [x] Add Dataloader like [pytorch's dataloader](https://pytorch.org/docs/stable/data.html)
- [x] Add stream interface for training loop to use the dataloader.
- [x] Add a sample using CIFAR-10

This pr uses dh-core. See following links.

https://github.com/DataHaskell/dh-core/blob/master/datasets/src/Numeric/Dataloader.hs
https://github.com/DataHaskell/dh-core/blob/master/datasets/src/Numeric/Datasets.hs

The dataloader supports batch, shuffle and transform.
In addition to these, the datasets supports various data-sets including CIFAR-10.


junjihashimoto:
image

austinvhuang:
Thank you @junjihashimoto ! Having a dataloader workflow is a big help.

@stites and @tscholak what do you think since we were discussing this recently? We had also talked about using pipes, but personally I'd be okay with tackling convergence in a separate PR.

junjihashimoto:
http://hackage.haskell.org/package/streaming-0.1.4.3/docs/Streaming-Prelude.html
We can transform stream-type into any stream-types via foldrT.

```
foldrT (\a p -> Streaming.yield a >> p) = id
foldrT (\a p -> Pipes.yield a     >> p) :: Monad m => Stream (Of a) m r -> Producer a m r
foldrT (\a p -> Conduit.yield a   >> p) :: Monad m => Stream (Of a) m r -> Conduit a m r
```

https://hackage.haskell.org/package/streamly-0.7.1/docs/Streamly-Internal-Data-Stream-StreamK-Type.html#t:Stream
I'm not sure if streamly can do this.
```
foldrT (\a p -> Streamly.yield a   >> p) :: (IsStream t, Monad m) => Stream (Of a) m () -> t m a
```

tscholak:
hi, I'm a bit confused about the goal and the problem.
why do we need to convert between streaming types? and which types are we trying to convert between?

junjihashimoto:
Before thinking convertions, dataloader's goal are followings from https://pytorch.org/docs/stable/data.html.

1. Load custom data
2. Preprocess / Transform
3. Support shuffle or randomize for training.
4. Batching
5. Single/Multiprocess

Both datasets and stream which datasets uses support number 1 and 4 mainly.(I know it provides interface for number 2 and 3.)
But number 5 is not supported.

In other hand, streamly supports number 5, but it does not support others.
https://hackage.haskell.org/package/streamly

Convertions between streaming types will help us achieve all goals.

junjihashimoto:
image

https://hackage.haskell.org/package/datasets-0.4.0/docs/src/Numeric.Datasets.CIFAR10.html#parseCifar

This parser is not so fast.

junjihashimoto:
This dataloader has issues for both abstraction and performance.
I will move it into experimental directory.

PR #268, opened 2020-01-10
Add IValue

junjihashimoto:
- [x] Add IValue to support TorchScript.
- [ ] Support all types of https://pytorch.org/docs/stable/jit.html#types


examples/load-torchscript/Main.hs
```diff
@@ -0,0 +1,33 @@
...
+  model <- load "resnet_model.pt"
+  mimg <- readImage "elephant.jpg"
+  case mimg of
+    Left err -> print err
+    Right img'-> do
+      let img = IVTensor $ D.toType D.Float $ hwc2chw img'
+      v <- forward model [img]
+      case v of
+        IVTensor v' -> print $ D.argmax (D.Dim 1) D.RemoveDim v'
+        _ -> print "Return value is not tensor."
```
junjihashimoto: resnet works, but maskrcnn throws this exception (https://gist.github.com/junjihashimoto/3326d9a45deb6419b5fba81c7ffa0c27) .

The exception may be the same as  one of this discussion. 
https://discuss.pytorch.org/t/fasterrcnn-resnet50-jit-trace/62337/9

junjihashimoto: https://github.com/pytorch/vision/pull/1407
https://github.com/pytorch/vision/blob/v0.5.0/torchvision/csrc/vision.cpp#L48

junjihashimoto: I'v realized what we need to use maskrcnn on haskell .

austinvhuang: Wow nice, you're able to run resnet inferences in Haskell?

junjihashimoto: Yes.  The inferences work as follows. It says a spatula.(Input image is a elephant. )

```
$ stack run load-torchscript
Tensor Int64 [1] [ 831]
```

Now I can run maskrcnn too in Haskell by registering some ops.
https://github.com/pytorch/vision/pull/1843

junjihashimoto: I've added some ivalue-types.
maskrcnn is available on haskell.

tscholak: Oh wow, I didn't see that you were working on this again. This is awesome!

junjihashimoto:
Now this pr does not support all types of https://pytorch.org/docs/stable/jit.html#types .
I want to support remained types on the next PR, because the code is long.

hasktorch/src/Torch/Script.hs
```diff
@@ -0,0 +1,251 @@
...
+  | IVTensorList [Tensor]
+  | IVBlob -- Blob
+  | IVGenericList [IValue]
+  | IVGenericDict [(IValue,IValue)]
+  | IVFuture -- Future
+  | IVDevice -- Device
+  | IVObject -- Object
+  | IVUninitialized
+  | IVCapsule -- Capsule
+  deriving (Show)
```
tscholak: 😍

hasktorch/src/Torch/Script.hs
```diff
@@ -0,0 +1,251 @@
...
+  | IVString String
+  | IVTensorList [Tensor]
+  | IVBlob -- Blob
+  | IVGenericList [IValue]
+  | IVGenericDict [(IValue,IValue)]
+  | IVFuture -- Future
+  | IVDevice -- Device
+  | IVObject -- Object
+  | IVUninitialized
+  | IVCapsule -- Capsule
```
junjihashimoto: These types like Capsule are not supported.

PR #272, opened 2020-01-16
data-parallel training of transformer language models on multiple devices

tscholak:
this is ready for review now.

I managed to implement a working* typed transformer language model that can be trained from pre-tokenized files on multiple devices. And I implemented a bunch of things to facilitate that along the way.


junjihashimoto:
Awesome! Exiciting work!

hasktorch/src/Torch/Typed/Device.hs
```diff
@@ -83,3 +112,123 @@ instance (GHasToDevice device' device f g) => GHasToDevice device' device (M1 i
...
+--   ) => 
+
+-- class HasCat fs g | fs -> g where
+--   cat :: HList fs -> g
+
+class HasScatter devices' device f gs | devices' device f -> gs where
+  scatter :: f -> HList gs
+
+instance
+  ( chunks ~ ListLength devices'
```
junjihashimoto: This property requires the number of devices to be the same length as input-chunks.

tscholak: yes, that's true

junjihashimoto: When the tensor's size is 100,   the size of tensor is 50 as follows.

-- >>> :type scatter @'[ '( 'D.CUDA, 0), '( 'D.CUDA, 1)] (ones @'[100,1] @'D.Float @'( 'D.CPU, 0))
  -- scatter @'[ '( 'D.CUDA, 0), '( 'D.CUDA, 1)] (ones @'[100,1] @'D.Float @'( 'D.CPU, 0))
  --  :: HList
  --       '[Tensor '( 'D.CUDA, 0) 'D.Float '[50, 1],
  --         Tensor '( 'D.CUDA, 1) 'D.Float '[50, 1]]

hasktorch/src/Torch/Typed/NN.hs
```diff
@@ -39,24 +41,38 @@ import           Torch.Typed.Factories
...
-
-data LinearSpec (inputFeatures :: Nat) (outputFeatures :: Nat)
-                (dtype :: D.DType)
-                (device :: (D.DeviceType, Nat))
+import           Torch.Typed.Device
+
+class HasForward f a b | f a -> b where
+  forward :: f -> a -> b
+  forwardStoch :: f -> a -> IO b
+  forwardStoch = (pure .) . forward
```
junjihashimoto: When we can only define forwardStoch, do we define forward = undefined?
This does not have to be changed.

hasktorch/src/Torch/Typed/Device.hs
junjihashimoto: If the tensor's size is 101, Are the sizes of tensor 50 and 51?

hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -1713,37 +1713,39 @@ type Chunk chunks dim shape dtype device = ChunkCheck shape dim (ChunkImpl (Chun
...
---          Tensor '( 'D.CPU, 0) 'D.Float '[3, 4]]
+--
+-- -- >>> :type chunk @6 @0 (ones :: CPUTensor 'D.Float '[19, 4])
+-- -- chunk @6 @0 (ones :: CPUTensor 'D.Float '[19, 4])
+-- --   :: HList
+-- --        '[Tensor '( 'D.CPU, 0) 'D.Float '[4, 4],
+-- --          Tensor '( 'D.CPU, 0) 'D.Float '[4, 4],
+-- --          Tensor '( 'D.CPU, 0) 'D.Float '[4, 4],
+-- --          Tensor '( 'D.CPU, 0) 'D.Float '[4, 4],
+-- --          Tensor '( 'D.CPU, 0) 'D.Float '[3, 4]]
```
junjihashimoto: Nice!

hasktorch/src/Torch/Typed/Device.hs
tscholak: yes, as defined by the `Chunk` type class, see https://github.com/hasktorch/hasktorch/blob/9b576b67bff54badf3fee8e470f6c652f832923c/hasktorch/src/Torch/Typed/Functional.hs#L1712 for the implementation.

there are cases in which `chunk` returns fewer chunks than requested, for instance, when you try to split a tensor of `'[2]` into 3 pieces. in this case, though, this instance wouldn't be available, because `chunks ~ ListLength devices'` would not hold.

hasktorch/src/Torch/Typed/Functional.hs
tscholak: well, it was nice before. for some reason, ghci refuses to expand these types now (I'm not sure what has changed that would cause that), and the doctests here failed. that's why I commented them for now.

hasktorch/src/Torch/Typed/NN.hs
tscholak: I'm trying something new here. not sure this will stay like this.
I think that `forward` should always be effect-free, while `forwardStoch` can have effects.
I don't believe there are any models people use that are effectful once the randomness is turned off. Therefore, `forward` can probably be always defined.
For instance, note how I defined a `HasForward` instance for `Dropout`:
```Haskell
instance HasForward Dropout (Tensor device dtype shape) (Tensor device dtype shape) where
  forward dropout input = unsafePerformIO $ Torch.Typed.NN.dropout dropout False input
  forwardStoch dropout input = Torch.Typed.NN.dropout dropout True input
```

hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -2455,23 +2456,9 @@ linear weight bias input = unsafePerformIO $ ATen.cast3 ATen.Managed.linear_ttt
...
-                       (CheckMatMul
-                           shape
-                           '[inputFeatures, outputFeatures]
-                           (ComputeMatMul
-                             (ReverseImpl shape '[]) '[outputFeatures, inputFeatures]))
-                       '[])
-                     '[outputFeatures])
+  :: forall (inputFeatures :: Nat) (outputFeatures :: Nat) (shape :: [Nat]) (shape' :: [Nat]) dtype device (shape'' :: [Nat])
+   . ( shape'' ~ MatMul shape '[inputFeatures, outputFeatures]
+     , shape' ~ Broadcast shape'' shape''
```
austinvhuang: This is much simpler, whyis it the previous version have these check operations interleaved and this one doesnt?

hasktorch/src/Torch/Typed/NN.hs
austinvhuang: Since randomness isn't the only possible effect (though maybe the most common one), what about forwardIO or forwardM or some other name that reflects the generality?

austinvhuang: Reminds me, this comes up in VAEs too.

tscholak: `forwardEffectfully`?

hasktorch/src/Torch/Typed/Functional.hs
tscholak: the bounds of the previous version were just copied from ghci, I think I used `:type linear'` and copy&pasted that in here. The new bounds are equivalent. If you look up the definition of `MatMul` and `Broadcast`, you'll see that they'll become identical to the previous ones if you expand them.

tscholak:
@austinvhuang @junjihashimoto since I have your attention, maybe you can help me to find a good solution for the build problems on the non-nix platforms. we've got a new dependency, `pipes-text`, for the transformer example program. The maintainer of `pipes-text` has not answered any requests for bumping its dependencies in a while, so I had to do it by patching its cabal file (https://github.com/hasktorch/hasktorch/blob/9b576b67bff54badf3fee8e470f6c652f832923c/nix/shared.nix#L114-L131). How can we do this for the other platforms? I'd rather avoid adding this package as a git submodule.

austinvhuang:
@tscholak  one option might be to create a fork repo in the organization and link to that (I changed your status to owner so you should have permissions for that). It doesn't have to be a submodule as long as the commit hash is specified I think.

I of recall seeing some discussion recently that the package ecosystem around pipes is not maintained so well these days. Would be too bad if that's the case since it's the streaming package I'm most familiar with.

junjihashimoto:
I think fork is acceptable, too. Both cabal and stack support git.
Even if the package is not maintained so long time, we can update it someday.
https://wiki.haskell.org/Taking_over_a_package
I really do not want to do such things.

PR #273, opened 2020-01-17
Matrix factorisation example

saeedhadikhanloo:


tscholak:
Thanks for you contribution, this is awesome!
I will have a closer look when I come home tonight.

examples/matrice-factorisation/Main.hs
```diff
@@ -0,0 +1,65 @@
...
+    pure $ MatrixFact u v
+
+flattenParameters :: MatrixFact -> [Parameter]
+flattenParameters (MatrixFact u v) = [u, v]
+
+replaceParameters :: MatrixFact -> [Parameter] -> MatrixFact
+replaceParameters state params_list = MatrixFact u v
+  where
+    u = params_list !! 0
+    v = params_list !! 1
```
tscholak: it *should* be possible to use `Parameterized` from `Torch.Parameter`

examples/matrice-factorisation/Main.hs
```diff
@@ -0,0 +1,65 @@
...
+lossMF :: MatrixFact -> Tensor
+lossMF mf = mse_loss (mulMF mf) r
+  where
+    r = asTensor ([[1, 2, 3], [4, 5, 6], [7, 8, 9]] :: [[Float]])
+
+num_iters = 2000
+
+main :: IO ()
+main = do
+  init <- sample $ (MatrixFactSpec 3 4)
```
tscholak: you *may* want to seed the random number generator so that the run is reproducible, see https://github.com/hasktorch/hasktorch/blob/master/examples/static-mnist-mlp/Main.hs#L143 for an example of that.
we have been working towards exposing the RNG so that one can pass it around as a value, but that's not ready yet.

tscholak: another thought: it seems the number of latent factors `k` is 4 and thus larger than the rank of your matrix `r` above. this is probably not a good choice, since typically you want to use use fewer factors than `rank r`.

tscholak: furthermore, `r` is fully populated. I would suggest that, maybe in another PR, you look at the more practically relevant case in which `r` is sparse. you can then split `r` in a training and validation set, where the validation task is to reconstruct the missing values.

austinvhuang:
Nice!

Would be nice to have other scalable formulations of basic recommendation algos as examples in the future (not required for this PR) like the [simple dense net](https://blog.fastforwardlabs.com/2018/04/10/pytorch-for-recommenders-101.html) or [wide-and-deep](https://ai.googleblog.com/2016/06/wide-deep-learning-better-together-with.html).

examples/matrice-factorisation/Main.hs
saeedhadikhanloo: Your points are great.
* I will make the seed fixed such that it becomes reproducible.
* I guess the optimal value for `k` will be `rank(r)` (not sure though).
* Sparse `r` will be dfinitely more interesting and closer to the applications.

examples/matrice-factorisation/Main.hs
saeedhadikhanloo: @tscholak 
I am not sure yet how to do it. Will look at it and come back to you.

saeedhadikhanloo:
@tscholak 
Thanks for the review. I will apply your points and come back to you again.

saeedhadikhanloo:
@austinvhuang 
Those are definitely good options for future PR. Thanks!

examples/matrice-factorisation/Main.hs
tscholak: 1. Great, thank you!
2. In the general case, in which `r` is sparse, the optimal value will probably be smaller than `rank r`. I would treat `k` as a hyper parameter. You can leave it as it though for now.
3. Yes, that’s no problem. I would have suggested that as well.

Cheers and thank you!

examples/matrice-factorisation/Main.hs
saeedhadikhanloo: I have not succeeded to write `MatrixFact` as an instance of `Parameterized`.

tscholak: ok, never mind, you did an amazing job nonetheless 😃

examples/matrice-factorisation/Main.hs
```diff
@@ -0,0 +1,105 @@
...
+    ( [ [1, 2, 3],
+        [4, 5, 6],
+        [7, 8, 9]
+      ] ::
+        [[Float]]
+    )
+
+lossMF :: MatrixFact -> Tensor
+lossMF mf = mse_loss (mulMF mf) r
+  where
```
tscholak: delete this line please

examples/matrice-factorisation/Main.hs
```diff
@@ -0,0 +1,105 @@
...
+
+main :: IO ()
+main = do
+  -- The rank of r is 2 and we therefore choose k equal to 2.
+  -- The gradient descent does not converge for k equal to 1, as expected.
+  manual_seed_L 123
+  init <- sample $ (MatrixFactSpec n m k)
+  trained <- foldLoop init num_iters gdBlock
+  printParams trained
+  pure ()
```
tscholak: you don’t need this line, since printParams returns IO ()

examples/matrice-factorisation/Main.hs
saeedhadikhanloo: Thanks!

examples/matrice-factorisation/Main.hs
saeedhadikhanloo: Done.

examples/matrice-factorisation/Main.hs
saeedhadikhanloo: Right. It is removed.

saeedhadikhanloo:
@tscholak 
Thanks for your quick feedbacks!
I am done with this PR but it seems I have limited writing access and cannot merge it. It would be great if you could do it. There is only one commit, you can rebase if you want.

PR #275, opened 2020-01-24
example: matrix factorisation with missing values

saeedhadikhanloo:


examples/matrix-factorisation/SparseRatingMatrix.hs
```diff
@@ -0,0 +1,92 @@
...
+
+instance Randomizable RatingsSpec Ratings where
+  sample (RatingsSpec n m l) = do
+    ratings <- randn' [1, l]
+    users <- randint' 1 n [1, l]
+    items <- randint' 1 m [1, l]
+    let list_ratings = toDouble <$> (getEntryByIndex ratings 0) <$> [0 .. (l -1)]
+        list_users = toInt <$> (getEntryByIndex users 0) <$> [0 .. (l -1)]
+        list_items = toInt <$> (getEntryByIndex items 0) <$> [0 .. (l -1)]
+    pure $ Ratings [Rating u i r | (u, i, r) <- zip3 list_users list_items list_ratings]
```
tscholak: while this is fine, IMO it would be more efficient if you defined a batched type, e.g.
```haskell
data RatingBatch = RatingBatch Items Users RatingValues deriving (Show)
newtype Items = Items Tensor deriving (Show)
newtype Users = Users Tensor deriving (Show)
newtype Ratings = Ratings Tensor deriving (Show)
```
the you wouldn't need to zip here but just assemble the data structure.

examples/matrix-factorisation/SparseRatingMatrix.hs
```diff
@@ -0,0 +1,92 @@
...
+
+type TensorEntries = Double
+
+lossMF :: MatrixFact -> [Rating] -> Tensor
```
tscholak:
```suggestion
lossMF :: MatrixFact -> RatingBatch -> Tensor
```

examples/matrix-factorisation/SparseRatingMatrix.hs
```diff
@@ -0,0 +1,92 @@
...
+lossMF (MatrixFact u v) rating_list =
+  let uv = mulMF (MatrixFact u v)
```
tscholak: you are computing the full matrix `uv` here. that may not always be possible due to its size.
you want a way to only compute the elements of `uv` that correspond to `users` and `items` in `RatingBatch`.

examples/matrix-factorisation/SparseRatingMatrix.hs
```diff
@@ -0,0 +1,92 @@
...
+      mse_loss' sum (Rating user item rating) = sum + (uv_ui - tensor_rating) ^ 2
+        where
+          uv_ui = getEntryByIndex uv user item
+          tensor_rating = asTensor rating
+   in foldl mse_loss' (asTensor (0 :: TensorEntries)) rating_list
```
tscholak: once you have a way to compute `uv` only for the `users` and `items` in `RatingBatch`, you can use the standard `mse_loss` function to calculate the l2 norm distance.

examples/matrix-factorisation/SparseRatingMatrix.hs
tscholak: I realize that this may not be easy.
in pytorch, there's https://pytorch.org/docs/stable/torch.html#torch.index_select.
I *think* the equivalent in hasktorch is https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Functional/Internal.hs#L336, but I'm not exactly sure.

tscholak: once you have that, you can use that
```latex
R_{ij} = \sum_l U_{il} V{lj}
```
that is, you'll have a bunch of vector multiplications.

tscholak: that can be written as a matrix multiplication as well, I just realize.
just use index select on `u` and `v` with `users` and `items`, respectively, and then `matmul` those together. that will give you the truncated `R` you need I think.
then you reshape it to a 1-dim tensor of length `batchSize` which you can feed into `mse_loss`.

tscholak: never mind, that's not the case ^^^

tscholak: run a batched vector multiplication, `bmv`, instead.

examples/matrix-factorisation/SparseRatingMatrix.hs
saeedhadikhanloo: Thanks @tscholak for your comment.
I just wanted to recall that I assumed that the batches are subsamples of our main data. I guess one of the sources of the slowness of the previous code is the ratings list size, and that is why we consider subsamples (batches) in each round of the gradient descent.
If we follow the above data types, for subsampling of `RatingBatch` we need to do it three times against `Items`, `Users` and `RatingValues` whilst we do it once if we zip them.
Please correct me if I'm wrong.

examples/matrix-factorisation/SparseRatingMatrix.hs
saeedhadikhanloo: What do you think about this?
```haskell
lossMF :: MatrixFact -> [Rating] -> Tensor
lossMF (MatrixFact u v) rating_list =
    let
      mse_loss' sum (Rating user item rating) = sum + (uv_ui - tensor_rating) ^ 2
        where
          uv_ui = matmul (u' @@ (user,()) ) (v' @@ ((),item) )
          u' = toDependent u
          v' = toDependent v
          tensor_rating = asTensor rating
   in foldl mse_loss' (asTensor (0 :: TensorEntries)) rating_list
```

examples/matrix-factorisation/SparseRatingMatrix.hs
saeedhadikhanloo: It is a great point. I tried to change the code such that we only multiply the rows and columns of `u,` and `v` when it is needed.

examples/matrix-factorisation/SparseRatingMatrix.hs
saeedhadikhanloo: I don't still see a clear of way of doing it. I would do it in this way:
1. making a `Double` list including the desired entries of `uv`. 
2. using `asTensor` to make it a tensor
3. looking at the `mse_loss` of the latter Tensor and the ratings.
I suspect if we can take gradients from it (since we use `toDouble` and `asTensor`).

tscholak:
Hey @saeedhadikhanloo, thanks for the updates. I think I should write again in a high-level form what I have in mind.

What I'd like to arrive at is code that uses batched tensor ops whenever possible. I think this can be accomplished in the following way:

1. start with a stream of input data where each item is isomorphic to (user idx, item idx, rating). You don't have tensors at this point yet. everything is represented using Haskell primitive types, e.g., (Int, Int, Float).
2. split the stream into chunks of fixed length n (the batch size) and arrange it in a form isomorphic to ([user id 1, ..., user id n], [item id 1, ..., item id n], [rating 1, ..., rating n]).
3. turn each batch into a tuple of (user id long tensor, item id long tensor, rating float tensor).
4. use the user long tensor of ids, holding [user id 1, ..., user id n], to extract from U the rows [{U_{user id 1, l}}_l, ..., {U_{user id n, l}}_l] in a float tensor. here, l runs from 0 to k-1, where k is the rank of the factorization UV. in this notation {a_l}_l is the row vector with elements a_0 to a_{k-1}.
5. use the item long tensor of ids, holding [item id 1, ..., item id n], to extract from V the columns [{V_{l, item id 1}}_l, ..., {V_{l, item id n}}_l] in a float tensor.
6. compute a batched vector product, using `bmv`, of the extracted rows of U and columns of V, that is, [{sum_l U_{user id 1, l} V_{l, item id 1}, ..., sum_l U_{user id n, l} V_{l, item id n}]. the result is a vector of length n (a tensor of shape [n]) 
7. use `mse_loss` on the result of step 6 and the rating vector of stem 2.
8. backprop and run optimizer step
9. go to 2

examples/matrix-factorisation/SparseRatingMatrix.hs
```diff
@@ -0,0 +1,96 @@
...
+    items_ids = (\x -> x -1) <$> items
+    users_ids = (\x -> x -1) <$> users
+    u' = toDependent u
+    v' = toDependent v
+    u'' = indexSelect u' (0 :: Int) (asTensor (items_ids :: [Int]))
+    v'' = indexSelect v' (1 :: Int) (asTensor (users_ids :: [Int]))
+    u''' = reshape [batch_size, 1, -1] u''
+    v''' = reshape [batch_size, -1, 1] v''
+    uv_ui = toDType (Double :: DType) $ bmm u''' v'''
+    loss = mse_loss uv_ui ratings_tensor
```
tscholak: looks great! does it work?

junjihashimoto:

saeedhadikhanloo:
You're welcome. It was a pleasure to do so.

PR #277, opened 2020-02-11
Release cleanup pass 1 (WIP)

austinvhuang:
- [x] Add dropout ops for untyped
- [x] Add `full` factory for for typed
- [x] Change cadd/csub/cmul/cdiv to addScalar/subScalar/mulScalar/divScalar for typed, update affected modules (test specs, transformer)
- [x] Add informative error message for casting errors of Haskell Tuple -> C++ Tuple (note these errors should only happen as bugs internal to the library)
- [x] Add additional missing untyped ops (conv, chunk, cat, clamp, hyperbolic fns) 
- [x] Make Torch an uber-module that re-exports most untyped modules
- [x] Simplify imports to use `import Torch` instead of importing specific modules
- [x] Basic run-to-completion test for untyped optim
- [x] README cleanups
- [x] Move image utils for MNIST into beginnings of a torchvision-like module
- [x] Comment out gd-field from cabal file for now until we can avoid requiring cairo system installs for all examples
- [x] Clean up some name clashes - randn' in TensorFactories and in Random


tscholak:
LGTM, thank you very much for the cleanup!

tscholak:
tests are failing because of the rnn example btw

junjihashimoto:
Awesome!

junjihashimoto:
deps/pytorch's hash is reverted.

examples/gaussian-process/Main.hs
```diff
@@ -7,12 +7,7 @@ import Control.Monad (foldM)
 
 import Prelude hiding (exp)
 
-import Torch.Tensor
-import Torch.DType (DType (Float))
-import Torch.TensorFactories (eye', ones', rand', randn', zeros')
-import Torch.Functional
-import Torch.Autograd
-import Torch.NN
+import Torch
```
junjihashimoto: 👍

hasktorch/src/Torch/Functional.hs
```diff
@@ -2,10 +2,35 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 
 module Torch.Functional (
-    module Torch.Functional
-  , module Torch.Functional.Internal
```
junjihashimoto: Internal is not exported.
When we need some functions like fft, do we add the test and export it at the time?

austinvhuang: @junjihashimoto what I've done is add the re-export for the function explicitly in Torch.Functional. The re-exports serve as a whitelist -

https://github.com/hasktorch/hasktorch/pull/277/files#diff-6f6f611002067ebc1c3697bd36244d43

austinvhuang:
> deps/pytorch's hash is reverted.

Weird, not sure how/when that happened. Is the hash in the master branch correct?

junjihashimoto:
Yes, master is correct.

austinvhuang:
> Yes, master is correct.

Ok - am out and about atm but will fix the hash this evening.

PR #278, opened 2020-02-11
Print current seed before test

junjihashimoto:
When using a certain seed, test fails. 
But we do not know the seed used in the test.
This pr outputs using seed before test to find out the cause of the error.

https://github.com/hasktorch/hasktorch/runs/437503475?check_suite_focus=true


austinvhuang:
LGTM Thanks @junjihashimoto !

junjihashimoto:
Thx!

PR #279, opened 2020-02-11
Example: Drawing the gradient field

saeedhadikhanloo:


junjihashimoto:
Awesome!

To fix CI-error, how about putting brew-install(brew install cairo
) here?
https://github.com/hasktorch/hasktorch/blob/master/.github/workflows/stack-macos.yaml#L16

austinvhuang:
Nice example. I personally like vega-based vis (like hvega), but I get that arrow-drawing might not be doable.

Could you add a README describing the example which also mentions the cairo dependency?

saeedhadikhanloo:
@austinvhuang 
Sure (sorry, I miss-understood).

junjihashimoto:
Oh, we need this one, too.
> brew install pkg-config

saeedhadikhanloo:
@austinvhuang @junjihashimoto 
Thanks for your comments. I wrote the following lines in README for mac users to run the example:
```
$ brew install pkg-config
$ brew install cairo
```
Actually I don't have a mac machine, so please let me know about it.

stack.yaml
```diff
@@ -21,6 +21,9 @@ extra-deps:
 - template-0.2.0.10@sha256:f822de4d34c45bc84b33a61bc112c15fedee6fa6dc414c62b10456395a868f85
 - inline-c-0.9.0.0
 - inline-c-cpp-0.4.0.0
+- Chart-cairo-1.9.1
+- cairo-0.13.8.0
+- gtk2hs-buildtools-0.13.8.0
```
apaszke: Feels a bit extreme to add those to the main config along with critical dependencies like `inline-c`. They're only used for one example, so maybe there's a way to separate them somehow?

junjihashimoto: How about using a comment?
BTW, these are not needed now.

apaszke: Comment makes sense, unless there's a programatic way to select dependencies.

PR #285, opened 2020-02-22
Add function documentation, clean up function conventions for consistency

austinvhuang:
- [ ] Add documentation to untyped Functional module

### DIFF
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -302,6 +302,22 @@ sigmoid
     -> Tensor -- ^ output
 sigmoid t = unsafePerformIO $ (cast1 ATen.sigmoid_t) t
 
+-- | softmax
+softmax 
+    :: Int -- ^ dimension
+    -> Tensor -- ^ input
+    -> Tensor -- ^ output
+softmax dim input = unsafePerformIO $ (cast3 ATen.softmax_tls) 
+    input dim (dtype input)
+
+-- | logSoftmax
+logSoftmax 
+    :: Int -- ^ dimension
+    -> Tensor -- ^ input
+    -> Tensor -- ^ output
+logSoftmax dim input = unsafePerformIO $ (cast3 ATen.log_softmax_tls) 
+    input dim (dtype input)
+
 -- | sin
 -- Returns a new tensor with the sine of the elements of input.
 sin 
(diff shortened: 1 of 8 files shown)


hasktorch/src/Torch/Functional.hs
```diff
@@ -91,175 +91,407 @@ instance Castable Reduction Int64 where
...
+-- | mean
+-- Returns the mean value of all elements in the input tensor.
+mean 
+    :: Tensor -- ^ input
+    -> Tensor -- ^ output
 mean t = unsafePerformIO $ (cast1 ATen.mean_t) t
 
-std :: Tensor -> Tensor
+-- | std
+-- Returns the standard-deviation of all elements in the input tensor.
```
adamwespiser: "standard deviation"

austinvhuang: Thanks, fixed

hasktorch/src/Torch/Functional.hs
```diff
@@ -455,10 +766,17 @@ conv2d' weight bias stride padding input =
...
-solve :: Tensor -> Tensor -> (Tensor,Tensor)
+solve 
+    :: Tensor 
+    -> Tensor 
+    -> (Tensor,Tensor)
 solve b a = unsafePerformIO $ (cast2 ATen.solve_tt) b a
 
-cholesky_inverse :: Tensor -> Tri -> Tensor
+-- | Solves a linear system of equations with a positive semidefinite matrix to be inverted given its Cholesky factor matrix uu .
+cholesky_inverse
```
championballer: @austinvhuang Should this also be camelCase? Similar to the choleskySolve change made in this PR?

austinvhuang: Yes it should be, just haven't gotten to all the bits to cleanup in this PR.

Basically the current informal convention is basically - code generated (such as the Functional/Internal module and everything in libtorch-ffi) is snake case, user facing is camelCase in line with Haskell convention, but there are still bits here and there to clean up (BCE is another one).

Feel free to add PR to fix other user-facing instances.

austinvhuang:
I'm going to go ahead and merge this so others can add cleanup PRs without overlapping with changes here.

hasktorch/src/Torch/Functional.hs
championballer: Gotcha. Thanks. :)

PR #288, opened 2020-02-22
Documentation for Functional.hs

Chetank99:
Follows https://github.com/hasktorch/hasktorch/pull/285
Added documentation and changes of mse_loss

### DIFF
--- a/examples/matrix-factorization/Main.hs
+++ b/examples/matrix-factorization/Main.hs
@@ -12,7 +12,7 @@ module Main where
 import Control.Monad (foldM, when)
 import MF
 import Torch.Autograd (grad, makeIndependent, toDependent)
-import Torch.Functional (matmul, mse_loss)
+import Torch.Functional (matmul, mseloss)
 import Torch.Internal.Managed.Type.Context (manual_seed_L)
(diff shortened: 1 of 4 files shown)


austinvhuang:
@Chetank99 2 comments:

1. renaming mse_loss (which was indeed a TODO) will also break some other examples, that don't happen to be picked up by the CI. But I've fixed those here anyway https://github.com/hasktorch/hasktorch/pull/285

2. Generally would avoid editing Internal.hs - it's originally code generated by some of @junjihashimoto 's code generation scripts. 

For now, can you revert your mse_loss changes, and we'll go ahead and merge your function documentation? Thanks!

Chetank99:
@austinvhuang Got it! It's done.
But how can we add documentation to  Internal.hs without errors from CI?
Thanks :v:

austinvhuang:
Although it should not break CI if you edit Internal.hs without introducing breaking changes, my recommendation is not to modify Internal.hs since it is generated code, which means that the next time the code generation is script any work you put into editing it disappears.

Note there are some non-deterministic aspects of CI unfortunately because it has external dependencies (package downloads, etc.). That might be what you're seeing.

PR #294, opened 2020-02-27
Library example

o1lo01ol1o:
I ended up writing a `default.nix` that would allow `hasktorch` to be included as a library in a cabal project with the necessary tooling.  I've dropped it into an example in case it helps users get started.


junjihashimoto:
LGTM!

Issue #233, opened 2019-11-04
Add type to almost all functions.


junjihashimoto:
About half of all functions are typed.
See https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/Native.hs.
The rest untyped functions are below.
We try to add type to the rest of the untyped function.
Please choose the functions you want to implement, then send pr.
When you choose them, mark following list.

- [ ] svd
- [ ] cholesky_solve
- [ ] diag
- [ ] affine_grid_generator
- [ ] as_strided
- [ ] batch_norm
- [ ] bilinear
- [ ] binary_cross_entropy_with_logits
- [ ] bincount
- [ ] chain_matmul
- [ ] constant_pad_nd
- [ ] convolution
- [ ] conv_transpose1d
- [ ] cosine_embedding_loss
- [ ] cudnn_affine_grid_generator
- [ ] cudnn_batch_norm
- [ ] cudnn_convolution
- [ ] cudnn_convolution_transpose
- [ ] cudnn_grid_sampler
- [ ] diag_embed
- [ ] diagflat
- [ ] diagonal
- [ ] dot
- [ ] einsum
- [ ] embedding_bag
- [ ] grid_sampler
- [ ] grid_sampler_2d
- [ ] grid_sampler_3d
- [ ] hinge_embedding_loss
- [ ] ger
- [ ] group_norm
- [ ] fft
- [ ] ifft
- [ ] rfft
- [ ] irfft
- [ ] index
- [ ] index_copy
- [ ] index_put
- [ ] instance_norm
- [ ] kl_div
- [ ] kthvalue
- [ ] native_layer_norm
- [ ] fbgemm_linear_int8_weight
- [ ] fbgemm_linear_quantize_weight
- [ ] fbgemm_pack_gemm_matrix_fp16
- [ ] fbgemm_linear_fp16_weight
- [ ] fbgemm_pack_quantized_matrix
- [ ] fbgemm_is_cpu_supported
- [ ] margin_ranking_loss
- [ ] max_pool1d_with_indices
- [ ] mkldnn_convolution
- [ ] miopen_batch_norm
- [ ] miopen_convolution
- [ ] miopen_convolution_transpose
- [ ] miopen_depthwise_convolution
- [ ] miopen_rnn
- [ ] mode
- [ ] mvlgamma
- [ ] narrow
- [ ] native_batch_norm
- [ ] batch_norm_stats
- [ ] batch_norm_elemt
[...]

PR #297, opened 2020-02-29
add topk typed implementation

o1lo01ol1o:
Assuming doctests don't balk at the test, this should be functional.   I needed a `FromJust` type family to work with the existing families which is maybe more principled than writing a dedicated type family at the cost of exposing unnecessary logic.

### DIFF
--- a/hasktorch/src/Torch/Typed/Functional.hs
+++ b/hasktorch/src/Torch/Typed/Functional.hs
@@ -3915,6 +3915,11 @@ type TopK k shape dim = TopKCheck k shape dim (ExtractDim dim shape) (ReplaceDim
 
 type TopKIdx k = If (k <=? 0) 0 1
 
+type family TopKDeviceAndDTypeCheck dtype (device :: (D.DeviceType, Nat)) :: Constraint where 
+  TopKDeviceAndDTypeCheck D.Bool _           = (TypeError (Text "topk is not defined for Bool tensors."))
+  TopKDeviceAndDTypeCheck D.Half '(D.CPU, _) = (TypeError (Text "topk is not defined for Half types on CPU."))
+  TopKDeviceAndDTypeCheck _ _ = ()
+
 
 -- | Returns the k largest (if largest is `True`) elements of the given input tensor along a given dimension.
 --
@@ -3929,7 +3934,7 @@ type TopKIdx k = If (k <=? 0) 0 1
 --
 topk 
   :: forall k dim shape dtype device 
-   . (KnownNat k, KnownNat dim, All KnownNat shape) 
+   . (KnownNat k, KnownNat dim, All KnownNat shape, TopKDeviceAndDTypeCheck dtype device) 
    => Tensor device dtype shape 
    -> Bool -- ^ if we're returning the top k largest (or, if False, the top k smallest)
    -> Bool -- ^ if the resulting k elements are themselves sorted
--- a/hasktorch/src/Torch/Typed/Functional.hs
+++ b/hasktorch/src/Torch/Typed/Functional.hs
@@ -3921,11 +3921,11 @@ type family TopKDeviceAndDTypeCheck dtype (device :: (D.DeviceType, Nat)) :: Con
 
 -- | Returns the k largest (if largest is `True`) elements of the given input tensor along a given dimension.
 --
--- >>> topk @3 @1 (ones :: CPUTensor 'D.Float '[2,3]) True True
+-- >>> topk @3 @1 True True (ones :: CPUTensor 'D.Float '[2,3])
 -- (Tensor Float [2,3] [[ 1.0000   ,  1.0000   ,  1.0000   ],
 --                     [ 1.0000   ,  1.0000   ,  1.0000   ]],Tensor Int64 [2,3] [[ 0,  1,  2],
 --                     [ 0,  1,  2]])
--- >>> topk @0 @1 (ones :: CPUTensor 'D.Float '[2,3]) True True
+-- >>> topk @0 @1 True True (ones :: CPUTensor 'D.Float '[2,3])
 -- (Tensor Float [2,0] [[],
 --                     []],Tensor Int64 [2,0] [[],
 --                     []])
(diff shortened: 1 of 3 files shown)


hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -3905,8 +3905,24 @@ median' input = unsafePerformIO $ ATen.cast3 ATen.Managed.median_tlb
...
+-- (Tensor Float [2,3] [[ 1.0000   ,  1.0000   ,  1.0000   ],
+--                     [ 1.0000   ,  1.0000   ,  1.0000   ]],Tensor Int64 [2,3] [[ 0,  1,  2],
+--                     [ 0,  1,  2]])
+topk 
+  :: forall k dim shape dtype device 
+   . (KnownNat k, KnownNat dim, All KnownNat shape, DimOutOfBoundCheck shape dim, FromJust (ExtractDim dim shape) >= k) 
+   => Tensor device dtype shape 
+   -> Bool -- ^ if we're returning the top k largest (or top k smallest)
+   -> Bool -- ^ if the resulting k elements are themselves sorted
+   -> (Tensor device dtype (FromJust (ReplaceDim dim shape k)), Tensor device dtype (FromJust (ReplaceDim dim shape 1)))
```
tscholak: nice!
alternatively, you could have defined a `TopKCheck` type family, maybe similar to `ChunkCheck`,
```Haskell
type family TopKCheck (shape :: [Nat]) (dim :: Nat) (result :: Maybe a) :: a where
  TopKCheck shape dim Nothing       = DimOutOfBound shape dim
  TopKCheck _     _   (Just result) = result
```
with
```Haskell
type TopK k shape dim = TopKCheck shape dim (ReplaceDim dim shape k)
```
This allows for customization of the error message the compiler produces, to the benefit of the user.

o1lo01ol1o: Ah, yes, that's helpful, I wasn't sure what the threshold was for introducing new type-level ops.  I'll add that in `Functional.hs`

hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -3905,8 +3905,32 @@ median' input = unsafePerformIO $ ATen.cast3 ATen.Managed.median_tlb
 -- argsort :: Tensor device dtype shape -> Int -> Bool -> Tensor device dtype shape
 -- argsort _input _dim _descending = unsafePerformIO $ (ATen.cast3 ATen.Managed.argsort_tlb) _input _dim _descending
 
--- topk :: Tensor device dtype shape -> Int -> Int -> Bool -> Bool -> (Tensor device dtype shape,Tensor device dtype shape)
--- topk _input _k _dim _largest _sorted = unsafePerformIO $ (ATen.cast5 ATen.Managed.topk_tllbb) _input _k _dim _largest _sorted
+type family TopKCheck (k :: Nat) (shape :: [Nat]) (dim :: Nat) (satd :: Maybe Nat) (result :: Maybe a) :: a where
+  TopKCheck _ shape dim _ Nothing       = DimOutOfBound shape dim
+  TopKCheck _ shape dim Nothing _       = DimOutOfBound shape dim
+  TopKCheck k shape dim (Just v) (Just result) = If ( k <=? v ) result (TypeError (Text "k must be less than or equal to the number of elements in the requested dimention and greater than or equal to 1"))
```
tscholak: awesome, this is even better! though, have you tried k=0? it should just work and return an empty tensor.
there’s a typo in the error message, “dimension”.

o1lo01ol1o: fixed typo and now supports k = 0

tscholak:
Thank you @o1lo01ol1o!
I it occurred to me that there is one more thing to clarify: dtype support.
It’s likely that topk won’t work for half precision tensors on cpu, but will work on cuda.
It’s also likely that topk is not implemented for Boolean tensors.
Can you check please?

o1lo01ol1o:
> Thank you @o1lo01ol1o!
> I it occurred to me that there is one more thing to clarify: dtype support.
> It’s likely that topk won’t work for half precision tensors on cpu, but will work on cuda.
> It’s also likely that topk is not implemented for Boolean tensors.
> Can you check please?

Ah right.  `Half` and `Bool` `CPUTensors` fail.  I don't presently have access to a cuda machine and I'm not sure where to look in the `Aten` source to determine support, do you?

tscholak:
@o1lo01ol1o I have a nixos box with cuda at home. I can check later when I get back there.
I don't know if there's a simple way to check support in the ATen sources. My approach has been to run all permutations and then write a type family that encodes the support.
My suggestion would be to use `StandardFloatingPointDTypeValidation device dtype` for now. that is conservative, it allows only `Float` and `Double` on CPU and additionally `Half` on CUDA.

o1lo01ol1o:
>  I can check later when I get back there.

Cool, when you can confirm, I'll add any additional restrictions on top of 03939dc

o1lo01ol1o:
Though if CI has CUDA maybe there could be tests added somewhere?

tscholak:
CI doesn't have CUDA, but there's a test suite that you can contribute to: 
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/test/Torch/Typed/FunctionalSpec.hs#L637-L886
As you can see, tests are also run on CUDA if it is available.

tscholak:
@o1lo01ol1o I've just tried `topk` for all dtypes on both cpu and cuda, and all combinations that are allowed given the `TopKDeviceAndDTypeCheck` type class work as expected.

I'd still like to add unit tests for `topk` before this gets merged. Also, it would be better I think if the order of arguments was changed such that the input tensor is in the last position. that will make function chaining easier on the eyes.

junjihashimoto:
@o1lo01ol1o 
I've marked topk.
https://github.com/hasktorch/hasktorch/issues/233

o1lo01ol1o:
@tscholak I've added a basic test and fixed the argument order.  Let me know if anything else is needed here.

PR #298, opened 2020-03-01
Upsample typed

o1lo01ol1o:
based on #297 for the `If` type family, this adds 2d, typed, upsampling functions.


austinvhuang:
Thanks very much @o1lo01ol1o! My only comment is we're following a convention where the tensor input is the last argument to make it easier to curry configuration parameters (those boolean flags here for example). Could you make that change?

o1lo01ol1o:
@austinvhuang done

austinvhuang:
LGTM, thanks!

PR #300, opened 2020-03-04
Add typed leakyReLU

momohatt:
This PR implements the refined type of LeakyReLU.

### DIFF
--- a/hasktorch/src/Torch/Typed/Functional.hs
+++ b/hasktorch/src/Torch/Typed/Functional.hs
@@ -4081,15 +4081,15 @@ hardTanh min_val max_val input =
 
 -- | leaky relu
 --
--- >>> dtype &&& shape $ leakyRelu (ones :: CPUTensor 'D.Float '[3,2]) 0.01
+-- >>> dtype &&& shape $ leakyRelu 0.01 (ones :: CPUTensor 'D.Float '[3,2])
 -- (Float,[3,2])
 leakyRelu
   :: forall a shape dtype device
    . (D.Scalar a, StandardFloatingPointDTypeValidation device dtype)
-  => Tensor device dtype shape -- ^ input
-  -> a -- ^ negative slope
+  => a -- ^ negative slope
+  -> Tensor device dtype shape -- ^ input
   -> Tensor device dtype shape -- ^ output
-leakyRelu input negativeSlope =
+leakyRelu negativeSlope input =
   unsafePerformIO $ ATen.cast2 ATen.Managed.leaky_relu_ts input negativeSlope
 
 -- | logarithm of the sigmoid
--- a/hasktorch/src/Torch/Typed/Functional.hs
+++ b/hasktorch/src/Torch/Typed/Functional.hs
@@ -4079,9 +4079,6 @@ hardTanh
 hardTanh min_val max_val input =
   unsafePerformIO $ ATen.cast3 ATen.Managed.hardtanh_tss input min_val max_val
 
--- leaky_relu :: Tensor device dtype shape -> Float -> Tensor device dtype shape
--- leaky_relu _input _negative_slope = unsafePerformIO $ (ATen.cast2 ATen.Managed.leaky_relu_ts) _input _negative_slope
-
 -- | leaky relu
 --
 -- >>> dtype &&& shape $ leakyRelu (ones :: CPUTensor 'D.Float '[3,2]) 0.01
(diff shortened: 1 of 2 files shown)


hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -4082,6 +4082,19 @@ hardTanh min_val max_val input =
...
 
+-- | leaky relu
+--
+-- >>> dtype &&& shape $ leakyRelu (ones :: CPUTensor 'D.Float '[3,2]) 0.01
+-- (Float,[3,2])
+leakyRelu
+  :: forall shape dtype device
+   . (StandardFloatingPointDTypeValidation device dtype)
+  => Tensor device dtype shape -- ^ input
+  -> Float -- ^ negative slope
```
momohatt: I'm not sure if we should use `Float` since I've seen others using `D.Scalar` and `Double`. Any thoughts?

junjihashimoto: It is better to test both types.
We test both types for relu here.
https://github.com/hasktorch/hasktorch/blob/42ec1031661f752c45048902bcff447608f07bbc/hasktorch/test/Torch/Typed/FunctionalSpec.hs#L187-L189
https://github.com/hasktorch/hasktorch/blob/8f743f72445a50d5a788909bde9f6741dfa30523/hasktorch/test/Torch/Typed/AuxSpec.hs#L69

junjihashimoto:
@momohatt  
I've marked leakyReLU.
https://github.com/hasktorch/hasktorch/issues/233

hasktorch/src/Torch/Typed/Functional.hs
junjihashimoto: doctest is a flexible. It can test not only value-level but also type-level.
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/Functional.hs#L1016-L1018
But  it is not comprehensive.

momohatt: Thank you, but I think I still don't get it...
My question was about the type of the parameter `negativeSlope` (not the dtype of `input`). Some functions use `Double` and some use `D.Scalar`, and I don't know which one is the most preferable in this case.
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/Functional.hs#L1192
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/Functional.hs#L4104

Also, should I add test to hasktorch/hasktorch/test/Torch/Typed/FunctionalSpec.hs as well? It seems that not all the functions have their tests there.

junjihashimoto: Thank you for pointing it out.
D.Scalar is better.　
Please see suffix of function. The suffix shows types of arguments.

The charactor of p is 'd' which means double.
> ATen.cast3 ATen.Managed.dropout_tdb input p train

In next case, the charactor of beta is 's'  which means scalar.
> softplus beta threshold input = unsafePerformIO $ ATen.cast3 ATen.Managed.softplus_tss input beta threshold

> Also, should I add test to hasktorch/hasktorch/test/Torch/Typed/FunctionalSpec.hs as well? It seems that not all the functions have their tests there.
Yes, you should.

momohatt: I see, thanks!

junjihashimoto: IMHO, FunctionalSpec is better than not implementing it.
However, I think it can be implemented with the next pr.

momohatt: Okay, thanks!

hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -4082,6 +4082,19 @@ hardTanh min_val max_val input =
 -- leaky_relu :: Tensor device dtype shape -> Float -> Tensor device dtype shape
 -- leaky_relu _input _negative_slope = unsafePerformIO $ (ATen.cast2 ATen.Managed.leaky_relu_ts) _input _negative_slope
```
momohatt: Should I remove these lines?

junjihashimoto: You should.

tscholak:
hello @momohatt , this is a fine first contribution, thanks a lot!
There are two suggestions:
1) can you please flip the arguments? for the functions we have worked on in this module, we've typically rearranged arguments such that the `input` tensor is in the last position, as to facility easy chaining with `.`.
2) There is a test suite for this module here, https://github.com/hasktorch/hasktorch/blob/master/hasktorch/test/Torch/Typed/FunctionalSpec.hs#L637-L886. Can you try to add a unit test for this function there please that checks it for all allowed combinations of `dtype` and `device`?

tscholak:
perfect, thank you @momohatt !

momohatt:
Thanks for the review!

PR #310, opened 2020-03-10
Expose function that builds a tensor from a JuicyPixel DynamicImage

MMesch:
This PR adds:

* _feature in Util.Image:_ the `createTensor` function that builds an untyped Tensor from a JuicyPixels `DynamicImage` type is exposed as `fromDynImage`. Use case: preprocess images with JuicyPixels before moving the data into a Torch tensor.

### DIFF
--- a/hasktorch/src/Torch/Utils/Image.hs
+++ b/hasktorch/src/Torch/Utils/Image.hs
@@ -31,27 +31,29 @@ readImage file = do
   dynamic_img <- I.readImage file
   case dynamic_img of
     Left err -> return $ Left err
-    Right img' -> do
-      case img' of
-        I.ImageY8 (I.Image width height vec) -> createTensor width height 1 D.UInt8 vec
-        I.ImageYA8 (I.Image width height vec) -> createTensor width height 2 D.UInt8 vec
-        I.ImageRGB8 (I.Image width height vec) -> createTensor width height 3 D.UInt8 vec
-        I.ImageRGBA8 (I.Image width height vec) -> createTensor width height 4 D.UInt8 vec
-        I.ImageYCbCr8 (I.Image width height vec) -> createTensor width height 3 D.UInt8 vec
-        I.ImageCMYK8 (I.Image width height vec) -> createTensor width height 4 D.UInt8 vec
-        _ -> return $ Left "Unsupported format"
+    Right img' -> fromDynImage img'
+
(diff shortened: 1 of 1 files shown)


hasktorch/src/Torch/Utils/Image.hs
```diff
@@ -31,27 +31,29 @@ readImage file = do
...
+
+fromDynImage :: I.DynamicImage -> IO (Either String D.Tensor)
+fromDynImage image = case image of
+  I.ImageY8 (I.Image width height vec) -> createTensor width height 1 D.UInt8 vec
+  I.ImageYA8 (I.Image width height vec) -> createTensor width height 2 D.UInt8 vec
+  I.ImageRGB8 (I.Image width height vec) -> createTensor width height 3 D.UInt8 vec
+  I.ImageRGBA8 (I.Image width height vec) -> createTensor width height 4 D.UInt8 vec
+  I.ImageYCbCr8 (I.Image width height vec) -> createTensor width height 3 D.UInt8 vec
+  I.ImageCMYK8 (I.Image width height vec) -> createTensor width height 4 D.UInt8 vec
+  _ -> return $ Left "Unsupported format, convert to 8 bit Image first"
```
MMesch: actually, if we use the `convertRGB8` function, we could probably get rid of the `Either`?

hasktorch/src/Torch/Utils/Image.hs
```diff
@@ -31,27 +31,29 @@ readImage file = do
...
   where
     createTensor width height channel dtype vec = do
```
MMesch: Is there a generic function for `Vec -> Tensor`? If not, should we make one?

junjihashimoto: There is not the function. I think we should make it.

hasktorch/src/Torch/Utils/Image.hs
junjihashimoto: If there is no broken implementation making the image-data, we can erase it

junjihashimoto:
Thx!
I will move this function into Torch.Vision, and get rid of the Either by supporting for all image-types.

hasktorch/src/Torch/Utils/Image.hs
junjihashimoto: I did it.
https://github.com/hasktorch/hasktorch/blob/85ea2c9/hasktorch/src/Torch/Vision.hs#L95-L110

PR #314, opened 2020-03-19
Added autograd example

championballer:
Added an example to allow beginners to understand the working of autograd in Hasktorch.


championballer:
@austinvhuang @junjihashimoto Please review this for changes/modifications.

examples/autograd/README.md
```diff
@@ -0,0 +1,20 @@
+# Basic data flow through autograd
+
+This example contains a very simple multi layer model meant to understand how autograd is used in Hasktorch to calculate gradients. The weights are taken to be equal to 1, and no activation has been applied to allow the user to conveniently verify the gradients for themselves. In the absence of the activation being applied and when the weights are equal to 1, the gradients for the respective weights will be equal to the value of input feature they are connected to.
```
apaszke: Can we please add an activation function? It doesn't have to be super complex, but differentiating through a linear computation is not the most impressive thing you can do. If you think that might be helpful, you could include the hand-derivation of the gradient to explain why does the user see some output values.

championballer: Yeah, that would make the example more effective. I will add an activation function, along with a derivation of the gradients.

championballer:
@apaszke This is ready for another review. Please have a look?

austinvhuang:
It's a nice note, but would rather not bloat the repo with pdfs (or binaries in general) unless it's really needed. Could you convert the derivation to mathjax in the readme markdown and rebase the commit history so that the pdf won't be in the repo?

championballer:
Sure. Will do.

championballer:
Sure. Will get this done. Thanks.

championballer:
@austinvhuang The derivation markdown can be correctly viewed [here](https://github.com/championballer/odin/blob/master/Hasktorch/autograd.md). I have been added that in the readme itself.

PR #290, opened 2020-02-23
Added parameter description  to all functions and documentation to some in Functional.hs for cleanup

championballer:

PR #315, opened 2020-03-19
Modified diag, SELU and CELU docstrings 

championballer:
SELU and CELU included latex formats were not being rendered properly (as also mentioned in #290) , plus there was an extra '|' in diag, which has also been removed to maintain uniformity in the docstrings of Functional.hs

### DIFF
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -289,13 +289,13 @@ relu
     -> Tensor -- ^ output
 relu t = unsafePerformIO $ (cast1 ATen.relu_t) t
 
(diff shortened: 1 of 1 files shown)


junjihashimoto:
Probably we can use latex syntax on haddock.
https://haskell-haddock.readthedocs.io/en/latest/markup.html#mathematics-latex

championballer:
@junjihashimoto Pushed changes that should allow latex rendering. Give it a look now?

championballer:
@junjihashimoto Great, realised the changes required to be made for proper rendering. Thanks. 

Pushed changes for the same.

junjihashimoto:
LGTM!
Thank you for your contribution.

Before:
image


After:
image

PR #325, opened 2020-03-25
add some distributions

tycho01:
I wanna make a start toward implementing distributions and the corresponding constraints.
tackling those typed seems kinda ambitious still (typed tensors, potentially type-classes), so I wanna start simple.


junjihashimoto:
Awesome!

hasktorch/src/Torch/Distributions/Distribution.hs
```diff
@@ -0,0 +1,76 @@
...
+    batch_shape :: a -> [Int]
+    event_shape :: a -> [Int]
+    expand :: a -> [Int] -> a
+    support :: a -> Constraint
+    mean :: a -> D.Tensor
+    variance :: a -> D.Tensor
+    sample :: a -> [Int] -> IO D.Tensor
+    log_prob :: a -> D.Tensor -> D.Tensor
+    entropy :: a -> D.Tensor
+    enumerate_support :: a -> Bool -> D.Tensor -- (expand=True)
```
junjihashimoto: Isn't it used to output tensor?  Or isn't statistics-package usefull?
(monad-bayes is based on it, so I want to know why not use it.)
For example, statistics's mean is below.
```
class Distribution d => MaybeMean d where
  maybeMean :: d -> Maybe Double
```
http://hackage.haskell.org/package/statistics-0.15.2.0/docs/Statistics-Distribution.html

tycho01: I actually was trying out an external package, `mwc-random`, in turn using `vector`; I kinda felt like the type conversions were complicating things and it'd be nice if we had a tensor-based version.
I based this on PyTorch, and I guess some of these operations may yield gradients as well, though disclaimer I'm not yet sure of that's useful in any way (and in fact I know `sample` explicitly disabled gradients there -- wasn't sure how to translate that here).

tycho01: so `mean` / `variance` return tensors now. I think potentially parts of these could end up as real numbers, with other parts as NaN instead. Under that presumption I think returning `Tensor` (as done in PyTorch) may make sense.

austinvhuang:
Is the snake case intentional? Some of the codegen stuff originally used snake case for consistency with libtorch, but for anything user facing we try to keep it camel case. Personally I prefer snake case, but for better or worse camel case is the norm in Haskell.

tycho01:
@austinvhuang:
> case

Thanks! It wasn't intentional, I just kinda did a close adaptation of the Python code without putting too much thought into the naming. I've adjusted them now.
(For variables like `probs2d`, that's a capital `2`, I swear. :upside_down_face:)

hasktorch/src/Torch/Functional.hs
```diff
@@ -929,3 +943,19 @@ flattenAll
...
+  -> IO Tensor
+multinomial_tlb t l b =
+  (cast3 ATen.multinomial_tlb) t l b
+
+-- TODO: rename
+bernoulli_t
+  :: Tensor
+  -> IO Tensor
+bernoulli_t =
+  cast1 ATen.bernoulli_t
```
junjihashimoto, line 961: Basically, I am want to merge this pr.
I'm concerned about calling convention.

tycho01: yeah... I guess we can either figure out a plan for the names at #352 first or adjust them after we make a decision there, I'm open to any names here :woman_shrugging:

austinvhuang:
I'm good with merging this. There's some discussion to work out with the naming convention for distributions, but let's tackle that separately.

tycho01:
okay, I've kinda cleaned this PR up now...

austinvhuang:
Thanks @tycho01 ! CI is still reporting a couple errors in Constraints.hs, possibly some remaining function name clashes with the merge.

austinvhuang:
So close to CI being green -
```
Failures:

  test/Torch/Distributions/BernoulliSpec.hs:66:5: 
  1) Torch.Distributions.Bernoulli logProb
       expected: [-0.6749387]
        but got: [-0.6749386]
```
I noticed `allClose` is commented out which seems like it would address this, any particular reason?

tycho01:
```
Torch.Distributions.Bernoulli logProb:
  test/Torch/Distributions/BernoulliSpec.hs:66:5: 

    toList (Just t) `shouldBe` [-0.6749387]

       expected: [-0.6749387]
        but got: [-0.6749386]

    F.allclose (toDynamic t) (D.asTensor [-0.6749387 :: Float]) 0.001 0.001 False `shouldBe` True

       expected: True
        but got: False
```

I seem to be not very good at using `allclose` somehow.
can anyone see how I misused this here?

austinvhuang:
Looks like the shape of t is length 2 while your value is length 1, maybe that's the issue?

```
let t :: Tnsr 'D.Float '[2] = ...
```

tycho01:
hm. considering the `got: [-0.6749386]`, I'm under the impression the `[2]` shape didn't actually hold.
(I'm not actually sure what that means.)

austinvhuang:
Thanks!

PR #326, opened 2020-03-25
add untyped expand/flattenAll

tycho01:
I left a commented version of `flatten` as in the typed module -- no clue what the reasoning was there.


austinvhuang:
LGTM, the bool is named "implicit" here https://pytorch.org/cppdocs/api/classat_1_1_tensor.html?highlight=expand#_CPPv4NK2at6Tensor6expandE11IntArrayRefb still not sure what that means though.

Could you add unit tests for these?

tycho01:
added, see #331

PR #327, opened 2020-03-26
Torch.Functional param order: subject-last

tycho01:
I think there may be some controversial functions in here -- let's discuss which ones we want what way. :)

For the part-tensor-part-scalar ops at least I think we'd probably want tensor param last.
For the non-associative ones there (subtract, divide), we'd probably need versions with either side (both `a` and `b` for e.g. `a - b`) as the tensor, but in that case we may wanna try and ensure the naming of the tensor-only and both half-tensor ops to somehow align to make sense as well...
I haven't put much thought into that yet.


tycho01:
since writing this PR I've had some feedback:
- don't add flipped variants
- don't flip some ops like `gt` -- subject-last would go against our intution of them as mathematical infix operators (also affects other comparison operators, `dot`, ...?)

austinvhuang:
This looks good to me, thanks for the consistency cleanup @tycho01 !

PR #329, opened 2020-03-26
rename all' / any' to allDim / anyDim

tycho01:
names `all'` / `any'` feel sorta inconsistent with `*Dim` names used for sum/min/max/median/mean... I feel the latter are more clear.


austinvhuang:
I'm good with this, can you double check - run + confirm that there are no examples that break due to using the old name?

tycho01:
yep! did a full replace, none of the examples actually use these. if they did, compilation should've failed in CI.

austinvhuang:
Excellent, thanks!

PR #318, opened 2020-03-21
Add tracer of torchscript

junjihashimoto:
- [x] Trace simple adder
- [x] Trace parameterized model
- [x] Build on macos(Probably ghc treats input-file as c language)
- [x] Insert parameters into traced model

mlp's torchscript is below.

```python
op_version_set = 1
class M(Module):
  __parameters__ = []
  def forward(self: __torch__.M,
    argument_1: Tensor,
    argument_2: Tensor,
    argument_3: Tensor,
    argument_4: Tensor,
    argument_5: Tensor,
    argument_6: Tensor,
    argument_7: Tensor) -> Tensor:
    _0 = torch.addmm(argument_3, argument_1, torch.t(argument_2), beta=1, alpha=1)
    _1 = torch.addmm(argument_5, torch.relu(_0), torch.t(argument_4), beta=1, alpha=1)
    _2 = torch.addmm(argument_7, torch.relu(_1), torch.t(argument_6), beta=1, alpha=1)
    return torch.log_softmax(_2, 1, 6)
```

PR #332, opened 2020-03-28
add functions for saving and loading parameters for untyped models

SurajK7:
I found these functions useful for saving and loading weights. Are they any good or just redundant?

### DIFF
--- a/hasktorch/src/Torch/Serialize.hs
+++ b/hasktorch/src/Torch/Serialize.hs
@@ -3,9 +3,22 @@ module Torch.Serialize where
 import qualified Torch.Internal.Managed.Serialize as S
 import Torch.Tensor
 import Torch.Internal.Cast
+import Torch.NN
+import Torch.Autograd
 
 save :: [Tensor] -> FilePath -> IO ()
 save inputs file = cast2 S.save inputs file
 
 load :: FilePath -> IO [Tensor]
 load file = cast1 S.load file
+
+saveParams :: Parameterized f => f -> FilePath -> IO ()
+saveParams model filePath = do 
+    let params =  map toDependent $ flattenParameters model
+    save params filePath
+
+loadParams :: Parameterized b => b -> FilePath -> IO b
+loadParams model filePath = do
+    tensors <- load filePath
+    let params = map IndependentTensor tensors
+    pure $ replaceParameters model params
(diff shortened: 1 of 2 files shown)


junjihashimoto:
LGTM!

SurajK7:
@junjihashimoto building upon this idea, can we convert weights from a pre-trained PyTorch model to a list of tensors as mentioned earlier and load them in a skeleton of the same model in hasktorch, this would facilitate in transfer learning as training models from scratch on imagenet data-set is a bit unfeasible for vision models. Does that sound good?

austinvhuang:
@SurajK7 This kind of simple serialization sort of works ok especially staying within Haskell (analogous to pickling in python) but serialization w/ deeper interop, we're looking into torchscript.

See:
- https://github.com/hasktorch/hasktorch/pull/268
- https://github.com/hasktorch/hasktorch/pull/318

SurajK7:
@austinvhuang I took a look at torchscript example earlier and from what I understood we can load pre-trained models and run inference in hasktorch but fine-tuning and transfer learning doesn't seem feasible via torchscript and I think that will greatly reduce the models' usability.

hasktorch/src/Torch/Serialize.hs
```diff
@@ -3,9 +3,22 @@ module Torch.Serialize where
...
 
 load :: FilePath -> IO [Tensor]
 load file = cast1 S.load file
+
+saveWts :: Parameterized f => f -> FilePath -> IO ()
+saveWts model filePath = do 
+    let wts =  map (toDependent) (flattenParameters model)
+    save wts filePath
+
+loadSavedWts :: Parameterized b => b -> FilePath -> IO b
```
apaszke: _Weights_ is a very specific term used for only some parameters. Can we please call those functions `saveWeights` and `loadWeights`?

SurajK7: Thank you for pointing out, I am at fault, they should be better-called parameters I will fix that.

SurajK7: @apaszke I believe you meant to say something along the line of `saveParams` and `loadParams` right?
Also, can you shed some light on the earlier thread?

> @austinvhuang I took a look at torchscript example earlier and from what I understood we can load pre-trained models and run inference in hasktorch but fine-tuning and transfer learning doesn't seem feasible via torchscript and I think that will greatly reduce the models' usability.

apaszke: Oh yes, good point! I meant `saveParameters` and `loadParameters`. Sorry for the mixup

junjihashimoto:
Thx!

PR #333, opened 2020-03-29
Added Autograd example

championballer:
The derivation markdown can be correctly viewed [here](https://github.com/championballer/hasktorch/blob/Autograd/examples/autograd/README.md).


austinvhuang:
Nice, thanks!

PR #335, opened 2020-03-31
Add pretrained AlexNet

SurajK7:
This PR implements pretrained Alexnet and all other necessary functions needed to implement it.

### DIFF
--- a/examples/alexNet/Main.hs
+++ b/examples/alexNet/Main.hs
@@ -68,7 +68,7 @@ alexnetForward AlexNet{..} input =
     . linear l2
     . relu
     . linear l1
-    . flatten 1 (-1) 
+    . flatten (Dim 1) (Dim (-1)) 
     . adaptiveAvgPool2d (6, 6)
     . maxPool2d (3, 3) (2, 2) (0, 0) (1, 1) False
     . relu
--- a/examples/alexNet/Main.hs
+++ b/examples/alexNet/Main.hs
@@ -14,7 +14,6 @@ import Torch.TensorFactories
 import Torch.TensorOptions
 import Torch.Tensor
 import qualified Torch.DType as D
-import qualified Torch.Functional.Internal as I
 
 data AlexNetSpec = AlexNetSpec { 
     conv1 :: Conv2dSpec,         
@@ -68,7 +67,7 @@ alexnetForward AlexNet{..} input =
     . linear l2
     . relu
     . linear l1
-    . I.flatten 1 (-1) 
+    . flatten 1 (-1) 
     . adaptiveAvgPool2d (6, 6)
     . maxPool2d (3, 3) (2, 2) (0, 0) (1, 1) False
     . relu
--- a/examples/alexNet/Main.hs
+++ b/examples/alexNet/Main.hs
@@ -1,0 +1,91 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RecordWildCards #-}
+
+module Main where
+    
+import GHC.Generics
+
+import Torch.Serialize
+import Torch.NN
+import Torch.Functional
+import Torch.TensorFactories
+import Torch.TensorOptions
+import Torch.Tensor
+import qualified Torch.Functional.Internal as I
+
+data ANSpec = ANSpec { 
+    conv1 :: Conv2dSpec,         
+    conv2 :: Conv2dSpec,         
+    conv3 :: Conv2dSpec,         
+    conv4 :: Conv2dSpec,
+    conv5 :: Conv2dSpec,
+    fc1   :: LinearSpec,              
+    fc2   :: LinearSpec,
+    fc3   :: LinearSpec
+    } deriving (Show, Eq)
+
+spec = 
+    ANSpec 
+        (Conv2dSpec 3 64 11 11 ) 
+        (Conv2dSpec 64 192 5 5 ) 
+        (Conv2dSpec 192 384 3 3 ) 
+        (Conv2dSpec 384 256 3 3 ) 
+        (Conv2dSpec 256 256 3 3 ) 
+        (LinearSpec (6*6*256) 4096) 
+        (LinearSpec 4096 4096) 
+        (LinearSpec 4096 1000)
+
+data AN = AN { 
+    c1 :: Conv2d,
+    c2 :: Conv2d,
+    c3 :: Conv2d,
+    c4 :: Conv2d,
+    c5 :: Conv2d,
+    l1 :: Linear,
+    l2 :: Linear,
+    l3 :: Linear
+    } deriving (Generic, Show)
+
+instance Parameterized AN
+instance Randomizable ANSpec AN where
+    sample ANSpec {..} = AN 
+        <$> sample (conv1)
+        <*> sample (conv2)
+        <*> sample (conv3)
+        <*> sample (conv4)
+        <*> sample (conv5)
+        <*> sample (fc1)
+        <*> sample (fc2)
+        <*> sample (fc3)
+
+an :: AN -> Tensor -> Tensor
+an AN{..} input = 
+    linear l3
+    . relu
+    . linear l2
+    . relu
+    . linear l1
+    . I.flatten 1 (-1) 
+    . adaptiveAvgPool2d (6, 6)
+    . maxPool2d (3, 3) (2, 2) (0, 0) (1, 1) False
+    . relu
+    . conv2d'' c5 (1, 1) (1, 1)
+    . relu
+    . conv2d'' c4 (1, 1) (1, 1)
+    . relu
+    . conv2d'' c3 (1, 1) (1, 1)
+    . maxPool2d (3, 3) (2, 2) (0, 0) (1, 1) False
+    . relu
+    . conv2d'' c2 (1, 1) (2, 2)
+    . maxPool2d (3, 3) (2, 2) (0, 0) (1, 1) False
+    . relu
+    . conv2d'' c1 (4, 4) (2, 2)
+    $ input
+
+main :: IO ()
+main = do
+    model <- sample spec
+    model <-  loadParams model "alexNet.pt"
+    let input = ones [1, 3, 224, 224] defaultOpts
+    print $ an model input
--- a/examples/alexNet/Main.hs
+++ b/examples/alexNet/Main.hs
@@ -7,14 +7,16 @@ module Main where
 import GHC.Generics
 
 import Torch.Serialize
+import Torch.Vision
 import Torch.NN
 import Torch.Functional
 import Torch.TensorFactories
 import Torch.TensorOptions
 import Torch.Tensor
+import qualified Torch.DType as D
 import qualified Torch.Functional.Internal as I
 
-data ANSpec = ANSpec { 
+data AlexNetSpec = AlexNetSpec { 
     conv1 :: Conv2dSpec,         
     conv2 :: Conv2dSpec,         
     conv3 :: Conv2dSpec,         
@@ -26,7 +28,7 @@ data ANSpec = ANSpec {
     } deriving (Show, Eq)
 
 spec = 
-    ANSpec 
+    AlexNetSpec 
         (Conv2dSpec 3 64 11 11 ) 
         (Conv2dSpec 64 192 5 5 ) 
         (Conv2dSpec 192 384 3 3 ) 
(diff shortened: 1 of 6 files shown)


championballer:
Did you get a chance to run inference using this on some popular dataset? Would love to know how the performance was.

SurajK7:
@championballer No, not yet, but since I have transferred weights directly from PyTorch model which was trained on ImageNet so the performance should be [similiar](https://pytorch.org/docs/stable/torchvision/models.html#classification), by performance you mean performance on metrics like top5-accuracy, right? Or something along the line of inference time and such because I am not sure about the latter.

championballer:
Yeah, I meant in terms of inference time.  

On a side note, do you think it would be better to somehow run this on a sample picture of the provided dimensions, to check its label? That would be a better check of the correct loading and performance of the model.

austinvhuang:
LGTM

examples/alexNet/Main.hs
```diff
@@ -0,0 +1,91 @@
...
+        <*> sample (conv2)
+        <*> sample (conv3)
+        <*> sample (conv4)
+        <*> sample (conv5)
+        <*> sample (fc1)
+        <*> sample (fc2)
+        <*> sample (fc3)
+
+an :: AN -> Tensor -> Tensor
+an AN{..} input =
```
apaszke, line 63: nit: Can we please use `alexnet` and `AlexNet` instead of `an`?

examples/alexNet/Main.hs
```diff
@@ -0,0 +1,91 @@
...
+instance Randomizable ANSpec AN where
+    sample ANSpec {..} = AN 
+        <$> sample (conv1)
+        <*> sample (conv2)
+        <*> sample (conv3)
+        <*> sample (conv4)
+        <*> sample (conv5)
+        <*> sample (fc1)
+        <*> sample (fc2)
+        <*> sample (fc3)
```
apaszke, line 60: Note to self: We should write generics for `Randomizable` for modules that don't have parameters

hasktorch/src/Torch/Functional.hs
```diff
@@ -551,9 +551,14 @@ maxPool2d
     -> Bool -- ^ ceil mode
     -> Tensor -- ^ input
     -> Tensor -- ^ output
-maxPool2d kernelSize stride padding dilation ceilMode self =
+maxPool2d (kernelSize0, kernelSize1) (stride0, stride1) (padding0, padding1) (dilation0, dilation1) ceilMode self =
```
apaszke, line 554: Wouldn't it be nicer to have a helper like `asList :: (Int, Int) -> [Int]` and simply apply that to every argument without explicit unpacking?

hasktorch/src/Torch/NN.hs
```diff
@@ -139,3 +139,34 @@ instance Parameterized Linear
...
+conv2d'' layer stride padding input = 
+  Torch.Functional.conv2d' w b stride padding input
+    where
+        w = toDependent (conv2dWeight layer)
+        b = toDependent (conv2dBias layer)
+
+instance Randomizable Conv2dSpec Conv2d where
+  sample Conv2dSpec{..} = do
+      w <- makeIndependent =<< randnIO' [outputChannelSize, inputChannelSize, kernelSize0, kernelSize1]
+      b <- makeIndependent =<< randnIO' [outputChannelSize]
```
apaszke, line 167: Can you please copy the initialization method from PyTorch? This one is unlikely to yield good results.

hasktorch/src/Torch/NN.hs
```diff
@@ -139,3 +139,34 @@ instance Parameterized Linear
...
+      return $ Conv2d w b
+
+instance Parameterized Conv2d
+
+instance Parameterized [Conv2d]
```
apaszke, line 172: I don't think we should be generating this instance FWIW

apaszke:
Also, this is great! Thanks for implementing it!

examples/alexNet/Main.hs
SurajK7: Yeah, sure.

hasktorch/src/Torch/Functional.hs
SurajK7: Yeah, sure will get this done.

hasktorch/src/Torch/NN.hs
SurajK7: Actually, I copied it from the typed implementation of Conv2d, maybe I should change it there as well?

hasktorch/src/Torch/NN.hs
SurajK7: Sure will remove it.

SurajK7: I think, for loading and updating parameters of a layer, it is necessary for the layer to be an instance of `Parameterized` class.  Or maybe I am missing something? Feel free to correct me.

SurajK7: I think, for loading and updating parameters of an untyped layer, it is necessary for the layer to be an instance of Parameterized class, given my implementation of alexnet or I'll have to modify the implementation a bit. Or maybe I am missing something? Feel free to correct me.

examples/alexNet/Main.hs
```diff
@@ -83,9 +88,23 @@ an AN{..} input =
...
     $ input
 
+normalize :: Either String Tensor -> Tensor
+normalize (Right img) = cat 0 [r', g', b']
+    where
+        img' = divScalar (255.0::Float) $ toType D.Float (hwc2chw img)
+        [r, g, b] = I.split (reshape [3, 224, 224] img') 1 0 
+        r' = divScalar (0.229::Float) (subScalar(0.485::Float) r) 
+        g' = divScalar (0.224::Float) (subScalar(0.456::Float) g) 
+        b' = divScalar (0.225::Float) (subScalar(0.406::Float) b)
```
SurajK7, line 98: I am not sure if this the best way to normalize. Any suggestions? @junjihashimoto

hasktorch/src/Torch/NN.hs
```diff
@@ -163,8 +163,9 @@ conv2d'' layer stride padding input =
 
 instance Randomizable Conv2dSpec Conv2d where
   sample Conv2dSpec{..} = do
-      w <- makeIndependent =<< randnIO' [outputChannelSize, inputChannelSize, kernelSize0, kernelSize1]
-      b <- makeIndependent =<< randnIO' [outputChannelSize]
+      w <- makeIndependent =<< kaimingUniform FanIn (LeakyRelu $ Prelude.sqrt 5.0) [outputChannelSize, inputChannelSize, kernelSize0, kernelSize1]
+      uniform <- randIO' [outputChannelSize]
+      b <- makeIndependent =<< pure (mulScalar ((2::Int) * kernelSize0 * kernelSize1) (subScalar (0.5 :: Float) uniform))
```
SurajK7, line 168: @apaszke Is this initialization fine?

SurajK7:
Update: As of now we can run inference on images of dimensions 224 * 224 pixels and of format PNG, bitmap and JPEG.

examples/alexNet/Labels.hs
```diff
@@ -0,0 +1,1007 @@
+module Labels (labels) where
+
+import Data.Array
+
+labels = array (0, 999) $ zip [0 .. 999] list
+
+list =
```
apaszke, line 7: I'm confused. Why do we have this huge file now?

hasktorch/src/Torch/NN.hs
apaszke: I don't see why would you need this instance for `AlexNet`, because you never use a list of `Conv2D`s? Also, if you really need it, then declare the instance in your code and not in the core library.

hasktorch/src/Torch/NN.hs
```diff
@@ -139,3 +139,35 @@ instance Parameterized Linear
 --     return $ Linear{..}
 
 instance Parameterized [Linear]
+
+data Conv2dSpec = 
+  Conv2dSpec {
+    inputChannelSize  :: Int, 
+    outputChannelSize :: Int, 
+    kernelSize0       :: Int, 
+    kernelSize1       :: Int
```
apaszke, line 148: Maybe instead of `0` and `1` suffixes call them `kernelHeight` and `kernelWidth`?

hasktorch/src/Torch/NN.hs
```diff
@@ -139,3 +139,35 @@ instance Parameterized Linear
...
+    } deriving (Show, Eq)
+
+data Conv2d = 
+  Conv2d { 
+    conv2dWeight :: Parameter, 
+    conv2dBias   :: Parameter
+    } deriving (Show, Generic)
+
+conv2d'' :: Conv2d -> (Int, Int) -> (Int, Int) -> Tensor -> Tensor
+conv2d'' layer stride padding input =
```
apaszke, line 158: Why does this have two apostrophes?

hasktorch/src/Torch/NN.hs
apaszke: Weight looks good, but bias seems to be missing a square root?

hasktorch/src/Torch/NN.hs
SurajK7: Sorry, I misunderstood I thought you were talking about the `Conv2d` instance. Yeah, it doesn't make sense to have the list instance there.

examples/alexNet/Labels.hs
SurajK7: Actually, the file contains the labels of 1000 classes from ImageNet dataset. For now, I have placed it in the alexNet directory maybe later we can move it to Torch.Vision, or if unnecessary remove it.

hasktorch/src/Torch/NN.hs
SurajK7: The `Torch` module exports both `Torch.NN` and `Torch.Functional` modules and there are already definitions for `conv2d` and `conv2d'` present in `Torch.Functional` module hence the two apostrophes.

examples/alexNet/Labels.hs
austinvhuang: Would prefer to keep data files out of the repo and out of the history as well so may need to rebase. We can have scripts to download data and metadata as needed, but let's try not to have dataset contents themselves in the repo unless there's a good reason.

hasktorch/src/Torch/NN.hs
austinvhuang: We haven't formalized it quite yet, but a convention we've started using on functions which implement the 'forward' inference on data with paramaterized instances is "xxxForward". For example, have been doing that in this PR with RNNs https://github.com/hasktorch/hasktorch/pull/319/files

SurajK7: I have made the necessary changes. Could you please have a look?

examples/alexNet/Labels.hs
apaszke: Also, it's usually good practice to avoid adding things "just in case", because they might not ever end up being useful and you can accumulate a bunch of dead code this way. Just something to keep in mind for the future!

SurajK7: Thank you for your suggestion, I will abide by it.

SurajK7:
I am thinking of adding a fine-tuning example based on AlexNet. Any suggestions for the data-set to work upon, I am thinking of [Oxford-IIIT Pet Dataset](http://www.robots.ox.ac.uk/~vgg/data/pets/). Thoughts?

junjihashimoto:
Let's try fine-tuning.
I think we have the problem of resource management on GPU.
If you have any problems, please let me know.
I hope you can solve it.

junjihashimoto:
BTW, I think we can merge this pr and open other issue using pet-data.
@SurajK7 Should we merge this?

championballer:
Hey guys, I have a suggestion in regards to flatten function in Functional.hs. @SurajK7 Can you use Dim datatype instead of Int? Will maintain the uniformity in the source file?

SurajK7:
@junjihashimoto Yeah, we can merge this PR, I will open a separate draft PR for the pet data-set example.
Also, @championballer thank you for pointing out I have done the needful.

hasktorch/src/Torch/Functional.hs
```diff
@@ -566,7 +566,15 @@ maxPool2d
...
-        self kernelSize stride padding dilation ceilMode
+        self 
+        (asList kernelSize)
+        (asList stride)
+        (asList padding)
+        (asList dilation)
+        ceilMode
+        where
+            asList :: (Int, Int) -> [Int]
+            asList (a0, a1) = [a0, a1]
```
austinvhuang, line 577: I've run into this elsewhere, would be preferable if we could add tuples to the casting mechanism although in this context it's a pretty minor thing (any thoughts @junjihashimoto ?)

Anyway out of scope to have to resolve in this PR.

Issue #328, opened 2020-03-26
get `Torch.Functional` and `Torch.Typed.Functional` on par with each other


tycho01:
I keep feeling like there's massive gaps in `Torch.Functional` compared to `Torch.Typed.Functional`, and I kinda wanna close those (tho after we figure out #327).

c.f. #233 which may add more typed functions still missing from the untyped API as well.

so I quickly checked -- these are missing from the `Torch.Functional` compared to the typed one:
```
abs
acos
adaptiveAvgPool1d
adaptiveAvgPool2d
adaptiveAvgPool3d
adaptiveMaxPool1d
adaptiveMaxPool2d
adaptiveMaxPool3d
addScalar
addmm
addmv
addr
all
all'
allclose
alphaDropout
any
any'
argmax
argmin
asin
atan
avgPool1d
avgPool2d
avgPool3d
baddbmm
binaryCrossEntropy
bitwiseNot
bmm
broadcastTensors
cat
ceil
celu
cholesky
choleskyInverse
choleskySolve
chunk
clamp
clampMax
clampMin
clone
constantPadNd1d
conv1d
conv2d
conv3d
convTBC
cos
cosh
cudnnIsAcceptable
det
digamma
divScalar
dot
dropout
eig
elu
embedding
emptyLike
enableEigenVectors
erf
erfc
erfinv
exp
expand
expm1
featureAlphaDropout
featureDropout
flattenAll
floor
frac
fullLike
gelu
geqrf
gru
gruCell
hardTanh
inverse
isComplex
isDistributed
isFloatingPoint
isNaN
isNonZero
isSameSize
isSigned
isclose
keepOrDropDimVal
l1Loss
layerNorm
leakyRelu
lgamma
linear
linear'
log
log10
log1p
log2
logDet
logSigmoid
logSoftmax
logSumExp
lstm
lstmCell
maskedFill
matrixPower
max
maxAll
maxDim
maxPool1d
maxPool2d
maxPool3d
maxValues
maybeNatVal
mean
mean'
meanAll
meanDim
median
median'
medianAll
medianDim
min
minAll
minDim
minValues
mish
mkldnnAdaptiveAvgPool2d
mkldnnLinear
mkldnnMaxPool2d
mm
mseLoss
mulScalar
mv
neg
nllLoss
numel
onesLike
orgqr
polygamma
pow
prelu
qScale
qZeroPoint
quantizedMaxPool2d
randLike
randnLike
reducedSVD
reductionVal
[...]

Issue #322, opened 2020-03-23
Keyword arguments


apaszke:
We have a lot of functions that suffer from Int/Bool blindness and I think we should get more serious about this. Haskell doesn't have keyword arguments built in, but there are some reasonable alternatives. See e.g. [this great post by Ed Yang](http://blog.ezyang.com/2010/09/keyword-arguments-in-haskell/). Any opinions?

o1lo01ol1o:
Boolean blindness has a few approaches:  http://oleg.fi/gists/posts/2019-03-21-flag.html

What is `Int` blindness?

apaszke:
Same thing as boolean blindness: we have many functions that take multiple consecutive `Int` arguments.

o1lo01ol1o:
>  Thanks for the reference!

np; oleg's notes are super helpful.  

If the issue is only one of disambiguation between consecutive arguments, then it seems like newtypes are fine:

```haskell
newtype MyFoo = MyFoo {unMyfoo :: Int} 
   deriving stock (Eq,  Show, Ord)
   deriving newtype (Num, Enum, Integral)
```

If the issue is in the "ordinal" semantics of the `Int` in question, you could go as far as a recursive `Enum`: 

```Haskell
data MyOrdinal = NilOrdinal | SomeMeaningfulNotionOfSuccesser MyOrdinal 
   deriving stock (Eq, Ord)

instance Enum MyOrdinal where
```

but this looks a bit too rarefied for my personal tastes.

tycho01:
Yeah, I noticed use of `Dim` and `KeepDim` seemed sorta inconsistent still. Having this applied all the way would be kinda nice, if occasionally verbose-ish compared to Python.

tycho01:
hmm, maybe `Dim` would allow unifying the `withDimName` function variants with their traditional counterparts...?

championballer:
@apaszke, did you guys reach an agreement on how to address this? newtype seems fair game. What do you think?

junjihashimoto:
`named` package may solve this situation.
https://github.com/monadfix/named
It also supports default-values.

PR #337, opened 2020-04-02
Parity between typed and untyped functional API

championballer:
This pull request is to judge reaction towards the pursuit of parity between typed and untyped functional api. If this has favourable reaction, then I will identify and add other functions which are available in typed api but not in untyped api as is identified as a goal in milestone 5 and has been also been identified in #328 in the future.

#322

### DIFF
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -946,7 +946,7 @@ flattenAll t =
 
 -- | smoothL1Loss
 smoothL1Loss
-  :: Int -- ^ reduction
+  :: Reduction -- ^ reduction
   -> Tensor -- ^ target
   -> Tensor -- ^ input
   -> Tensor -- ^ output
@@ -954,7 +954,7 @@ smoothL1Loss reduction target input = unsafePerformIO $ (cast3 ATen.smooth_l1_lo
 
 -- | softMarginLoss
 softMarginLoss 
-  :: Int -- ^ reduction
+  :: Reduction -- ^ reduction
   -> Tensor -- ^ target
   -> Tensor -- ^ self
   -> Tensor -- ^ output
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -943,3 +943,26 @@ flattenAll
   -> Tensor -- ^ output
 flattenAll t =
   unsafePerformIO $ (cast3 ATen.flatten_tll) t (0 :: Int) (-1 :: Int)
+
+-- | smoothL1Loss
+smoothL1Loss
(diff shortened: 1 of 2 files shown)


hasktorch/src/Torch/Functional.hs
```diff
@@ -943,3 +943,26 @@ flattenAll
...
+  -> Tensor -- ^ output
+smoothL1Loss reduction target input = unsafePerformIO $ (cast3 ATen.smooth_l1_loss_ttl) input target reduction
+
+-- | softMarginLoss
+softMarginLoss 
+  :: Int -- ^ reduction
+  -> Tensor -- ^ target
+  -> Tensor -- ^ self
+  -> Tensor -- ^ output
+softMarginLoss reduction target input = unsafePerformIO $ (cast3 ATen.soft_margin_loss_ttl) input target reduction
```
apaszke, line 961: We need to create a data type that encodes the reductions, because it's completely unclear what's reduction 0 and what's reduction 1 (plus, there's no reduction 289346). You'll need to add something along the lines of:
```hs
data Reduction = Option1 | Option2 | ...

instance Castable Reduction Int where
  ...
```

championballer: Alright.

championballer: @apaszke Please have a look now? The datatype for reduction was already present. Had to just use it in the functions.

hasktorch/src/Torch/Functional.hs
```diff
@@ -943,3 +943,26 @@ flattenAll
   -> Tensor -- ^ output
 flattenAll t =
   unsafePerformIO $ (cast3 ATen.flatten_tll) t (0 :: Int) (-1 :: Int)
+
+-- | smoothL1Loss
+smoothL1Loss
+  :: Reduction -- ^ reduction
+  -> Tensor -- ^ target
+  -> Tensor -- ^ input
```
apaszke, line 951: Most other losses take `input` before the `target`, so it's best not to change the convention.

championballer: Yeah, makes sense.

championballer: @apaszke Do you think, we should rather change others, since that would also satisfy the currying principle?

apaszke: I think that in most cases loss functions are applied together to the input and the target, so there's usually no currying involved, and this order is more natural I think. I'd keep it as it used to be.

championballer: Roger that. :)

junjihashimoto:
Thx!

PR #338, opened 2020-04-04
[WIP] Add keyword arguments with default

junjihashimoto:
Case study of keyword arguments using Data.Default.

Other implementations of keyword arguments :
- https://www.reddit.com/r/haskell/comments/f7427x/overloadedlabels/
- https://github.com/monadfix/named
- https://gist.github.com/chrisdone/7dddadd089e6a5d2e3e9445c4692d2c2
- https://stackoverflow.com/questions/7781096/is-there-a-better-way-to-have-optional-arguments-in-haskell

### DIFF
--- a/hasktorch/hasktorch.cabal
+++ b/hasktorch/hasktorch.cabal
@@ -77,6 +77,8 @@ library
                     , safe-exceptions
                     , zlib >= 0.6
                     , inline-c
+                    , data-default
+                    , generic-lens
 
 test-suite spec
   type:               exitcode-stdio-1.0
@@ -120,6 +122,9 @@ test-suite spec
                     , QuickCheck
                     , directory
                     , JuicyPixels
+                    , data-default
+                    , generic-lens
+                    , lens
 
 test-suite doctests
   type:               exitcode-stdio-1.0
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -1,5 +1,11 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE DeriveGeneric #-}
 
 module Torch.Functional (
     module Torch.Functional,
@@ -45,6 +51,8 @@ import qualified Torch.Functional.Internal as Internal
 import Torch.Internal.Cast
 import Torch.Internal.Class
 import Data.Int
+import Data.Default
+import GHC.Generics
 
 import Torch.Scalar
 import Torch.Tensor
@@ -762,17 +770,29 @@ constantPadNd1d padding value input = unsafePerformIO $ cast3
   padding
   value
 
+data Conv1dSpec = Conv1dSpec
+  { stride :: Int
+  , padding :: Int
+  , dilation :: Int
+  , groups :: Int
+  } deriving Generic
+
+instance Default Conv1dSpec where
+  def = Conv1dSpec
+        { stride = 1
+        , padding = 0
+        , dilation = 1
+        , groups = 1
+        }
+
 -- | Applies a 1D convolution over an input signal composed of several input planes.
 conv1d
-  :: Tensor -- ^ weight
+  :: Conv1dSpec
+  -> Tensor -- ^ weight
   -> Tensor -- ^ bias
-  -> Int -- ^ stride
-  -> Int -- ^ padding
-  -> Int -- ^ dilation
-  -> Int -- ^ groups
   -> Tensor -- ^ input
   -> Tensor -- ^ output
-conv1d weight bias stride padding dilation groups input =
+conv1d Conv1dSpec{..} weight bias input =
     unsafePerformIO $ (cast7 ATen.conv1d_tttllll)
         input
         weight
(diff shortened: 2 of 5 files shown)


hasktorch/test/FunctionalSpec.hs
```diff
@@ -126,11 +132,11 @@ spec = do
         kernel1 = 1
         input0 = 5
         input1 = 6
-        x = conv2d'
+        x = conv2d
+              ( def & (field @"stride" .~ (1,1))
```
junjihashimoto, line 136: This `field @"stride" .~ (1,1)` is to avoid DuplicateRecordFields.

hasktorch/src/Torch/Functional.hs
```diff
@@ -762,17 +770,29 @@ constantPadNd1d padding value input = unsafePerformIO $ cast3
...
+  , groups :: Int
+  } deriving Generic
+
+instance Default Conv1dSpec where
+  def = Conv1dSpec
+        { stride = 1
+        , padding = 0
+        , dilation = 1
+        , groups = 1
+        }
```
apaszke, line 786: I've been thinking about this syntax, and I'm not sure if it's that nice. It seems like the expected usage is to say
```haskell
conv2d (def { stride = ..., padding = ...})
```
but that's already quite verbose. I have two ideas for how to improve it.
**Be more verbose** and keep requiring all arguments, but with human-readable names. So you'd probably have `data Stride = UnitStride | Stride Int | Stride2D (Int, Int)`. Similarly there would be `data Padding = NoPadding | Padding Int | Padding2D (Int, Int)`.

**Less verbose** We could have a monoid for options where `mempty` is the default, and then for every option you can define an "overriding element". So for example you could have an expression like `Stride (2, 2) <> Dilation (4, 4)` to be equivalent to `def { stride = (2, 2), dilation=(4, 4) }`.

Here are example calls using both of those (all calls are equivalent):
```haskell
conv2d UnitStride (Padding 4) (Dilation 2) SingleGroup input
conv2d (Padding 4 <> Dilation 2) input
conv2d (Dilation 2 <> Padding 4) input
```

cc @austinvhuang @tscholak

hasktorch/hasktorch.cabal
```diff
@@ -120,6 +122,9 @@ test-suite spec
                     , QuickCheck
                     , directory
                     , JuicyPixels
+                    , data-default
```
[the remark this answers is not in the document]
junjihashimoto: It is not necessary to use it.
I think users are familer with the package, because dozens of packages uses it.
https://www.stackage.org/lts-14.5/package/data-default-0.7.1.1/revdeps

hasktorch/src/Torch/Functional.hs
junjihashimoto: It looks good.
But I'm not sure this (`Stride (2, 2) <> Dilation (4, 4)`) is monoid.(It's trivial)

junjihashimoto: How about this one?
```haskell
{-#LANGUAGE MultiParamTypeClasses#-}
data Stride = Stride Int
data Padding = Padding Int
data Dilation = Dilation Int
data Groups = Groups Int
data Tensor

data Conv1dSpec =
  Conv1dSpec
    Stride
    Padding
    Dilation
    Groups

class Default a where
  def :: a

instance Default Conv1dSpec where
  def =
    Conv1dSpec
      (Stride 1)
      (Padding 0)
      (Dilation 1)
      (Groups 1)

class UpdateSpec a b where
  (+=) :: a -> b -> a

instance UpdateSpec Conv1dSpec Stride where
  (+=) (Conv1dSpec _ b c d) a = Conv1dSpec a b c d
  
instance UpdateSpec Conv1dSpec Padding where
  (+=) (Conv1dSpec a _ c d) b = Conv1dSpec a b c d
  
instance UpdateSpec Conv1dSpec Dilation where
  (+=) (Conv1dSpec a b _ d) c = Conv1dSpec a b c d
  
instance UpdateSpec Conv1dSpec Groups where
  (+=) (Conv1dSpec a b c _) d = Conv1dSpec a b c d

conv1d
  :: Conv1dSpec
  -> Tensor
  -> Tensor
  -> Tensor
  -> Tensor
conv1d = undefined  
  
main = do
  let v = conv1d (def += Stride 2 += Padding 1) undefined  undefined  undefined
  return ()
```

junjihashimoto: UpdateSpec's instance could be derived by Generics automatically.

apaszke: Well you can make the ADT I proposed a monoid. Think of `Padding 2` as an action that takes a `Conv1DSpec` and produces a new `Conv1DSpec` with padding replaced by 2. Composition of such action is associative (it's just function composition). To get the result, you feed this chain with the default set of parameters and out comes the desired setting! I like this much more than any approach that requires you to spell out `def` unnecessarily, because it makes the code quite ugly IMO.

junjihashimoto: Is it like this?
```
data Conv1dSpec
  = Stride Int
  | Padding Int
  | Dilation Int
  | Groups Int
  | Pair Conv1dSpec Conv1dSpec

instance Monoid Conv1dSpec where
  mempty =  Stride 1 <> Padding 0 <> Dilation 1 <> Groups 1
  mappend a b = Pair a b
```

apaszke: Yeah it's effectively something like this. Then we need to interpret those instances into opts, but that's the general idea.

junjihashimoto: In addition to the interpretation of instances, the following duplicated constructors are not allowed.

```
data Conv1dSpec
  = Stride Int
  | Padding Int
  | Dilation Int
  | Groups Int
  | Pair Conv1dSpec Conv1dSpec

data Conv2dSpec
  = Stride (Int,Int)
  | Padding (Int,Int)
  | Dilation Int
  | Groups Int
  | Pair Conv2dSpec Conv2dSpec
```

apaszke: Well we could call them `Stride2D` I guess. Not sure what to do with `Dilation` though 😕

junjihashimoto: Another idea is here.

```haskell
{-# LANGUAGE MultiParamTypeClasses #-}

data Tensor

newtype Stride = Stride Int
newtype Padding = Padding Int
newtype Dilation = Dilation Int
newtype Groups = Groups Int

newtype Stride2d = Stride2d (Int,Int)


data Conv1dSpec
  = Conv1dSpec
    Stride
    Padding
    Dilation
    Groups

class Update a b where
  (<+>) :: a -> b -> a
  (<|>) :: (Update a c) => b -> c -> a -> a
  (<|>) b c a = a <+> b <+> c

instance Update Conv1dSpec Stride where
  (<+>) (Conv1dSpec _ b c d) a = Conv1dSpec a b c d

instance Update Conv1dSpec Padding where
  (<+>) (Conv1dSpec a _ c d) b = Conv1dSpec a b c d

instance Update Conv1dSpec Dilation where
  (<+>) (Conv1dSpec a b _ d) c = Conv1dSpec a b c d

instance Update Conv1dSpec Groups where
  (<+>) (Conv1dSpec a b c _) d = Conv1dSpec a b c d

conv1d
  :: (Conv1dSpec -> Conv1dSpec)
  -> Tensor
  -> Tensor
  -> Tensor
  -> Tensor
conv1d spec = undefined
--  let stride = 
  
main = do
  let v = conv1d ((Stride 2) <|> (Padding 1)) undefined  undefined  undefined
  return ()
```

junjihashimoto: I've come up with a better definition. It is different one.

PR #339, opened 2020-04-05
Modified Functions in untyped api to include Dim datatype instead of Int

championballer:


championballer:
I don't know why that 1 check is failing. It would be great if someone can help me with it.

Local build and tests also worked fine.

junjihashimoto:
LGTM!
Thx!

PR #345, opened 2020-04-11
allow untyped optimization for typed LSTM

tycho01:
I was making a model that included both parts known statically as well as dynamic ones, meaning typed optimization was out for me.
This means I'd need to fall back to untyped optimization, while I was still using the typed LSTM here as well. As a result, I needed to ensure that supported untyped optimization (i.e. instance `A.Parameterized`). This PR adds that.

I'm open to feedback on the function `untypeParam` -- I was under the impression we didn't have something like it yet, and I wanted to use a name that wouldn't clash with `toDynamic`, wouldn't sound like a noun (`dynamicParameter`), and ideally wouldn't be super long, as this may get repeated a bunch in lists (see `LSTMLayer` instance). So I was a bit on the fence, but I'm open to e.g. `untypeParameter` or other ideas if preferred.

### DIFF
--- a/hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
+++ b/hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
@@ -150,6 +150,54 @@ instance
       <*> (makeIndependent =<< pure zeros)
       <*> (makeIndependent =<< pure zeros)
 
+instance A.Parameterized (LSTMLayer inputSize hiddenSize directionality dtype device) where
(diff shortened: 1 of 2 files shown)


austinvhuang:
Any thoughts on this PR @tscholak ?

tscholak:
@austinvhuang it's probably a good addition, haven't looked at the implementation in more detail, though.

what would be nice is a test that checks that these instances do the same as the typed ones

tycho01:
at this point I wonder if I could've just had it automatically derive these :no_mouth:

PR #346, opened 2020-04-11
generalize A.Parameterized instances

KiaraGrouwstra:
I was checking our instantiations for untyped `Parameterized`, and figured we could generalize them a bit.
Ideally, I'd like for us to be able to automatically derive these.
I hope that could for example help replace our manual `Parameterized` instances for our untyped RNN examples (`LSTMCell`, `GRUCell`, `ElmanCell`). That might involve using `Generic`, but I'm not quite sure how that stuff works yet. Would love to get feedback on that. :upside_down_face:


tycho01:
@tscholak thanks for the encouraging words :joy:, the typed API tends to make me feel dumb :sweat_smile: but lemme try a shot at that one as well.
(that's why I'm in program synthesis, to compensate for my own inadequacies. :joy:)

ohh, I see there's a `flattenParameters :: (Generic f, Parameterized' (Rep f)) => f -> [Parameter]`... does that mean we can just have `LSTMCell` derive `Generic` then scrap its manual `Parameterized` instantiation?

tycho01:
Compiler:
> The constraint ‘Scalar a’ is no smaller than the instance head ‘Parameterized a’ (Use UndecidableInstances to permit this)

Uh, I don't know if adding `UndecidableInstances` would make it good or if I did something dumb.

I looked at `HList` module and a potential Parameterized derivation there, and felt smart for a bit appending some ops with `h`, but still feel kinda way in over my head there.

```
instance (Parameterized a, Parameterized (HList as)) => Parameterized (HList (a ': as)) where
  flattenParameters = flattenParameters a `happend` flattenParameters as  -- TODO: join/flatten nested hlist
  -- replaceOwnParameters hparams = hmap replaceOwnParameters  -- TODO: distribute hparams
  -- replaceOwnParameters hparams = replaceOwnParameters a ??? `happend` replaceOwnParameters as ???
```

`flattenParameters` seems conceptually straight-forward, tho I don't quite see a built-in way to join nested `HList`s yet (sounds llike an `HMonad`?).

for `replaceOwnParameters`, the untyped API made it kinda simple with the monadic handling of the params with `Traversable`, but for the typed `Parameterized` this gets more complex -- as the param hlist is passed straight in, you suddenly kinda need to know which parts to pass where, which I don't think I know...

tycho01:
weird, it's complaining about some overlap.

```
  src/Torch/NN.hs:141:10: error:
      • Overlapping instances for Parameterized Linear
          arising from a use of ‘Torch.NN.$dmreplaceOwnParameters’
        Matching instances:
          instance Scalar a => Parameterized a
            -- Defined at src/Torch/NN.hs:59:10
          instance Parameterized Linear -- Defined at src/Torch/NN.hs:141:10
      • In the expression: Torch.NN.$dmreplaceOwnParameters @(Linear)
        In an equation for ‘replaceOwnParameters’:
            replaceOwnParameters = Torch.NN.$dmreplaceOwnParameters @(Linear)
        In the instance declaration for ‘Parameterized Linear’
      |
  141 | instance Parameterized Linear
      |          ^^^^^^^^^^^^^^^^^^^^
  ```

```
  src/Torch/NN.hs:184:10: error:
      • Overlapping instances for Parameterized Conv2d
          arising from a use of ‘Torch.NN.$dmflattenParameters’
        Matching instances:
          instance Scalar a => Parameterized a
            -- Defined at src/Torch/NN.hs:59:10
          instance Parameterized Conv2d -- Defined at src/Torch/NN.hs:184:10
      • In the expression: Torch.NN.$dmflattenParameters @(Conv2d)
        In an equation for ‘flattenParameters’:
            flattenParameters = Torch.NN.$dmflattenParameters @(Conv2d)
        In the instance declaration for ‘Parameterized Conv2d’
      |
  184 | instance Parameterized Conv2d
      |          ^^^^^^^^^^^^^^^^^^^^
```

I don't really see how qualify as [`Scalar`](https://hasktorch.github.io/hasktorch-0.2.0.0/html/Torch-Scalar.html#t:Scalar) tho...

```
data Linear = Linear { weight :: Parameter, bias :: Parameter } deriving (Show, Generic)
```

```
data Conv2d = 
  Conv2d { 
    conv2dWeight :: Parameter, 
    conv2dBias   :: Parameter
    } deriving (Show, Generic)
```

cc @junjihashimoto

tscholak:
hm, no luck here it seems

will have a look as well asap

tycho01:
basically I understand little of `OVERLAPS` / `OVERLAPPABLE` / whatever pragmas as Junji suggested so that's probably most of the reason here :sweat_smile:

PR #347, opened 2020-04-11
add untyped mkAdam

tycho01:


austinvhuang:
You don't want to have a utility function that defaults to zero-ing out parameters. Instead you want to make use of the randomizable instances to use the appropriate initialization for the architecture.

tycho01:
> You don't want to have a utility function that defaults to zero-ing out parameters. Instead you want to make use of the randomizable instances to use the appropriate initialization for the architecture.

@austinvhuang hm, then I guess the typed `mkAdam` is currently also not how we'd want it? cuz I just ported this from that one. :sweat_smile:

austinvhuang:
Can go ahead and merge this as a first cut, but we should revisit this with initializers in mind.

@tscholak any particular reason parameters are zero'd out in the typed version? (guessing convenience but maybe there's more to it).

PR #348, opened 2020-04-11
untyped optim for typed tensor/parameter

KiaraGrouwstra:
I had to use untyped optimization for my model, as not everything is known statically, but where possible still preferred to use typed tensors and parameters. This made it complain typed `Parameter` did not instantiate untyped `Parameterized`, so I wanted to fix that here.

I noticed typed `Parameterized` also handled `Tensor`, so I figured I'd have the untyped one handle that (both untyped and typed `Tensor` while we're at it) as well.

Kinda doubted about where to put stuff so as to avoid circular dependencies... I suppose ideally you'd have all instances in their parent module so all would be imported in one go, tho that gets tricky once typed and untyped modules would start importing one another...


junjihashimoto:
Do you add some tests using these instances?

hasktorch/src/Torch/NN.hs
```diff
@@ -43,6 +44,14 @@ class Parameterized f where
   default replaceOwnParameters :: (Generic f, Parameterized' (Rep f)) => f -> ParamStream f
   replaceOwnParameters f = to <$> replaceOwnParameters' (from f)
 
+instance Parameterized Tensor where
+  flattenParameters _ = []
+  replaceOwnParameters = return
+
+instance Parameterized (Torch.Typed.Tensor.Tensor device dtype shape) where
+  flattenParameters _ = []
+  replaceOwnParameters = return
```
junjihashimoto, line 53: I think this typed instance should be in Torch.Typed.Parameter.

tycho01: Thanks, I moved that one now.

junjihashimoto: Thx!

austinvhuang:
Can we substitute `pure` for `return`? Otherwise this LGTM, thanks @tycho01, apologies for not having a chance to look sooner.

Given its ubiquity, have you noticed any compile time (or runtime) slowdown adding a typeclass to Tensor?

tycho01:
@austinvhuang thanks, I changed the `pure` now.
I hadn't really noticed a difference with this in my own code, though that wasn't as big of a project.
so honest answer, if the impact is significant then hopefully we should notice from the CI times over here.

tycho01:
interesting, some test error:
```
test/OptimSpec.hs:52:24: warning: [-Wsimplifiable-class-constraints]
    • The constraint ‘TensorLike a’ matches
        instance (Data.Reflection.Reifies a Torch.DType.DType,
                  Foreign.Storable.Storable a) =>
                 TensorLike a
          -- Defined in ‘Torch.Tensor’
      This makes type inference for inner bindings fragile;
        either use MonoLocalBinds, or simplify it using the instance
    • In the type signature: extract :: TensorLike a => Parameter -> a
      In an equation for ‘show’:
          show (Rosen x y)
            = show (extract x :: Float, extract y :: Float)
            where
                extract :: TensorLike a => Parameter -> a
                extract p = asValue $ toDependent p
      In the instance declaration for ‘Show Rosen’
   |
52 |             extract :: TensorLike a => Parameter -> a
   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

like, I can replace that, sure.
is that what we want tho?

tycho01:
ohh cool, updating fixed that.

tycho01:
@austinvhuang on performance, looks like the [last](https://github.com/hasktorch/hasktorch/runs/937151173) `cabal-linux` ci run on master took 41m, while this one finished in 36m.
I'm inclined to conclude there was no significant performance impact then.
(`nix-linux` build did finish slower so I'm inclined to conclude CI times are just erratic, randomly finishing faster or slower on similar code, as with `cabal-linux`.)

PR #351, opened 2020-04-12
warn about IndependentTensor vs makeIndependent

tycho01:
I wanted to add this warning based on an error I ran into:
`CppStdException "Exception: One of the differentiated Tensors does not require grad; type: std::runtime_error"`


junjihashimoto:
Thx!

PR #352, opened 2020-04-12
 expose sampling fns in Torch.Functional

tycho01:
forgot about this branch.
added the sampling functions we're skipping in codegen.
I'm open to feedback on function names. :neutral_face:


junjihashimoto:
Exposed functions have internal suffixes like `_t`.
I think we should use  suffix having meaning like `All`.

junjihashimoto:
BTW, I appreciate your PR, because we need those sampling functions.

austinvhuang:
Re: function names, couple approaches, none of which are great:

- rest of library (and haskell is camelcase) - change convention to be all camelcase. I guess with the types they'd be capital letters like xxxTDBTT. Ugly but consistent
- reserve snake case for just the signature portion so xxx_tdbtt where xxx is camelcase. Weird to mix camel and snake case together.
- snake case everything as it is now, which is inconsistent w/ rest of the api for no particular reason other than this is a part of the api where a few variants are nice to have.
- be selective about functions to include (do we need the specialized variants given you could curry defaults? is there a performance difference? not sure...)
- come up with some sort of typeclass pattern to handle polymorphism. not sure how this would work or if it's  possible to come up with something that is both pleasant to use and sufficiently general.

Honestly not super thrilled about any of these, but just getting the discussion going in case someone has better ideas.

tycho01:
I mean alternatively I could just come up with some hot distinguishing names like `rreluWithNoise'''` then we probably throw them out once we kinda work out something better :woman_shrugging:

austinvhuang:
@tycho01 yes for now use the prime convention if there aren't too many variants or camelcase convention (having snake case just for this set of functions is too different from the rest of the library). Then we can go ahead and merge this. Thanks!

tycho01:
@austinvhuang thanks, I've tried renaming them now.
The `normal` functions followed a bit of a different pattern, so I've tried to use names there sort of reflecting that. Let me know if you'd prefer primes there too somehow.

austinvhuang:
LGTM, let's revisit the choices if we start to see more example code depending on these functions.

Can you try merging the latest changes on master to your branch to see if that resolves the CI? @junjihashimoto fixed some of the issues we had with github actions lately.

tycho01:
Whoops, there is already a `normal` in `Torch.Random` -- in PyTorch they are named `Normal` (distribution class) vs. `normal` (random generator function).
Our other random generation functions all start with `rand` tho. Maybe that could make sense for that `normal` as well?
Technically a breaking change for those already referencing the `normal` function.

austinvhuang:
@tycho01 I think you're hitting on an ambiguity that needs some work. Samplers can arguably "fit" in 3 things - Functional, TensorFactories, and Random. We need to clean up guidelines on what should go where. We should have this discussion, but I'm leaning towards merging this into Random (eliminating any redundancy along the way).

We haven't formalized this, I've thought of the rand prefix as more for typical sampling/shuffling sort of uses cases rather than modeling (yes the boundary gets kind of blurry and maybe we should deprecate randn in favor of normal).

Another emerging convention is that the default versions in Random take a generator parameter to enable reproducibility. The "convenience" versions that use an implicit global RNG have an IO suffix.

tycho01:
@austinvhuang This is why I just stick to the [full function list](https://hasktorch.github.io/doc-index-All.html), cuz the module stuff makes no sense to me. :joy:

Are the monadic random sampling functions not random? Are the generator-based random sampling functions not factories of tensors? Should these sets be closer together? Would it make sense to have them closer yet separated like `Torch.Random.IO` vs `Torch.Random.Gen` or whatever? Why do some random functions have `rand` in the name yet not others?

I honestly don't know so whatever. I've now suffixed the function names with `IO` and moved them to the `TensorFactories` module together with the other IO-based random sampling functions.

austinvhuang:
You're right, the module and naming conventions need a rethink. Re typed/untyped inconsistencies that's just because we have more work to do - so if you see anything feel free to make a PR to make it more consistent.

Would like to merge this as is if we can get the CI to be green (not sure what happened with cabal-linux, but re-running now). Thanks @tycho01

tycho01:
:heavy_check_mark: :smile: @austinvhuang

PR #357, opened 2020-04-18
Added function definitions for untyped api to reduce parity

championballer:
Extension of #337 to address #328 . Includes stack, sumDim, topK, triu, tril, unSqueeze functions along with tests in untyped api.


austinvhuang:
If there are more descriptive docstrings from pytorch can we incorporate those? Otherwise looks good.

championballer:
Sure.

PR #362, opened 2020-04-19
GHC 8.10: nix

tycho01:
edit: `nix-linux` CI error:
```
Configuring language-haskell-extract-0.2.4...
...
Setup: Encountered missing or private dependencies:
template-haskell <2.16
```

hm. wth?

cc @junjihashimoto, @tscholak

### DIFF
--- a/nix/shared.nix
+++ b/nix/shared.nix
@@ -1,4 +1,4 @@
-{ compiler ? "ghc883" }:
+{ compiler ? "ghc8101" }:
 
 let
   libtorch_src = pkgs:


junjihashimoto:
For now, I think it is difficult to support nix, because some packages like language-haskell-extract and blaze-markup does not support ghc-8.10.

package-sets of nix is based on ghc-883.
https://github.com/NixOS/nixpkgs/pull/85178

To support ghc-8.10, we have to tweak package-dependency. 
Following code will be helpful.
https://github.com/hasktorch/hasktorch/blob/master/nix/shared.nix#L140-L143

tscholak:
@tycho01 as Junji said, you’ve got to wait until nixpkgs has switched to 8.10. That can take while...

tscholak:
I am confident that `haskell.nix` (see #377) will give us ghc 8.10 support on nix.

PR #363, opened 2020-04-20
Add feature-extraction AlexNet

SurajK7:
For now, I have used lists for data-loading so the implementation is slow and not memory efficient(~19 GB footprint). 
The model achieves ~80% training accuracy after 5000 iterations but validation accuracy seems rather dismal, I will try to figure what's going wrong.
Meanwhile, any suggestions or feedback would be greatly appreciated!


SurajK7:
Everything seems to work fine now. I trained the final layer for 5 epochs and achieved an accuracy of ~70% on the test set which is good enough for a fine-grained classification dataset such as [ours](http://www.robots.ox.ac.uk/~vgg/data/pets/).

junjihashimoto:
Awesome!

junjihashimoto:
I feel a little strange.
Before this pr, alexnet-examples loaded pre-existing weight including top-layer.
This pr loads pre-existing weight except for top-layer, and trains the weight of top-layer to learn new features.
So previous example's use-case is different from this one.

SurajK7:
The previous example was just to showcase the loading of pretrained parameters and could be used to run inference on images as per ImageNet categories. This example feature-extracts AlexNet for non-trivial data-sets. So as such, there was no use-case of the previous example.

SurajK7:
I wrote a PyTorch like dataset-abstraction for better loading of data but I doubt that would be of any use now. So, this will be it from my side. Thank you.

junjihashimoto:
I don't care about the dataload abstraction in this example.
So it's ok.

examples/alexNet/setup.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+wget --load-cookies /tmp/cookies.txt "https://docs.google.com/uc?export=download&confirm=$(wget --quiet --save-cookies /tmp/cookies.txt --keep-session-cookies --no-check-certificate 'https://docs.google.com/uc?export=download&id=1lH-GwVcNY0Jmf10zMRhGbbXaTe-LIH76' -O- | sed -rn 's/.*confirm=([0-9A-Za-z_]+).*/\1\n/p')&id=1lH-GwVcNY0Jmf10zMRhGbbXaTe-LIH76" -O alexNet.pt && rm -rf /tmp/cookies.txt
```
junjihashimoto, line 3: Is it some official data or yours?

junjihashimoto: If it is yours, I'm thinking of putting it here(https://github.com/hasktorch/libtorch-binary-for-ci/releases).

SurajK7: Nah it's just the extracted parameters as mentioned in the [gist](https://gist.github.com/SurajK7/90de501ae7cf332b722b06a1dabb527d).

Any pointers to what went wrong with my GSoC proposal?

junjihashimoto: There are some reasons for it, but I don't know if I can write it here.
Could you please check the feedback from @austinvhuang  in a private chat?

junjihashimoto: BTW, thank you for sharing the gist.

austinvhuang:
Thanks for this contribution @SurajK7 . Just two minor:

1) Add the gist code snippets in the directory. Makes it easier to reproduce end-to-end and gets rid of an unneeded dependency on a gist.

2) Move the google docs artifact someplace that's more intended for artifacts, could even be a github release artifact in your own repo (or we can move it to a release here if that works better) as @junjihashimoto suggested

Optional but would be nice - is there no way to avoid having to use both C++ and Python handle the serialization stuff? Would be nice if it could just be one Python script.

SurajK7:
I think it would be better if we could move the extracted parameters here.
Also, I don't think a single python script would suffice.

austinvhuang:
Thanks for adding the code @SurajK7 .

Move the serialized data artifact you mean? We're trying to keep artifacts/binaries out of the repo if at all possible.

austinvhuang:
@SurajK7 also I think if you merge master into your branch that might clear the last CI issue you have.

SurajK7:
>or we can move it to a release here if that works better

@austinvhuang, I meant we can go ahead with the quoted suggestion.

austinvhuang:
@SurajK7 - I created a release for the artifact in this repo - https://github.com/hasktorch/hasktorch-artifacts/releases/tag/alexnet 

See if it works to use this link instead https://github.com/hasktorch/hasktorch-artifacts/releases/download/alexnet/alexNet.pt

austinvhuang:
Thanks!

junjihashimoto:
@austinvhuang 
Thanks  for supporting this pr.
@SurajK7
Thanks!

PR #365, opened 2020-04-22
Added datatype for diagonal in untyped api

championballer:
Would having a datatype for diagonal help? Since more functions crept up with that as a parameter in the last PR (#357 ) i.e. tril and triu in addition to the previously present diag. Thought of this change in reference to #322 .

### DIFF
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -101,6 +101,8 @@ instance Castable Reduction Int64 where
   uncast 1 f = f ReduceMean
   uncast _ f = f ReduceSum
 
+data Diag = Diag Int
+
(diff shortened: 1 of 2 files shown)


junjihashimoto:
@championballer @austinvhuang 
How about using Dim instread of Diag?
I think both are similar.

championballer:
@junjihashimoto I think they are different in the sense that the diag parameters work around how to perform the function with reference to the main diagonal. Using dim instead might cause confusion.

junjihashimoto:
It was just a misunderstanding. It is OK. 
https://pytorch.org/docs/stable/torch.html#torch.diag

junjihashimoto:
LGTM!

PR #369, opened 2020-04-27
get-deps: install python dependency pyyaml

tycho01:
We so far called [`$PYTHON aten/src/ATen/gen.py`](https://github.com/hasktorch/hasktorch/blob/master/deps/get-deps.sh#L127), which depends on `pyyaml`.
If that wasn't preinstalled, the script failed.

### DIFF
--- a/deps/get-deps.sh
+++ b/deps/get-deps.sh
@@ -117,6 +117,12 @@ if [[ ! -d build ]]; then
 mkdir build
 fi
 
+PIP=pip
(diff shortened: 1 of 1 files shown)


austinvhuang:
LGTM, did this come up as an issue after the libtorch 1.5 migration?

tycho01:
@austinvhuang nah I experienced it before already, just hadn't bothered to fix it in here.

PR #376, opened 2020-06-08
Stop using haskell-finalizer for libtorch's objects

junjihashimoto:
libtorch's objects are deleted by haskell-finalizer, now.
http://hackage.haskell.org/package/base-4.14.0.0/docs/Foreign-Concurrent.html#v:newForeignPtr
The deletions run in a different thread than GC.

This pr changes haskell-finalizer to c-finalizer so that it runs in GC.

See a following link for the difference between haskell-finalizer and c-finalizer. http://blog.ezyang.com/2014/05/the-cost-of-weak-pointers-and-finalizers-in-ghc/


libtorch-ffi/src/Torch/Internal/Unmanaged/Type/Dimname.hs
```diff
@@ -40,11 +39,11 @@ newDimname_n _x =
 
 
 
-deleteDimname :: Ptr Dimname -> IO ()
-deleteDimname object = [C.throwBlock| void { delete $(at::Dimname* object);}|]
+foreign import ccall unsafe "hasktorch_finalizer.h &delete_dimname"
+  c_delete_dimname :: FunPtr ( Ptr Dimname -> IO ())
```
junjihashimoto, line 43: inline-c may be able to  generate these haskell & c++ codes.

PR #283, opened 2020-02-17
Add stack-with-nix-linux.yml

junjihashimoto:

PR #377, opened 2020-06-11
haskell.nix

tscholak:
Hey, this is ready for review now.
With the help of @cdepillabout, I've redone the whole nix infrastructure.
I've based it on IOHK's `haskell.nix`, https://input-output-hk.github.io/haskell.nix.


cabal.project.local~
```diff
@@ -0,0 +1 @@
+tests: True
```
cdepillabout, line 1: Maybe this wasn't intended to be committed?

codegen/codegen.cabal
```diff
@@ -51,16 +51,16 @@ executable codegen-exe
...
-  default-language: Haskell2010
-  type:            exitcode-stdio-1.0
-  hs-source-dirs:  test
-  main-is:         doctests.hs
-  ghc-options:     -Wall -threaded
-  build-depends:
-      base
-    , doctest
-    , megaparsec
+--test-suite doctests
```
cdepillabout, line 54: haskell.nix unfortunately doesn't have a good story for doctests :-\

nix/sources.json
```diff
@@ -0,0 +1,62 @@
...
+        "homepage": "",
+        "owner": "input-output-hk",
+        "repo": "hackage.nix",
+        "rev": "5c74d8440ce1925e6542ab028e27ac93b70a4f59",
+        "sha256": "1rfaf8zwciryshyfg02wz0q7phimi37bmp9rl1y6gl13k3pmj812",
+        "type": "tarball",
+        "url": "https://github.com/input-output-hk/hackage.nix/archive/5c74d8440ce1925e6542ab028e27ac93b70a4f59.tar.gz",
+        "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+    },
+    "haskell-nix": {
```
cdepillabout, line 14: I've never actually used niv, but I'm wondering if it is a mistake to have two `haskell-nix` keys here?  I would think that only one of them would actually end up getting read.

Nix attribute sets can't have repeated keys:

```nix
$ nix repl
nix-repl> { a = 3; a = 4; }
error: attribute 'a' at (string):1:10 already defined at (string):1:3
```

tscholak: Hi @cdepillabout! Thanks for reviewing :)
Where have you spotted the second `haskell-nix` key? I can’t find it

release.nix
```diff
@@ -1,7 +1,37 @@
...
-  { inherit (shared)
-      hasktorch_cpu
-    ;
-  }
+  {
+  	libtorch = hsPkgs.libtorch-ffi.components.all;
+
+  	hasktorch = hsPkgs.hasktorch.components.all;
+
+  	all-examples = hsPkgs.examples.components.all;
```
cdepillabout, line 9: Unless you have some specific reason, you almost never want to use the `all` component.  Instead, you want to pull out the components that matter to you (usually the library, exes, tests, and benchmarks).  This is because the `all` component is not just a simple `symlinkJoin`, but actually a completely separate derivation that just compiles all of the components in one derivation (similar to how nixpkgs haskell stuff works).  

haskell.nix provides some helper functions to easily pull out components.

cabal.project.local~
tscholak: oops

nix/sources.json
cdepillabout: Oh, my bad!  I misread `hackage-nix` and `haskell-nix` as both `haskell-nix`!

Please ignore this comment :-)

cdepillabout:
I think this is looking really good!  On the chat you said this was about 90% done, so I'm not sure what else you're thinking of adding, but I'd say it is pretty complete!

If I was to suggest one thing, it would be to make a slight architectural change.

Ideally, this repo would have three main nix files:

1. A file defining the upstream nixpkgs repo you want to use.  You already have this, defined in your niv files.

1. A file defining an overlay that contains all the hasktorch-related stuff.

    This file should define all the top-level derivations that hasktorch wants to export.  This file would export the things from the current `shared.nix` file and `release.nix` file.  You would ideally add some other top-level derivations from that are directly used by your CI as well.

1. A file that applies the overlay from 2. on the nixpkgs from 1., and exports the whole resulting nixpkgs.

    This file would be directly used in `default.nix` and `shell.nix`.  All you'd have to do is pull out the relevant derivations.

The overlays in 2. could be split as necessary, but they currently aren't too big, so I wasn't thinking this would be necessary yet.

I'd also suggest changing `default.nix` to instead produce a single derivation, since that is what most people are used to for a top-level `default.nix` file.  That is to say, most people expect to do a git clone, immediately run `nix-build`, and get a `result/` dir full of executables they can try to run.

-------------------------

I don't think this needs to be done in this PR, but possibly in a future PR.  It might make the nix stuff here a little easier to use.

shared.nix
```diff
@@ -0,0 +1,90 @@
...
+      hsPkgs = pkgs.haskell-nix.cabalProject {
+        src = pkgs.haskell-nix.haskellLib.cleanGit { name = "hasktorch"; src = ./.; };
+        compiler-nix-name = haskellCompiler;
+        modules = [
+          ({ config, ... }: {
+            packages.libtorch-ffi.configureFlags = [
+              "--extra-lib-dirs=${pkgs.torch}/lib"
+              "--extra-include-dirs=${pkgs.torch}/include"
+              "--extra-include-dirs=${pkgs.torch}/include/torch/csrc/api/include"
+            ];
```
cdepillabout, line 61: I'm somewhat surprised you need these lines here.

I was under the impression that as long as you have the `libtorch-ffi.cabal` file setup correctly, you won't need to pass `extra-lib-dirs` and `extra-include-dirs` here.

(Although maybe you do actually need `--extra-include-dirs=${pkgs.torch}/include/torch/csrc/api/include`.  That looks like some sort of include-within-include...?)

release.nix
```diff
@@ -1,7 +1,37 @@
...
+	static-mnist-mlp = hsPkgs.examples.components.exes.static-mnist-mlp;
+	static-xor-mlp = hsPkgs.examples.components.exes.static-xor-mlp;
+	typed-transformer = hsPkgs.examples.components.exes.typed-transformer;
+	vae = hsPkgs.examples.components.exes.vae;
+	xor-mlp = hsPkgs.examples.components.exes.xor-mlp;
+
+	all-experimental = hsPkgs.experimental.components.all;
+
+  	dataloader-cifar10 = hsPkgs.experimental.components.exes.dataloader-cifar10;
+  	untyped-nlp = hsPkgs.experimental.components.exes.untyped-nlp;
```
cdepillabout, line 36: There is a helper function in nixpkgs that should allow you to pull most of these out at once:

https://github.com/input-output-hk/haskell.nix/blob/15fc6b3774774d3b457dfe03006c0c0bdae6d7ad/lib/default.nix#L98
https://github.com/input-output-hk/haskell.nix/blob/15fc6b3774774d3b457dfe03006c0c0bdae6d7ad/lib/default.nix#L55

Basically, you pass `hsPkgs` to `selectLocalPackages`, and then call something like `getAllComponents` or `collectComponents`.

shared.nix
```diff
@@ -0,0 +1,90 @@
...
+, ...
+}:
+let
+  libtorchSrc = pkgs:
+    let src = pkgs.fetchFromGitHub {
+      owner = "stites";
+      repo = "pytorch-world";
+      rev = "6dc929a791918fff065bb40cbc3db8a62beb2a30";
+      sha256 = "140a2l1l1qnf7v2s1lblrr02mc0knsqpi06f25xj3qchpawbjd4c";
+    };
```
cdepillabout, line 16: I've never actually used niv, so I'm not completely sure about this, but maybe this should be managed with `niv` as well?

release.nix
tscholak: ah, I was hoping that something like this exists, but I couldn't find it. thanks so much for the pointer!

shared.nix
tscholak: you are absolutely right! good catch, this was just copy-pasted from the old derivations :D

shared.nix
tscholak: yeah, figuring these extra libs and includes out took a long, very long time. it's all the result of a lot of trail and error...

cdepillabout: I feel you.  We went through the exact same process for our nix stuff at work.

default.nix
```diff
@@ -1,22 +1,14 @@
+args@{ cudaVersion ? null, ...}:
```
cdepillabout, line 1: I'd suggest removing the `...` from this argument list.

Nix doesn't allow you to pass arguments from the command line that aren't directly specified in an argument list.  I've seen at least 5 or 6 people separately run into this limitation (myself included): https://github.com/NixOS/nix/issues/1360

(The only reason you'd want `...` here is if you plan on programatically importing this `default.nix` file from a separate nix derivation.)

codegen/codegen.cabal
cdepillabout: Oh, I should add that you probably want to leave this `doctests` component enabled, but explicitly disable it within haskell.nix.

You can probably do it with a module like the following:

```nix
{
  packages.codegen.components.tests.doctests.buildable = false;
}
```

(Or something like this.  I haven't actually looked at the haskell.nix docs to confirm this...)

tscholak: Hm, I'm all in favor of this, but I've got some trouble with your suggested solution. After adding this configuration item to the `modules` section of `cabalProject`, I'm getting
```
$ nix-build ./shared.nix -A defaultCpu.hsPkgs.codegen.components.tests.doctests
error: The option `packages.codegen.components.tests.doctests.buildable' has conflicting definitions, in `/nix/store/mxzz2cj5a8c32b39jjd0qkk9vj8kp5l9-source/modules/plan.nix' and `<unknown-file>'.
```

tscholak: am I changing this attribute?
https://github.com/input-output-hk/haskell.nix/blob/master/modules/package.nix#L145

shared.nix
tscholak: I moved this to `sources.json` :)

default.nix
tscholak: I think it could make sense for a third party to pass arguments to `default.nix`. If someone wants to import this repo and its nix expressions, then arguments passed here will be routed to `shared.nix` where they can affect a bunch of things. I'll think I'll keep this as is.

release.nix
tscholak: so I removed `release.nix` entirely, didn't see the point for it.

I used `collectComponents` in `default.nix`. Let me know what you think :)

cdepillabout: I think this looks good.

If I were to make a suggestion, it would be to take out one level of indirection for some of these top level attributes.

For instance, as it is now, if I want to build the `mnist-mlp` example, I have to do the following:

```console
$ nix build -L examples.examples.mnist-mlp -f ./default.nix
```

It would probably be slightly easier to only have to type `nix build -L examples.mnist-mlp -f ./default.nix`.  (Although this is only a minor nitpick)

codegen/codegen.cabal
cdepillabout: I thought about this a little more.

Ideally what we want is for `haskell.nix` to build the doctest test, but not actually run it.

It should be possible with a change like this:

```diff
diff --git a/shared.nix b/shared.nix
index 1d9f322f..7318ced0 100644
--- a/shared.nix
+++ b/shared.nix
@@ -56,6 +56,8 @@ let
             ];
             packages.libtorch-ffi.flags.cuda = USE_CUDA;
             packages.libtorch-ffi.flags.gcc = USE_GCC;
+
+            packages.codegen.components.tests.doctest.doCheck = false;
           })
         ];
       };
```

However, for whatever reason, this doesn't appear to actually work.  The `codegen` doctests appear to be run when you build the `checks`:

```console
$ nix-build shared.nix -A defaultCpu.hsPkgs.codegen.checks
trace: Using latest index state for hasktorch!
trace: Using index-state: 2020-06-11T00:00:00Z for hasktorch
trace: WARNING: license "GPL-2.0-or-later AND BSD-3-Clause" not found
these derivations will be built:
  /nix/store/2rhf2f9bd7c2rrn5i11wsdn58ffyzbvi-codegen-test-doctests-0.1.0.0-check.drv
building '/nix/store/2rhf2f9bd7c2rrn5i11wsdn58ffyzbvi-codegen-test-doctests-0.1.0.0-check.drv'...
unpacking sources
unpacking source archive /nix/store/7wf5xaqr82vv0wnyvl651cq9yb8gr562-codegen-test-doctests-0.1.0.0
source root is codegen-test-doctests-0.1.0.0
building

<no location info>: error: can't find file: src/ParseFunctionSig.hs

builder for '/nix/store/2rhf2f9bd7c2rrn5i11wsdn58ffyzbvi-codegen-test-doctests-0.1.0.0-check.drv' failed with exit code 1
error: build of '/nix/store/2rhf2f9bd7c2rrn5i11wsdn58ffyzbvi-codegen-test-doctests-0.1.0.0-check.drv' failed
```

However, I don't know why this would be happening, since the upstream `haskell.nix` seems pretty clear that if you set `doCheck` for the test components, the checks aren't actually run:

https://github.com/input-output-hk/haskell.nix/blob/643c219a5ef0a4b2570c8d056fa827214b4a0b64/builder/hspkg-builder.nix#L100-L102

You can see that checks with `doCheck` set to `false` should be getting filtered out of the `codegen.checks` attribute set, however that doesn't appear to be happening.

(Oh, and I'm not sure if this was clear, but keep in mind that when you build the `test` components, the tests don't actually get run.  You have to explicitly build the `check` attribute.  You can play around with this with a command like `nix-build shared.nix -A defaultCpu.hsPkgs.codegen.components.tests.spec` vs `nix-build shared.nix -A defaultCpu.hsPkgs.codegen.checks.spec`)
-----------------------------

So I'm not really sure what to suggest here. As long as you don't explicitly call `defaultCpu.hsPkgs.codegen.checks.spec` anywhere, I guess you won't have any problems.

.github/workflows/nix-linux.yml
```diff
@@ -26,4 +26,3 @@ jobs:
         skipNixBuild: false
         name: hasktorch
         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
-        nixBuildArgs: -A hasktorch-codegen -A libtorch-ffi_cpu  -A hasktorch_cpu -A hasktorch-examples_cpu -A hasktorch-experimental_cpu
```
cdepillabout, line 29: I'm not sure if this is on purpose, but one problem with this is that none of the tests will be run by nix, because the `default.nix` file doesn't expose any of the `checks`.

shared.nix
cdepillabout: I played around with this.  Here is what I found out.

Because the `libtorch-ffi.cabal` file specifies `torch` as an `extra-library`, the normal `extra-include-dirs` and `extra-lib-dirs` are specified by default, so you don't need to specify them explicitly here:

```console
$ git diff
diff --git a/shared.nix b/shared.nix
index 1d9f322f..834edd68 100644
--- a/shared.nix
+++ b/shared.nix
@@ -47,15 +47,17 @@ let
         src = pkgs.haskell-nix.haskellLib.cleanGit { name = "hasktorch"; src = ./.; };
         compiler-nix-name = haskellCompiler;
         modules = [
           ({ config, ... }: {
             packages.libtorch-ffi.configureFlags = [
-              "--extra-lib-dirs=${pkgs.torch}/lib"
-              "--extra-include-dirs=${pkgs.torch}/include"
               "--extra-include-dirs=${pkgs.torch}/include/torch/csrc/api/include"
             ];
             packages.libtorch-ffi.flags.cuda = USE_CUDA;
$ nix-build shared.nix -A defaultCpu.hsPkgs.libtorch-ffi.components.library
these derivations will be built:
  /nix/store/rnwjv7sc64rnsjncy7lwiwb0xjd608cy-libtorch-ffi-lib-libtorch-ffi-1.5.0.0.drv
building '/nix/store/rnwjv7sc64rnsjncy7lwiwb0xjd608cy-libtorch-ffi-lib-libtorch-ffi-1.5.0.0.drv'...
...
configuring
Configure flags:
--prefix=/nix/store/zih86zmz8ivgk0h7p90m061my1gr5y0q-libtorch-ffi-lib-libtorch-ffi-1.5.0.0 lib:libtorch-ffi --extra-include-dirs=/nix/store/yblksd7ajzpsbvill1bgmi2ba4mdz5fi-libtorch-1.5/include ... --extra-lib-dirs=/nix/store/yblksd7ajzpsbvill1bgmi2ba4mdz5fi-libtorch-1.5/lib ...
```

I guess technically you could set `packages.libtorch-ffi.components.library.configureFlags` instead of `packages.libtorch-ffi.configureFlags`, because you shouldn't need those configure flags for the tests, but it probably won't hurt anything either way.

cdepillabout: Also, as far I could tell, instead of directly setting `configureFlags`, you should be able to do something like:

```nix
packages.libtorch-ffi.components.library.includeDirs = [
  "${pkgs.torch}/include/torch/csrc/api/include"
];
```

However, this doesn't actually appear to work: https://github.com/input-output-hk/haskell.nix/issues/679

codegen/codegen.cabal
cdepillabout: Ooops, looks like I made a mistake above.  I had written `packages.codegen.components.tests.doctest.doCheck`, but that is incorrect.

Here is the correct diff:

```diff
diff --git a/shared.nix b/shared.nix
index 1d9f322f..7318ced0 100644
--- a/shared.nix
+++ b/shared.nix
@@ -56,6 +56,8 @@ let
             ];
             packages.libtorch-ffi.flags.cuda = USE_CUDA;
             packages.libtorch-ffi.flags.gcc = USE_GCC;
+
+            packages.codegen.components.tests.doctests.doCheck = false;
           })
         ];
       };
```

It should be `doctests` instead of `doctest`.

With this set correctly, the `doctest` test won't actually get run:

```console
$ nix-build shared.nix -A defaultCpu.hsPkgs.codegen.checks
/nix/store/ga351x83q4kayhqadw6rrrpp4zzbx7gk-codegen-test-spec-0.1.0.0-check
```

----------------------------------------

Also, unrelated, but to answer your question above:

> ```console
> $ nix-build ./shared.nix -A defaultCpu.hsPkgs.codegen.components.tests.doctests
> error: The option `packages.codegen.components.tests.doctests.buildable' has conflicting definitions, in `/nix/store/mxzz2cj5a8c32b39jjd0qkk9vj8kp5l9-source/modules/plan.nix' and `<unknown-file>'.
> ```

When working with nix modules, sometimes you need to use functions like `mkForce` to get things working:

```diff
diff --git a/shared.nix b/shared.nix
index 1d9f322f..c4c343b0 100644
--- a/shared.nix
+++ b/shared.nix
@@ -47,8 +47,8 @@ let
         src = pkgs.haskell-nix.haskellLib.cleanGit { name = "hasktorch"; src = ./.; };
         compiler-nix-name = haskellCompiler;
         modules = [
-          ({ config, ... }: {
-            # packages.codegen.components.tests.doctests.buildable = false;
+          ({ config, lib, ... }: {
+            packages.codegen.components.tests.doctests.buildable = lib.mkForce false;
             packages.libtorch-ffi.configureFlags = [
               "--extra-lib-dirs=${pkgs.torch}/lib"
               "--extra-include-dirs=${pkgs.torch}/include"
```

cdepillabout:
One other suggestion is to pin down the `index-state` for hackage: https://input-output-hk.github.io/haskell.nix/user-guide/materialization/

This hopefully gives slightly better reproducibility.

.github/workflows/nix-linux.yml
tscholak: oh, I've got to do something about that, thanks!

tscholak:
Hi @junjihashimoto, can I ask you to have a look at this?
I have to cover a lot of ground here, and I'm not sure I am missing regressions.

junjihashimoto:
Sure!

tscholak:
Something that I haven’t been able to figure out are the github action failures for build-shell-cache and build-ci-cache. The builds crash (?) at some point and afterwards the logs do not contain any useful information. I just see:
```
2020-06-13T23:56:06.1021277Z ##[section]Starting: Request a runner to run this job
2020-06-13T23:56:06.3612224Z Can't find any online and idle self-hosted runner in current repository that matches the required labels: 'ubuntu-latest'
2020-06-13T23:56:06.3612279Z Can't find any online and idle self-hosted runner in current repository's account/organization that matches the required labels: 'ubuntu-latest'
2020-06-13T23:56:06.3612308Z Found online and idle hosted runner in current repository's account/organization that matches the required labels: 'ubuntu-latest'
2020-06-13T23:56:06.5123335Z ##[section]Finishing: Request a runner to run this job
```
which doesn’t make any sense, because there is always plenty of log output just leading up to the crashes.

.github/workflows/nix-linux.yml
junjihashimoto: This is for testing only CPU's library not CUDA's one to reduce compilation-time, because building everything costs time and disk.

junjihashimoto: CUDA's one is here. https://github.com/hasktorch/hasktorch/blob/master/.github/workflows/nix-linux-cu10.yml#L32

.github/workflows/nix-linux-cu9.yml
```diff
@@ -6,27 +6,54 @@ on:
       - master
 
 jobs:
-  build:
-
+  tests:
+    runs-on: macOS-latest
```
junjihashimoto, line 10: We should use not macOS but ubuntu.

junjihashimoto: It may be the cause of the build-shell-cache crash.

junjihashimoto:
In haddock's build, ghc's full build runs. 
Therefore, it seems that it has taken a long time and has failed.

cabal.project.freeze.ghc810
```diff
@@ -1,266 +0,0 @@
-constraints: any.Cabal ==3.0.2.0,
```
junjihashimoto, line 1: Should we delete this?

junjihashimoto:
Everything else looks good.
@tscholak @cdepillabout 
Thank you very much.

cabal.project.freeze.ghc810
tscholak: I don't know, honestly. Did you add it?

.github/workflows/nix-linux-cu9.yml
tscholak: thanks, I fixed this, but the issues remained. I pushed my locally built dependencies to cachix now, and this brought down the build time to 11 minutes. It succeeds now it seems.

tscholak:
@cdepillabout @junjihashimoto @stites @austinvhuang @tycho01 

tests are finally passing, I think this is ready to be merged. Now would be the time to voice your concerns, if you have any.

note that, once this is merged, the old way of doing nix with hasktorch ceases to exist. merging this PR thus will invalidate https://github.com/hasktorch/hasktorch/pull/379 and https://github.com/hasktorch/hasktorch/pull/283.

some time ago, @stites wrote a disclaimer message that has been displayed after launching a nix-shell. This hasn't been reimplemented yet in the new `haskell.nix`-based code. I'd like to do that after this PR. We should also see if the documentations needs to be updated.

There's also an example of how to use the old nix overlays on a downstream code base, https://github.com/hasktorch/hasktorch/tree/master/examples/library-example. This has also not been updated, yet.

btw, thanks to @cdepillabout, there's already work on top of this lining up: https://github.com/hasktorch/hasktorch/pull/382

tycho01:
okay. I'd gladly try it out given an updated example of external consumption, but no objections to a merge here -- the status quo was no working nix support so not much to break there.

PR #382, opened 2020-06-13
Get stack.yaml working again, building with nix

cdepillabout:
This PR gets stack.yaml usable again for actually compiling with `stack`.

### DIFF
--- a/nix/stack-shell.nix
+++ b/nix/stack-shell.nix
@@ -1,0 +1,33 @@
+
+# This shell file is specifically to be used with Stack.
+#
+# This file allows using Stack's built-in Nix integration.  This means that you
+# can compile hasktorch with Stack by using a command like `stack --nix build`.
+# Stack will use Nix to download and build required system libraries (like GHC
+# and libtorch), and then build Haskell libraries like normal.
+#
+# This approach allows for more reproducibility than using Stack without Nix.
+
+{ withHoogle ? false
+}@args:
+
+let
+  shared = import ../shared.nix {};
+  pkgs = shared.defaultCpu.pkgs;
+in
+
+with pkgs;
+
+haskell.lib.buildStackProject {
+  name = "stack-build-hasktorch";
+  ghc = haskell.compiler.ghc883;
+  extraArgs = [
+    "--extra-include-dirs=${pkgs.torch}/include/torch/csrc/api/include"
+  ];
+  buildInputs = [
+    git # needed so that stack can get extra-deps from github
+    torch
+    zlib
+  ];
+  inherit withHoogle;
+}
--- a/.github/workflows/stack-nix-linux.yml
+++ b/.github/workflows/stack-nix-linux.yml
@@ -1,0 +1,39 @@
+name: stack-nix-linux
+
+on: [push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: free disk space
+        run: |
+          sudo swapoff -a
+          sudo rm -f /swapfile
+          sudo apt -y purge ghc* cabal-install* php* || true
+          sudo apt autoremove -y || true
+          sudo apt autoclean -y || true
+          docker rmi $(docker image ls -aq)
+          df -h
+      - name: Setup packages
+        run: |
+          sudo apt update -qq
+          sudo apt -y --allow-downgrades --allow-remove-essential --allow-change-held-packages install cmake curl wget unzip git libtinfo-dev python3 python3-yaml
+          (wget -qO- https://get.haskellstack.org/ | sh) || true
+      - uses: cachix/install-nix-action@v8
+      - uses: cachix/cachix-action@v6
+        with:
+          name: hasktorch
+          signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+      - run: nix-env -iA cachix -f https://cachix.org/api/v1/install
+      - run: cachix use iohk
+      - name: Build
+        run: stack --nix build
+      - name: Test
+        run: |
+          stack --nix test codegen
+          stack --nix test libtorch-ffi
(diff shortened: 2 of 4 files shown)


junjihashimoto:
This is helpful. Thx!
But I think following CI may fail, because it does not use nix.
https://github.com/hasktorch/hasktorch/blob/master/.github/workflows/stack-linux.yml

cdepillabout:
@junjihashimoto Hey, sorry, I should have marked this as a draft.  This is based on #377, but it could either get pulled into that PR, or I could wait until that is merged in and then rebase this on `master`.

However, I'm actually surprised that the `stack-linux.yml` CI isn't already failing, since the `stack.yaml` file appears to reference directories that don't exist:

https://github.com/hasktorch/hasktorch/blob/2d4ed145dc3fb9d701090a2ea863b027db07731d/stack.yaml#L11-L16

----------------------------------------------

There are three improvements that could be made to this PR:

- Like @junjihashimoto said, this PR changes the `stack.yaml` to force use of nix.  However, I imagine some users would want to not use nix.  The `stack.yaml` could be changed to not force use of nix.
- The `nix/stack-shell.nix` duplicates some of the code from `shared.nix`.  The `shared.nix` file could be refactored slightly to not require this duplication.
- The `nix/stack-shell.nix` hardcodes use of the non-cuda CPU build.  This restriction should be lifted.

The three of these things could either be fixed in this PR, or I could send a follow-up PR addressing them.

junjihashimoto:
We can merge this PR after fixing first improvement and adding new CI script to test stack-with-nix, I think.(It's up to you to send follow-up PR or fix other improvements in this PR.)

cdepillabout:
@junjihashimoto Thanks for taking a quick look at this, by the way.

> after fixing first improvement 

Do you mean that you'd like the `stack.yaml` file to not force the use of Nix?  Or are you saying that it is currently okay as it currently works?

junjihashimoto:
@cdepillabout I mean I'd like the stack.yaml file to not force the use of Nix, because forcing the use of nix breaks the existing workflow.

cdepillabout:
@junjihashimoto Okay, sounds good!

I'm about to go to bed, so I'll have to work on this more tomorrow.  (I'm surprised you're up so late!)

junjihashimoto:
I can concentrate coding when my child is sleeping.

cdepillabout:
Just for clarification, here's my understanding of the current, non-nix workflow with `stack`:

1. Figure out and install the required dependencies for `deps/get-deps.sh` (just `pip3` and `python3`?)
1. Run `deps/get-deps.sh` to actually get the dependencies.
1. Source `setenv` to set LD_LIBRARY_PATH to point to the dependencies.
1. Run `stack build`.

--------------------------------

I have changed the stack.yaml to add back in the `extra-include-dirs` and `extra-lib-dirs`.

This should make sure that the above workflow still works.  I don't think specifying `extra-include-dirs` and `extra-lib-dirs` in the stack.yaml file causes any problems when building with nix as long as those directories don't exist.

cdepillabout:
This should be ready for review (although keep it mind it still needs to be rebased onto `master` after #377 is merged in).

cdepillabout:
Here's the CI job for the new stack nix GitHub Actions Workflow added in this PR.  It is currently passing: https://github.com/cdepillabout/hasktorch/runs/769418280?check_suite_focus=true

junjihashimoto:
Thx! I have two questions.
Is stack's result cached to cachix?
What does iohk's cache provide?

cdepillabout:
>  Is stack's result cached to cachix?

Sort of.  When `stack` is building with nix support, it automatically uses `libtorch` from `pytorch-world`:

https://github.com/hasktorch/hasktorch/pull/382/files#diff-4b5ed511007286bc1cc4ea938a1964c7R46

I'm not 100% sure how the github cachix action works, but I believe that this `libtorch` will be downloaded from the hasktorch cache, and pushed to the cachix cache if it ever needs to be built from scratch.

However, the actual Haskell packages that `stack` builds will not be pushed to cachix (or stored in any cache that I can see).

I imagine that it would be possible to cache the Haskell packages build by `stack` (similar to how is often done in Travis-CI), but it doesn't appear that the current `stack-linux.yml` workflow is setup to do this, so I didn't worry about it in this `stack-nix-linux.yml` workflow either.

> What does iohk's cache provide?

I basically copied this from the `nix-linux.yml` file in #377.

I'm not exactly 100% sure what these overlays are providing here, but it is quite possible that the current build doesn't actually get anything at all from iohk's cache:

https://github.com/hasktorch/hasktorch/pull/382/files#diff-4b5ed511007286bc1cc4ea938a1964c7R31

Due to how overlays work, it is hard to say whether we use anything from `haskell.nix`, which means it is hard to say whether the IOHK cache helps us here.

However, I imagine adding the IOHK cache doesn't hurt us at all, so I went ahead and added it.

Also, at some point we may also want to change the GHC used by `stack` to be the same one we use to do the normal nix builds, in which case we definitely will need the IOHK cache:

https://github.com/hasktorch/hasktorch/pull/382/files#diff-4b5ed511007286bc1cc4ea938a1964c7R40

nix/stack-shell.nix
```diff
@@ -0,0 +1,33 @@
...
+
+{ withHoogle ? false
+}@args:
+
+let
+  shared = import ../shared.nix {};
+  pkgs = shared.defaultCpu.pkgs;
+in
+
+with pkgs;
```
cdepillabout, line 19: I was able to simplify this by just pulling out `pkgs` from the `shared.nix` file.

Again, this should eventually be updated to not always use `defaultCpu`, but take an cuda argument to figure out whether to use non-cuda cpu, or cuda.  However, I think this could be added in a future PR.

cdepillabout:
Now that #377 is merged in, I have rebased this onto `master`.

junjihashimoto:
Thx!

PR #385, opened 2020-06-14
Enable doctests to build by default again.

cdepillabout:
This PR makes sure the doctests for `hasktorch` build by default.

This also fixes up a small problem with the doctests.

This PR also makes sure that the doctests are run in CI.  Currently, `master` does not appear to be actually running the doctests.

### DIFF
--- a/hasktorch/hasktorch.cabal
+++ b/hasktorch/hasktorch.cabal
@@ -13,7 +13,8 @@ cabal-version:       >=1.10
 
 Flag disable-doctest
  Description: Disable doctest. ToDo: This flag is to avoid relocation-error of ghci for macos.
-   Default:       os(darwin)
+ Default:     False
+ Manual:      True
 
 library
  exposed-modules:     Torch
@@ -133,7 +134,7 @@ test-suite spec
                     , inline-c-cpp
 
 test-suite doctests
-  if flag(disable-doctest)
+  if os(darwin) || flag(disable-doctest)
     Buildable: False
   else
     Buildable: True
--- a/hasktorch/test/doctests.hs
+++ b/hasktorch/test/doctests.hs
@@ -4,12 +4,10 @@ import Test.DocTest
 
 main :: IO ()
 main = doctest
-  [ "-XOverloadedStrings"
-  , "-XOverloadedLists"
-  , "-XDataKinds"
+  [ "-XDataKinds"
   , "-XScopedTypeVariables"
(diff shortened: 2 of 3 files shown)


hasktorch/hasktorch.cabal
```diff
@@ -13,7 +13,8 @@ cabal-version:       >=1.10
 
 Flag disable-doctest
  Description: Disable doctest. ToDo: This flag is to avoid relocation-error of ghci for macos.
-   Default:       os(darwin)
+ Default:     False
+ Manual:      True
```
cdepillabout, line 17: As far as I could tell, `cabal` just doesn't support this `Default: os(darwin)` syntax.  It appears that by indenting `Default` past where it should be, this problem was masked.

By indenting `Default` the correct amount and using `os(darwin)`, I was seeing this when running `stack build`:

```console
$ stack --nix test hasktorch:doctests
Unable to parse cabal file from package /home/illabout/git/hasktorch2/hasktorch/hasktorch.cabal

- 16:17: 
unexpected Not a boolean: os
```

tscholak: funky!

hasktorch/test/doctests.hs
```diff
@@ -4,12 +4,10 @@ import Test.DocTest
 
 main :: IO ()
 main = doctest
-  [ "-XOverloadedStrings"
-  , "-XOverloadedLists"
```
cdepillabout, line 8: The extensions listed here will apply to _all_ Haskell files in this repo, so you have to be careful which extensions you list here.

In general, it is not safe to list extensions like `OverloadedStrings` and `OverloadedLists`, which will cause type-checking unification problems when used on .hs files that don't expect them.

The safe solution to this is to use the Doctest `$setup` blocks in the Haskell source files, as I have done in `hasktorch/src/Torch/Typed/Functional.hs` above.  This turns on the extensions for _only_ the doctests.

tscholak: thank you, that is a great suggestion!

cdepillabout:
You can see that the doctests are now being successfully run in CI:

https://github.com/hasktorch/hasktorch/pull/385/checks?check_run_id=769976905#step:6:1442

junjihashimoto:
LGTM!
@cdepillabout @tscholak 
If you want to merge this PR first, please merge.

PR #386, opened 2020-06-14
Add a typed mode operator.

cdepillabout:
This PR adds a typed `mode` operator, as per #233.

I just talked to @tscholak, and he suggested that I should check that this operator actually works on CUDA.  I don't have an nvidia GPU, so I haven't done this, but I would appreciate if someone could check this for me.

Also, because I added the doctests, I didn't add any hspec tests, but I could do this if necessary.

(Note that this PR is based on the current `master`, which does not yet contain a fix for actually running the doctests in CI, but I did check that the doctests I added here run correctly locally.)

### DIFF
--- a/hasktorch/src/Torch/Typed/Functional.hs
+++ b/hasktorch/src/Torch/Typed/Functional.hs
@@ -1406,6 +1406,38 @@ allclose
 allclose rtol atol equalNaN input other =
   unsafePerformIO $ ATen.cast5 ATen.Managed.allclose_ttddb input other rtol atol equalNaN
 
+-- | mode
+-- See https://pytorch.org/docs/stable/torch.html#torch.mode.
+--
+-- >>> t = fromJust [[0, 5], [0, 2], [3, 5]] :: CPUTensor 'D.Int64 '[3, 2]
+--
+-- >>> (modes :: CPUTensor 'D.Int64 '[2], indicies :: CPUTensor 'D.Int64 '[2]) = mode @0 @DropDim t
+-- >>> (dtype modes, shape modes, D.asValue (toDynamic modes) :: [Int])
+-- (Int64,[2],[0,5])
+-- >>> (dtype indicies, shape indicies, D.asValue (toDynamic indicies) :: [Int])
+-- (Int64,[2],[1,2])
+--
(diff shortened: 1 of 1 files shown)


cdepillabout:
Here's the pytorch docs for `mode` just in case: https://pytorch.org/docs/stable/torch.html#torch.mode

junjihashimoto:
Thx! CUDA's test is here.
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/test/Torch/Typed/FunctionalSpec.hs

cdepillabout:
@junjihashimoto Thanks for the pointer.

So you're suggesting that I add a CUDA-enabled test for `mode` to 
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/test/Torch/Typed/FunctionalSpec.hs, and then ask someone to run the test suite on a machine with an nvidia GPU?

junjihashimoto:
Yes, I have native GPU.

junjihashimoto:
Or we can do a test after merging.
It's in the cache so we can easily test it.

junjihashimoto:
Is there a good way? The test of gpu is ideally done on ci.

austinvhuang:
LGTM, more testing esp on gpu is always helpful but I also think it's relatively low risk having the same form as median / mean.

junjihashimoto:
Yeah, the risk is low. Do you want to merge? I think we can add another PR for hspec.

tscholak:
Hi all, there are some minor issues with this, but we can merge as is.
I’ve started working on the hspec tests in another branch.

cdepillabout:
@tscholak @junjihashimoto Sorry, I should have replied here earlier, but I was working on writing the hspec tests locally, but I didn't yet have anything I felt comfortable pushing.

When @tscholak opens a PR for the hspec tests, I'll try to take a look and figure out what I could have done differently in this PR.

tscholak:
Hi @cdepillabout, no problem! We are very glad you made this contribution, thanks a lot!
I’ve found some issues in the typed FunctionalSpec suite that are not all related to the mode function. I’m in the middle of sorting them out, expect a PR tomorrow :)

PR #388, opened 2020-06-16
Fix static-xor-mlp example

tscholak:
the static-xor-mlp example was broken. I fixed it.

- [x] add type alias for device, move device to cuda by default
- [x] add workaround for wrong runtime device for haskell literal: `0.5 :: Tensor '( 'D.CUDA, 0) 'D.Float '[]` does currently not lead to a cuda tensor at runtime.
- [x] resolve autodiff issue with mseLoss arguments being swapped


junjihashimoto:
LGTM!

PR #389, opened 2020-06-16
Move literal number to device's place

junjihashimoto:
When we define cuda's number and use it, it throws following exception.
This PR is to fix the exception.
``` 
0.5 :: Tensor '( 'D.CUDA, 0) 'D.Float '[]
```
```
*** Exception: CppStdException "Exception: the derivative for 'target' is not implemented; type: std::runtime_error"
```


junjihashimoto:
@tscholak Thank you for your help.

PR #390, opened 2020-06-17
Add data loading primitives

andredaprato:


hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -0,0 +1,79 @@
...
+makeFoldWithTransform transforms dataset numEpochs = createTrainLoop
+  where createTrainLoop  =  do
+          -- TODO: we can allow different buffer sizes
+          -- which would be necessary for data echoing
+            (toTransformBox, fromTransformBox, sealTransform) <- spawn' (bounded 1)
+            (toBatches, fromBatches, sealBatch) <- spawn' (bounded 1)
+            batchReader <-  async $ do runEffect $ replicateM_ numEpochs $ readBatches dataset toTransformBox 
+            transformer <-  async $ do runEffect $ runTransforms transforms fromTransformBox toBatches
+                                       atomically sealTransform
+                                       atomically sealBatch
```
junjihashimoto, line 66: batchReader and transformer are not  used. Do we need to check the returned Async?

andredaprato: This is a remnant from how I was doing things earlier. As it is now I don't
think there is a need to use the returned async values, since these threads
shouldn't terminate until the thread which is receiving data stops asking for data
(And even then, I think the threads would sit idle if the thread receiving batches asked to receive less epochs of data than are yielded here).

andredaprato: Looking at this again has helped me realize something though. In the batch
reading thread we should be able to repeat the action infinitely (rather than
numEpochs number of times), and then force the clean up of those streams
after the fold has finished. This should make it so that we couldn't have a Consumer sitting waiting for data that would never be yielded, nor a Producer waiting to yield data forever.

tscholak:
Very nice work @andredaprato! I left some comments and suggestions in Slack.
I think this approach has promise, and I'm very interested to see where we can take this!

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -0,0 +1,112 @@
...
+                                        }
+
+class Dataset dataset batch  where
+  getBatch :: dataset -> Iter -> IO batch
+  numIters :: dataset -> Int
+
+class Dataset dataset batch => ConcurrentDataset dataset batch where
+  getBatchConcurrently :: WorkerId -> dataset -> Iter -> IO batch
+
+data RunBatch = Final | KeepTrain  deriving (Eq, Show)
```
tscholak, line 51: could you make `RunBatch` isomorphic to `Maybe`?

tscholak:
```suggestion
data Run batch = Done | More batch deriving (Eq, Show)
```
perhaps?

hasktorch/test/PipelineSpec.hs
```diff
@@ -0,0 +1,61 @@
...
+  numIters (MockData iters)  = iters
+
+instance ConcurrentDataset MockData Int where
+  getBatchConcurrently _ = getBatch 
+
+testFoldTimeout :: MockData -> IO ()
+testFoldTimeout dataset = do
+  (toBatches, fromBatches, sealBatch) <- spawn' (bounded 1)
+  thread <- async $ runEffect $ readBatches dataset toBatches
+  fold <- pure $ foldFromProducer (takeBatch fromBatches)
```
tscholak, line 33:
```suggestion
  let fold = foldFromProducer (takeBatch fromBatches)
```

examples/static-xor-mlp/Main.hs
```diff
@@ -110,26 +108,41 @@ xor t = (1 - (1 - a) * (1 - b)) * (1 - (a * b))
...
+         , KnownDevice device
+         , RandDTypeIsValid device 'D.Float
+         , ComparisonDTypeIsValid device 'D.Float
+         ) => Dataset Xor (Tensor device 'D.Float '[batchSize, 2]) where
+  getBatch _ _ =  Torch.Typed.Tensor.toDType @D.Float .
+                  gt (Torch.Typed.Tensor.toDevice @device (0.5 :: CPUTensor 'D.Float '[]))
+                  <$> rand @'[batchSize, 2] @'D.Float @device
+  numIters = iters 
+
+type Device = '( 'D.CUDA, 0)
```
tscholak, line 123: do you have a GPU you can test this with?

examples/static-xor-mlp/Main.hs
```diff
@@ -110,26 +108,41 @@ xor t = (1 - (1 - a) * (1 - b)) * (1 - (a * b))
...
-      <$> rand @'[256, 2] @'D.Float @Device
+instance ( KnownNat batchSize
+         , KnownDevice device
+         , RandDTypeIsValid device 'D.Float
+         , ComparisonDTypeIsValid device 'D.Float
+         ) => Dataset Xor (Tensor device 'D.Float '[batchSize, 2]) where
+  getBatch _ _ =  Torch.Typed.Tensor.toDType @D.Float .
+                  gt (Torch.Typed.Tensor.toDevice @device (0.5 :: CPUTensor 'D.Float '[]))
+                  <$> rand @'[batchSize, 2] @'D.Float @device
+  numIters = iters
```
tscholak, line 121: :+1:

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -0,0 +1,112 @@
...
+
+type Iter = Int
+type WorkerId = Int
+
+data DatasetMock m tensor = DatasetMock { getBatchMock :: Int -> m tensor
+                                        , numItersMock :: Iter
+                                        }
+
+class Dataset dataset batch  where
+  getBatch :: dataset -> Iter -> IO batch
```
tscholak, line 45: safe indexing by means of exceptions?

tscholak, line 45:
```suggestion
class (MonadPlus m) => Dataset m dataset batch where
  getBatch :: dataset -> Iter -> m batch
```

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -0,0 +1,112 @@
...
+makeFoldWithTransform :: (MonadIO m, Dataset dataset batch)  
+  => (batch -> batch')
+  -> dataset
+  -> IO (L.FoldM m batch' b -> m b)
+makeFoldWithTransform transforms dataset = do
+          -- TODO: we can allow different buffer sizes
+          -- which would be necessary for data echoing
+            (toTransformBox, fromTransformBox, sealTransform) <- spawn' (bounded 1)
+            (toBatches, fromBatches, sealBatch) <- spawn' (bounded 1)
+            async $ runEffect $ forever $ readBatches dataset toTransformBox
```
tscholak, line 84: hm, do you need `forever` here?

hasktorch/test/PipelineSpec.hs
```diff
@@ -0,0 +1,61 @@
...
+  threads <- forM [1..numWorkers] $ \workerId -> async $ runEffect $ readBatchesConcurrently workerId dataset toBatches
+  fold <- pure $ foldFromProducer (takeBatch fromBatches)
+  timedOut <- async $ fold $  FoldM takeBatchThenTimeout (pure 0) pure
+  mapM_ wait threads
+  cancel timedOut
+
+takeBatchThenTimeout :: Int -> Int -> IO Int
+takeBatchThenTimeout _ input = threadDelay 50000 >> pure input
+
+runTest :: IO () -> IO (Maybe ())
```
tscholak, line 51: can you add a little doc string that explains that `Nothing` corresponds to a timeout, please?

hasktorch/test/PipelineSpec.hs
```diff
@@ -0,0 +1,61 @@
...
+
+
+testConcurrentFoldTimeout :: MockData  -> Int -> IO ()
+testConcurrentFoldTimeout dataset numWorkers = do
+  (toBatches, fromBatches, sealBatch) <- spawn' (bounded numWorkers)
+  threads <- forM [1..numWorkers] $ \workerId -> async $ runEffect $ readBatchesConcurrently workerId dataset toBatches
+  fold <- pure $ foldFromProducer (takeBatch fromBatches)
+  timedOut <- async $ fold $  FoldM takeBatchThenTimeout (pure 0) pure
+  mapM_ wait threads
+  cancel timedOut
```
tscholak, line 46: can you rewrite this please such that you are using `makeConcurrentFold`?

hasktorch/test/PipelineSpec.hs
```diff
@@ -0,0 +1,61 @@
...
+defaultTimeout :: Int
+defaultTimeout = 100000  
+
+newtype MockData = MockData Int
+instance Dataset MockData Int where
+  getBatch (MockData iters) _ = pure $ iters
+  numIters (MockData iters)  = iters
+
+instance ConcurrentDataset MockData Int where
+  getBatchConcurrently _ = getBatch
```
tscholak, line 27: hm, I think you should add a delay here, otherwise the data loading is instant.

hasktorch/test/PipelineSpec.hs
```diff
@@ -0,0 +1,61 @@
...
+testFoldTimeout dataset = do
+  (toBatches, fromBatches, sealBatch) <- spawn' (bounded 1)
+  thread <- async $ runEffect $ readBatches dataset toBatches
+  fold <- pure $ foldFromProducer (takeBatch fromBatches)
+  timedOut <- async $ fold $ FoldM takeBatchThenTimeout (pure 0) pure
+  wait thread
+  cancel timedOut
+
+
+testConcurrentFoldTimeout :: MockData  -> Int -> IO ()
```
tscholak, line 39: I'm trying to make sense of this. Can you maybe add some documentation that explains the expected behaviour?
I've come up with this:

working: `-`
idling:  `.`

#### with 1 worker and batching as slow as folding:

```
mailbox:  00000010000010000
worker 0: |-----|-----|----
fold:     |.....|-----|----
```

#### with 2 workers and batching twice as slow as folding:

```
mailbox:  00000021100010010
worker 0: |-----|-----|----
worker 1: |-----|..|-----|-
fold:     |.....|--|--|--|-
```

hasktorch/src/Torch/Data/Pipeline.hs
andredaprato: Yep, this makes a lot more sense.

examples/static-xor-mlp/Main.hs
andredaprato: Yep.

hasktorch/src/Torch/Data/Pipeline.hs
andredaprato: This is definitely better than IO, but we should figure out what the best constraints `m` and `dataset` should be. `MonadPlus` seems reasonable for now.

hasktorch/src/Torch/Data/Pipeline.hs
andredaprato: Yes, since `readBatches` only yields data for 1 epoch, `forever` makes it yield an indefinite number of epochs.

tscholak:
when you've got a chance, can I ask you to rebase from `master`? thanks :)

tscholak:
@andredaprato looks like the tests are failing on macOS:
```
2020-07-02T18:44:03.2766330Z Failures:
2020-07-02T18:44:03.2766380Z 
2020-07-02T18:44:03.2766460Z   test/PipelineSpec.hs:68:5:
2020-07-02T18:44:03.2766560Z   1) Pipeline Tests data is flowing
2020-07-02T18:44:03.2766650Z        expected: Just ()
2020-07-02T18:44:03.2766740Z         but got: Nothing
2020-07-02T18:44:03.2766780Z 
2020-07-02T18:44:03.2767280Z   To rerun use: --match "/Pipeline/Tests data is flowing/"
2020-07-02T18:44:03.2767370Z 
2020-07-02T18:44:03.2767470Z   test/PipelineSpec.hs:70:5:
2020-07-02T18:44:03.2767560Z   2) Pipeline Tests concurrent datasets yield concurrently
2020-07-02T18:44:03.2767650Z        expected: Just ()
2020-07-02T18:44:03.2767750Z         but got: Nothing
2020-07-02T18:44:03.2767810Z 
2020-07-02T18:44:03.2768240Z   To rerun use: --match "/Pipeline/Tests concurrent datasets yield concurrently/"
```
not sure why, will run it on my Mac and see if they pass

andredaprato:
Ok thanks, I think it might be that the timeout window is too short.

tscholak:
`timeoutConcurrent = 70000` works for me locally

andredaprato:
I find it odd that it needs that much more time on macos. Because the batch reading threads should never be idling, but it could very well just be overhead, so I think it is fine to change it.

tscholak:
@andredaprato actually, we need to increase it even further:
```hs
defaultTimeout = 30000
timeoutConcurrent = 70000
```
I ran
```
$ while cabal run hasktorch:test:spec -- --match "/Pipeline"; do :; done
```
to run the tests over and over again until they fail.
And with the timeouts above they were still somewhat flaky.

hasktorch/test/PipelineSpec.hs
```diff
@@ -0,0 +1,91 @@
...
+-- | Fold:     |.....|--|--|--|--|
+spec = do
+  it "Tests data is flowing" $
+    (runTest defaultTimeout (testFoldTimeout $ MockData 2)) `shouldReturn` (Just ())
+  it "Tests concurrent datasets yield concurrently" $
+    (runTest timeoutConcurrent (testConcurrentFoldTimeout (MockData 3) 2)) `shouldReturn` (Just ())
+  it "Tests that failures in dataset processing aren't yielded" $
+    (runTest 10000 (testFoldTimeout $ FailureDataset 4)) `shouldReturn` (Just ())
+  it "Tests batches are still yielded after failures" $
+    (runTestExpectTimeout 10000 (testFoldTimeout $ FailureDataset 5)) `shouldReturn` (Just ())
```
tscholak, line 91: the tests still fail on macOS.
I think this has been a distraction long enough now, let's wrap this in `#if !defined(darwin_HOST_OS)` and `#endif` such that it is deactivated on macOS. (See https://github.com/haskell/unix/blob/master/System/Posix/Signals.hsc#L379, for example.)

andredaprato: I added this suggestion and CI looks like it succeeds now.

tscholak: awesome @andredaprato!
Let's merge it!

PR #392, opened 2020-06-19
Add hspec for typed mode

### LINKED ISSUE #398: Readme refers to removed example
It looks like `examples/library-example` was removed in 5bd688e0bb46cc84361cc11ad361d6bf0e405573, but there is still a reference to it in the [readme](https://github.com/hasktorch/hasktorch#using-as-a-library-in-a-project-via-nix).

Should we remove the reference for now, and open an issue to add a similar example using the new `haskell.nix` infrastructure? (From my perspective as a newcomer this would be useful.)

tscholak:
- [x] add hspecs for `mode` and related reduction ops
- [x] add appropriate dtype and device constraints
- [x] clean up constraint type families
- [ ] make doctests work -- they are not working yet :/
- [x] rewrite README.md, closes #398


tscholak:
@cdepillabout Here's the follow-up on #386. I'm still having trouble with the doctests.

junjihashimoto:
I think the code is good.
But doctest throws exceptions.
image

hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -211,38 +206,165 @@ min input = unsafePerformIO $ ATen.cast1 ATen.Managed.min_t input
...
+  => Tensor device dtype shape
+  -> Tensor device dtype shape'
+mean input = unsafePerformIO $ ATen.cast3 ATen.Managed.mean_tlb
+                                        input
+                                        (natValI @dim)
+                                        (keepOrDropDimVal @keepOrDropDim)
+
+-- | Computes the median while carrying out a full reduction of all tensor dimensions.
+--
+-- >>> dtype &&& shape $ median (ones :: CPUTensor 'D.Float '[2,2])
```
junjihashimoto, line 278: This is just typo. Use not median but medianAll.

junjihashimoto: Others look good.

junjihashimoto:
LGTM!

junjihashimoto:
It was closed by mistake.

cdepillabout:
@tscholak When you're compiling with Nix, `haskell.nix` unfortunately doesn't work well with `cabal-doctests`.

Here's a PR where I am trying to add support for `cabal-doctests` to `haskell.nix`, but it was pretty hacky and I never ended up pushing it through:

https://github.com/input-output-hk/haskell.nix/pull/427

I'd suggest doing one of the following:

- Disable the doctests when building with Nix.

    This shouldn't be too bad, since you still have the doctests running on the stack-based CI things.

- Switch back to using `doctest` directly, and not using `cabal-doctests`.

   I wouldn't necessarily recommend this, since `doctest` is pretty hacky as-is, but it is sometimes possible to get the doctests running with haskell.nix like this.

tscholak:
@cdepillabout I think I will go with option 1 for now, disabling doctests when building with nix

PR #393, opened 2020-06-22
improve haskell.nix code

tscholak:
- [x] reorganize the nix code according to https://github.com/input-output-hk/cardano-skeleton
- [x] add a companion repo that shows how to use hasktorch downstream, see https://github.com/hasktorch/hasktorch-skeleton
- [ ] add back jupyter lab support. This is currently unsolved, see https://github.com/input-output-hk/haskell.nix/issues/724#issuecomment-649626448


nix/default.nix
```diff
@@ -0,0 +1,92 @@
...
+, cudaMajorVersion ? null
+}:
+
+# assert that the correct cuda versions are used
+assert cudaSupport -> (cudaMajorVersion == "9" || cudaMajorVersion == "10");
+
+let
+  sources = import ./sources.nix { inherit pkgs; }
+    // sourcesOverride;
+  iohKNix = import sources.iohk-nix {};
```
tycho01, line 18: that's some funny capitalization :sweat_smile:

shell.nix
```diff
@@ -1,25 +1,66 @@
...
+
+  devops = stdenv.mkDerivation {
+    name = "devops-shell";
+    buildInputs = [
+      niv
+    ];
+    shellHook = ''
+      echo "DevOps Tools" \
+      | ${figlet}/bin/figlet -f banner -c \
+      | ${lolcat}/bin/lolcat
```
tycho01, line 54: wait is this example code :joy:

tscholak: this is taken from https://github.com/input-output-hk/cardano-skeleton/blob/master/shell.nix

here's a demo of that:
```
$ nix-shell -A devops ./shell.nix
trace: Using latest index state for hasktorch!
trace: Using index-state: 2020-06-11T00:00:00Z for hasktorch
trace: WARNING: license "GPL-2.0-or-later AND BSD-3-Clause" not found
                 ######                #######               
                  #     # ###### #    # #     # #####   ####  
                  #     # #      #    # #     # #    # #      
                  #     # #####  #    # #     # #    #  ####  
                  #     # #      #    # #     # #####       # 
                  #     # #       #  #  #     # #      #    # 
                  ######  ######   ##   ####### #       ####  
                                                              
                      #######                             
                         #     ####   ####  #       ####  
                         #    #    # #    # #      #      
                         #    #    # #    # #       ####  
                         #    #    # #    # #           # 
                         #    #    # #    # #      #    # 
                         #     ####   ####  ######  ####  
                                                          
NOTE: you may need to export GITHUB_TOKEN if you hit rate limits with niv
Commands:
  * niv update <package> - update package
```

.github/workflows/nix-linux.yml
```diff
@@ -20,30 +20,10 @@ jobs:
       - run: nix-env -iA cachix -f https://cachix.org/api/v1/install
```
tycho01, line 20: btw! if you're DRYing these up, I think repetition of like the disk cleanup blocks could be factored out using like [yaml anchors](https://camel.readthedocs.io/en/latest/yamlref.html).

.github/workflows/nix-linux-cu9.yml
```diff
@@ -36,24 +36,4 @@ jobs:
...
-      - uses: actions/checkout@v2
-      - name: free disk space
-        run: |
-          sudo swapoff -a
-          sudo rm -f /swapfile
-          sudo apt -y purge ghc* cabal-install* php* || true
-          sudo apt autoremove -y || true
-          sudo apt autoclean -y || true
-          docker rmi $(docker image ls -aq)
-          df -h
```
junjihashimoto, line 52: It is necessary when there are not enough disks.
Do you want to erase it?

.github/workflows/nix-jupyterlab.yml
```diff
@@ -17,17 +17,14 @@ jobs:
...
-      run: |
-        git submodule init && git submodule update
-    - uses: cachix/cachix-action@releases/v5
+    - uses: cachix/install-nix-action@v8
+    - uses: cachix/cachix-action@v6
       with:
         name: hasktorch
         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
         skipNixBuild: false
         file: experimental/jupyterlab/default.nix
```
junjihashimoto, line 26: It doesn't need to be changed, but it might be better to do the same as hasktorch-skeleton.

junjihashimoto:
Awesome!

.github/workflows/nix-jupyterlab.yml
tscholak: Hi Junji, I think this action needs to be redone soon anyway because the Jupyter Lab derivation will change. At the moment it looks like the workflow will be similar to running any other nix shell:
```sh
$ nix-shell ./nix/jupyter-shell.nix
```

.github/workflows/nix-linux.yml
tscholak: another time

.github/workflows/nix-linux-cu9.yml
tscholak: I think this was done by accident, I put it back :)

PR #394, opened 2020-06-25
Update references to GHC version to 8.8.3

mcwitt:
The project seems to have moved to GHC 8.8.3 (via lts-15.5), but there are still several references to 8.6.5

- [x] README docs on setting up a dev environment in VS Code. Should we just remove references to version numbers here, and instead include a comment to the effect that the hie version should match the project's ghc version?
- [x] ~`tested-with` [attribute](https://github.com/mcwitt/hasktorch/blob/351f41759f2d6cdf71b0e84720477897258a6aa1/examples/library-example/library-example.cabal#L15). Does it make sense to change this, or is there something upstream that needs updating?~ (this is being reworked as part of haskell.nix transition)
- [ ] ~[Dockerfile](https://github.com/mcwitt/hasktorch/blob/b47488810e0a29f2f5648e12468a00453c95cf4b/Dockerfile#L10). I left this unchanged for the time being (is the image still on 8.6.5 for a reason?)~ for future


junjihashimoto:
Thx!

tscholak:
Thank you!
The section about HIE needs to be rewritten slightly. The preferred language server backend Is ghcide at the moment. The nix shell provides it.

mcwitt:
Sounds good, I updated the readme instructions to reflect that ghcide is preferred and included in the shell env:
- Remove instructions for HIE, since ghcide is now preferred
- Remove editor-specific parts; link to ghcide readme instead

One other question, is it worth trying to bump `ghc-8.6.5` in the Dockerfile for this PR? It wasn't clear to me whether this is still being maintained.

tscholak:
Cool, thank you @mcwitt!
The `Dockerfile` needs some love.
It may be better to remove it and add a Nix derivation that builds and publishes a docker image that can be used with `nvidia-docker`.
It's probably best to leave it alone for now.

I'm much in favour of merging your changes to the README, so that I can fold them into my current rewrite, what do you think?

examples/library-example/library-example.cabal
```diff
@@ -12,7 +12,7 @@ category:            Games
 build-type:          Simple
 extra-doc-files:     README.md
                    , CHANGELOG.md
-tested-with:         GHC == 8.6.5
+tested-with:         GHC == 8.8.3
```
tscholak, line 15: this file (and directory) and directory has been deleted. the library example lives now in another repo so that it can be easily cloned, see https://github.com/hasktorch/hasktorch-skeleton.

mcwitt:
@tscholak sounds good, the readme changes should be ready to merge!

PR #400, opened 2020-06-27
dedupe by yml anchors, what about across files tho?

tycho01:


tscholak:
> what about across files tho?

you could rewrite all of that in `dhall` and convert it to yaml at the end ;)

tycho01:
@tscholak ohh cool! I haven't actually used Dhall yet I think. :sweat_smile:

tscholak:
I'm not entirely serious here.
There's a lot of repetition between the nix action scripts, but I think adding `dhall` here would be a bit extreme.

PR #405, opened 2020-06-28
add ormolu and brittany to the shell tools

tscholak:
the haskell language server plugin can use either to format hs code


junjihashimoto:
LGTM!

PR #406, opened 2020-06-29
Refactor distillation, add regularized training, Ceil/Bool ADT for maxpool

austinvhuang:
- [x] Clean up interfaces and modularize distillation example. 
- [x] Add regularized training module to distillation
- [x] Some rudimentary plotting using hvega
- [x] Add Ceil/Floor ADT to maxpool functions (previously was Bool)
- [x] Add `linearForward` as forward function on Linear ADT (eventually will disambiguate the Functional linear from the ADT linearForward, but will need to update other examples, for now leave both linearForward and linear functions on ADT as aliases).

To-do in subsequent PR:
- [ ] Apply pruning transformation to networks trained with regularization.


junjihashimoto:
Awesome!

examples/distill/Regularization.hs
```diff
@@ -0,0 +1,104 @@
...
+    toHtmlFile "plotInit.html" plt
+    system "open plotInit.html"
+    print "weights0 l1"
+    plt <- strip (toDependent . weight . cnnFC0 $ l1Model)
+    toHtmlFile "plot0l1.html" plt
+    system "open plot0l1.html"
+    print "weights0 l2"
+    plt <- strip (toDependent . weight . cnnFC0 $ l2Model)
+    toHtmlFile "plot0l2.html" plt
+    system "open plot0l2.html"
```
junjihashimoto, line 82: It is usefull, but `open` is a macos's command. linux uses xdg-open or gnome-open.

hasktorch/src/Torch/Functional.hs
```diff
@@ -838,26 +868,35 @@ maxPool2d kernelSize stride padding dilation ceilMode self =
...
-    :: (Int,Int,Int) -- ^ kernel size
-    -> (Int,Int,Int) -- ^ stride
-    -> (Int,Int,Int) -- ^ padding
-    -> (Int,Int,Int) -- ^ dilation
-    -> Bool -- ^ ceil mode
-    -> Tensor -- ^ input
-    -> Tensor -- ^ output
-maxPool3d kernelSize stride padding dilation ceilMode self =
-    unsafePerformIO $ (cast6 ATen.max_pool3d_tllllb)
-        self kernelSize stride padding dilation ceilMode
```
junjihashimoto, line 853: Should we remove maxPool3d?

examples/distill/Regularization.hs
junjihashimoto: I think someone want to change this output destination (in my case, it is png), so can it be replaced with an environment variable?

austinvhuang: Good call. The syscall to open was just a personal convenience for me as I was debugging, I think we should just remove it for simplicity rather than wrangling a cross-platform file open. 

Might also be good to put these files in some specific place or require a command line arg so as not to clutter the working directory of the user without them being aware of the side effect.

hasktorch/src/Torch/Functional.hs
austinvhuang: I think this was accidental as I was moving the maxpool ops to another spot in the file. Thanks for the catch.

austinvhuang:
Merging though outstanding CI resolution is pending w/ https://github.com/hasktorch/hasktorch/pull/427

PR #407, opened 2020-06-29
Add slice-operators

junjihashimoto:
This PR adds slice-operators.
It is made in the same way as C++ syntax.

Pytorch(Python) | Pytorch(C++)| This PR
-- | -- | -- 
None | None | None
Ellipsis | Ellipsis | Ellipsis
... | "..." | Ellipsis
123 | 123 | 123
True | true | True
False | false | False
: or :: | Slice() or Slice(None, None) or Slice(None, None, None) | ()
1: or 1:: | Slice(1, None) or Slice(1, None, None) | Slice 1
:3 or :3: | Slice(None, 3) or Slice(None, 3, None) | Slice (None,3)
::2 | Slice(None, None, 2) | Slice (None,None,2)
1:3 | Slice(1, 3) | Slice (1,3)
1::2 | Slice(1, None, 2) | Slice (1,None,2)
:3:2 | Slice(None, 3, 2) | Slice (None,3,2)
1:3:2 | Slice(1, 3, 2) | Slice (1,3,2)
torch.tensor([1, 2]) | torch::tensor({1, 2}) | asTensor([1,2])

### DIFF
--- a/hasktorch/src/Torch/Tensor.hs
+++ b/hasktorch/src/Torch/Tensor.hs
@@ -273,6 +273,16 @@ newtype RawTensorIndex = RawTensorIndex (ForeignPtr ATen.TensorIndex)
     ATen.tensorIndexList_push_back vec i
   ATen.index t vec >>= (return . Unsafe)
 
+(@=) :: TensorIndex a => Tensor -> (a,Tensor) -> Tensor
+(Unsafe t') @= (idx,(Unsafe v)) = unsafePerformIO $ do
+  let idxs = pushIndex [] idx
+  t <- ATen.clone_t t'
+  vec <- ATen.newTensorIndexList
+  forM_ idxs $ \(RawTensorIndex i) -> do
+    ATen.tensorIndexList_push_back vec i
+  ATen.index_put_ t vec v
+  return $ Unsafe t
+
 data None = None
   deriving (Show, Eq)
 
--- a/hasktorch/src/Torch/Tensor.hs
+++ b/hasktorch/src/Torch/Tensor.hs
@@ -188,6 +188,16 @@ indexSelect
  -> Tensor
 indexSelect t dim indexTensor = unsafePerformIO $ (cast3 ATen.index_select_tlt) t dim indexTensor
 
+-- | Slices the input tensor along the selected dimension at the given range. 
+slice
+  :: Tensor -- ^ input
+  -> Int -- ^ dim
+  -> Int -- ^ start
+  -> Int -- ^ end
+  -> Int -- ^ step
+  -> Tensor
+slice _self _dim _start _end _step = unsafePerformIO $ (cast5 ATen.slice_tllll) _self _dim _start _end _step
+
 -- | Returns a tensor with the same data and number of elements as input, but with the specified shape.
 reshape 
  :: [Int] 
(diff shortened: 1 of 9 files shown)


austinvhuang:
LGTM, thanks!

Looks like some stackage dependency is failing?

image

hasktorch/src/Torch/Tensor.hs
```diff
@@ -218,20 +238,112 @@ toCUDA t = unsafePerformIO $ (cast1 ATen.tensor_cuda) t
...
 
+
+-- ToDo:
+--   1, offset may have a bug.
+--   2, Implement Ellipsis and Boolean
+
+instance TensorIndex (Slice (Integer,Integer)) where
+  indexWith (t, offset) (Slice (start,end)) = (slice t offset (fromIntegral start) (fromIntegral end) 1, offset+1)
+
+instance TensorIndex (Slice (Int,Int)) where
```
tscholak, line 288: I think you can use the appropriate integral type class to save on repetition

hasktorch/src/Torch/Tensor.hs
```diff
@@ -320,7 +432,8 @@ instance {-# OVERLAPPING #-}TensorLike a => TensorLike [a] where
 
   asTensor v = asTensor' v defaultOpts
 
-  asValue t = unsafePerformIO $ do
+  asValue t' = unsafePerformIO $ do
+    let t = if isContiguous t' then t' else contiguous t'
```
tscholak, line 436: do we have a good understanding of which ops support non-contiguous tensors?

hasktorch/test/TensorSpec.hs
```diff
@@ -16,138 +16,159 @@ import Data.Int
...
+    it "lists having different length" $ do
+      (print (asTensor ([[1],[1,2]] :: [[Double]]))) `shouldThrow` anyException
+    it "cast of Tensor" $ do
+      let x = asTensor ([0..10] :: [Int])
+      (dtype (toType Float x)) `shouldBe` Float
 
+  describe "indexing" $ do
+    it "pick up a value" $ do
+      let x = asTensor ([[[0,1,2],[3,4,5]],[[6,7,8],[9,10,11]]] :: [[[Int]]])
+      show (x @@ (1::Int,0::Int,2::Int)) `shouldBe` "Tensor Int64 []  8"
```
tscholak, line 158: I think it would be better to compare tensor values and properties directly rather than relying on the Show instance here

tscholak:
> LGTM, thanks!
> 
> 
> 
> Looks like some stackage dependency is failing?
> 
> 
> 
> image
> 
> 

I’m preparing a PR right now that bumps these dependencies

junjihashimoto:
Thank you for your reviewing!

hasktorch/src/Torch/Tensor.hs
junjihashimoto: I've tried integral type class. 
```
instance {-# OVERLAPS #-}(Integral a) => TensorIndex (Slice (a,a)) where
  indexWith (t, offset) (Slice (start,end)) = (slice t offset (fromIntegral start) (fromIntegral end) 1, offset+1)
```
But I've met following errors.

image

hasktorch/src/Torch/Tensor.hs
junjihashimoto: https://discuss.pytorch.org/t/what-ops-make-a-tensor-non-contiguous/3676/6

hasktorch/test/TensorSpec.hs
junjihashimoto: Yeah, I've replaced it.

hasktorch/src/Torch/Tensor.hs
tscholak: interesting, is it true though that every op in `Torch.Functional` can handle non-contiguous tensors as input?
I'm trying to figure out if we have to track "continuousness" at the type-level in `Torch.Typed`...

hasktorch/src/Torch/Tensor.hs
```diff
@@ -218,29 +239,157 @@ toCUDA t = unsafePerformIO $ (cast1 ATen.tensor_cuda) t
...
+
+instance {-# OVERLAPS #-} (Integral a) => TensorIndex (Slice (None,a)) where
+  pushIndex vec (Slice (_,end)) = unsafePerformIO $ do
+    idx <- ATen.newTensorIndexWithSlice 0 (fromIntegral end :: CInt) 1
+    return ((RawTensorIndex idx):vec)
+
+instance {-# OVERLAPS #-} TensorIndex (Slice ()) where
+  pushIndex vec (Slice ()) = unsafePerformIO $ do
+    idx <- ATen.newTensorIndexWithSlice 0 (maxBound :: CInt) 1
+    return ((RawTensorIndex idx):vec)
```
tscholak, line 360: this is much better, awesome!

PR #410, opened 2020-06-30
import Torch.Typed

tscholak:
contains #408 

simplifies the imports for the typed API.
In many cases, only `import Torch.Typed` is necessary.


junjihashimoto:
LGTM!

PR #411, opened 2020-06-30
Add -j 2 to reduce used memory

junjihashimoto:


junjihashimoto:
-j2 option may not be related to build failure.

.github/workflows/nix-linux-cu10.yml
```diff
@@ -21,11 +21,12 @@ jobs:
           df -h
       - uses: cachix/install-nix-action@v10
       - run: nix-env -iA cachix -f https://cachix.org/api/v1/install
+      - run: cachix use hasktorch
```
junjihashimoto, line 24: This setting may speed up a test.

.github/workflows/nix-linux.yml
```diff
@@ -42,5 +43,9 @@ jobs:
...
           name: hasktorch
           signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
-      - run: cachix use iohk
-      - run: nix-build
+      - run: |
+          cachix use iohk
+          nix-build -j 2
+          nix-build -j 2 -A haskellPackages.libtorch-ffi.checks.spec
+          nix-build -j 2 -A haskellPackages.hasktorch.checks.spec
+          nix-build -j 2 -A haskellPackages.examples.checks.spec
```
junjihashimoto, line 51: Check success or failure of CI by reducing the number of jobs.

.github/workflows/nix-jupyterlab.yml
```diff
@@ -21,6 +21,6 @@ jobs:
       with:
         name: hasktorch
         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
-        skipNixBuild: false
-        file: nix/jupyter-shell.nix
-        nixBuildArgs: --option sandbox false 
+    - run: |
+        cachix use iohk
+        nix-build nix/jupyter-shell.nix -j 2 --option sandbox false
```
junjihashimoto, line 26: cachix-action@v6 does not support nixBuildArgs. 
cachix-action@v6  uploads caches in post action.

tscholak:
I've been restarting jobs whenever I passed by a couple of times today. It still crashes:
```

**Comment by junjihashimoto** [comment]
This is weird, because job is one process and github-actions's timeout is 6 hours.
I think some process is using memory unexpectedly.

**Comment by junjihashimoto** [comment]
In addition to this, Torch.Typed.AuxSpec is lightweight test not using memory so much, I think.

**Comment by junjihashimoto** [comment]
Before running nix-build, free space of memory is 6GB. It  is enough.
image

File: .github/workflows/nix-jupyterlab.yml [hunk 4]
```diff
@@ -23,4 +23,4 @@ jobs:
         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
     - run: |
         cachix use iohk
-        nix-build nix/jupyter-shell.nix -j 1 --option sandbox false 
+        nix-shell nix/jupyter-shell.nix -j 1 --option sandbox false --command echo
```
tscholak, line 26: that should be left at `nix-build`. the intention here is to build the shell derivation and cache it.

.github/workflows/nix-linux.yml
tscholak: looks like it's still failing

tscholak:
ok, let's go with this for now, I don't want to have others suffer because we haven't figured this out

junjihashimoto:
Thank you for your reviewing!

PR #409, opened 2020-06-29
implement typed narrow

o1lo01ol1o:

PR #413, opened 2020-06-30
Add typed diag

mcwitt:
#233 

- [x] diag

**EDIT**: this PR just adds `diag` to `Torch.Typed.Functional`. The comments below the line are resolved.

TODO
- [x] Fix invalid dtype, device combos exposed by tests

----
(outdated)

So far I've only attempted `diag`. I approached this by splitting into 2 branches (`upperDiag` and `lowerDiag`), which return diagonals above and below the main diagonal, respectively (this is so the index can be a natural).

I also considered implementing this as a single function with a `Bool` type parameter, i.e. something like

``` haskell
diag
  :: forall (index :: Nat) (upper :: Bool) (i :: Nat) (j :: Nat) device dtype
   . (KnownNat index, KnownBool upper)
  => Tensor device dtype '[i, j] -- ^ input
  -> Tensor device dtype '[Min i j - index] -- ^ output
diag t = unsafePerformIO $ ATen.cast2 ATen.Managed.tensor_diag_l t index
  where
    absIndex = natValI @index
    index = if boolVal (Proxy @upper) then absIndex else -absIndex
```

but splitting into 2 branches seemed simpler, although I'm happy to revisit this.

### DIFF
--- a/hasktorch/src/Torch/Typed/Functional.hs
+++ b/hasktorch/src/Torch/Typed/Functional.hs
@@ -1191,8 +1191,31 @@ transpose2D
   -> Tensor device dtype '[j, i] -- ^ output
 transpose2D = transpose @0 @1
 
--- diag :: Tensor device dtype shape -> Int -> Tensor device dtype shape
--- diag t index = unsafePerformIO $ (ATen.cast2 ATen.Managed.tensor_diag_l) t index
+-- | diag with positive index (above the main diagonal)
+--
+-- >>> dtype &&& shape $ upperDiag @0 (ones :: CPUTensor 'D.Float '[3,2])
+-- (Float,[2])
+-- >>> dtype &&& shape $ upperDiag @1 (ones :: CPUTensor 'D.Float '[3,2])
+-- (Float,[1])
+upperDiag
+  :: forall (index :: Nat) (i :: Nat) (j :: Nat) device dtype
+   . KnownNat index
+  => Tensor device dtype '[i, j] -- ^ input
+  -> Tensor device dtype '[Min i j - index] -- ^ output
+upperDiag t = unsafePerformIO $ ATen.cast2 ATen.Managed.tensor_diag_l t (natValI @index)
+
+-- | diag with negative index (below the main diagonal)
+--
+-- >>> dtype &&& shape $ lowerDiag @0 (ones :: CPUTensor 'D.Float '[3,2])
+-- (Float,[2])
+-- >>> dtype &&& shape $ lowerDiag @1 (ones :: CPUTensor 'D.Float '[3,2])
+-- (Float,[2])
+lowerDiag
+  :: forall (index :: Nat) (i :: Nat) (j :: Nat) device dtype
+   . KnownNat index
+  => Tensor device dtype '[i, j] -- ^ input
+  -> Tensor device dtype '[Min i j - index] -- ^ output
+lowerDiag t = unsafePerformIO $ ATen.cast2 ATen.Managed.tensor_diag_l t (- natValI @index)
 
 -- | all
 -- See https://pytorch.org/docs/stable/tensors.html#torch.BoolTensor.all.
--- a/hasktorch/src/Torch/Typed/Functional.hs
+++ b/hasktorch/src/Torch/Typed/Functional.hs
@@ -1194,7 +1194,7 @@ transpose2D = transpose @0 @1
 type family DiagShape (index :: Nat) (shape :: [Nat]) :: [Nat] where
   DiagShape i '[n] = '[n + i, n + i] -- 1d -> 2d case
   DiagShape i '[m, n] = '[Min m n - i] -- 2d -> 1d case
-  DiagShape _ _ =  TypeError (Text "The input must be matrix or vector.")
+  DiagShape _ _ =  TypeError (Text "The input must be a matrix or a vector.")
 
 -- | diag
 --
--- a/hasktorch/src/Torch/Typed/Functional.hs
+++ b/hasktorch/src/Torch/Typed/Functional.hs
@@ -1191,31 +1191,24 @@ transpose2D
   -> Tensor device dtype '[j, i] -- ^ output
 transpose2D = transpose @0 @1
 
--- | diag with positive index (above the main diagonal)
+-- | diag
 --
--- >>> dtype &&& shape $ upperDiag @0 (ones :: CPUTensor 'D.Float '[3,2])
+-- >>> dtype &&& shape $ diag @0 Upper (ones :: CPUTensor 'D.Float '[3,2])
 -- (Float,[2])
--- >>> dtype &&& shape $ upperDiag @1 (ones :: CPUTensor 'D.Float '[3,2])
+-- >>> dtype &&& shape $ diag @1 Upper (ones :: CPUTensor 'D.Float '[3,2])
 -- (Float,[1])
-upperDiag
+-- >>> dtype &&& shape $ diag @1 Lower (ones :: CPUTensor 'D.Float '[3,2])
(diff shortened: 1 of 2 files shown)


tscholak:
Hi @mcwitt, thanks for having a stab at this! Well done!

I only had a brief look, but it seems you could indeed merge these two functions. It seems that since the output type does not depend on weather its lower or upper half, you don't have to bring this boolean up to the type level. You can instead use an input value instead. I recommend using `Torch.Functional.Tri(Upper, Lower)`.

tscholak:
Another thought: You have implemented the case 2d input -> 1d output.
`torch.diag` also supports 1d input -> 2d output, see https://pytorch.org/docs/master/generated/torch.diag.html.

hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -1191,6 +1191,10 @@ transpose2D
   -> Tensor device dtype '[j, i] -- ^ output
 transpose2D = transpose @0 @1
 
+type family DiagShape (index :: Nat) (shape :: [Nat]) :: [Nat] where
+  DiagShape i '[n] = '[n + i, n + i] -- 1d -> 2d case
+  DiagShape i '[m, n] = '[Min m n - i] -- 2d -> 1d case
```
tscholak, line 1196: great!

hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -1200,11 +1204,13 @@ transpose2D = transpose @0 @1
...
+  :: forall (index :: Nat) (shape :: [Nat]) (shape' :: [Nat]) device dtype
+   . ( KnownNat index
+     , shape' ~ DiagShape index shape
+     )
   => Tri -- ^ upper or lower
-  -> Tensor device dtype '[i, j] -- ^ input
-  -> Tensor device dtype '[Min i j - index] -- ^ output
+  -> Tensor device dtype shape -- ^ input
+  -> Tensor device dtype shape' -- ^ output
 diag tri t = unsafePerformIO $ ATen.cast2 ATen.Managed.tensor_diag_l t
```
tscholak, line 1214: you may want to look into writing a test suite for this function in `Torch.Typed.FunctionalSpec`, see, e.g., https://github.com/hasktorch/hasktorch/pull/409/files#diff-e428518635181a3dbe949a37402aa368

tscholak: for instance, the type signature claims that this function works for all `device`s and all `dtype`s. you may want to make sure that this is accurate :)

mcwitt: on it, thanks!

hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -1191,8 +1191,31 @@ transpose2D
   -> Tensor device dtype '[j, i] -- ^ output
 transpose2D = transpose @0 @1
 
--- diag :: Tensor device dtype shape -> Int -> Tensor device dtype shape
--- diag t index = unsafePerformIO $ (ATen.cast2 ATen.Managed.tensor_diag_l) t index
+type family DiagShape (index :: Nat) (shape :: [Nat]) :: [Nat] where
+  DiagShape i '[n] = '[n + i, n + i] -- 1d -> 2d case
+  DiagShape i '[m, n] = '[Min m n - i] -- 2d -> 1d case
+  DiagShape _ _ =  TypeError (Text "The input must be a matrix or a vector.")
```
tscholak, line 1197: that's a great addition and will help the user, thanks!
```suggestion
  DiagShape _ shape =  TypeError (Text "The input must be a matrix or a vector, but it has " :<>: ListLength shape  :<>: " dimensions.")
```
if you can, you may also want to guard against invalid `i`s :)

hasktorch/test/Torch/Typed/FunctionalSpec.hs
```diff
@@ -698,6 +698,18 @@ instance
...
+
+instance
+  ( TensorOptions shape dtype device
+  , TensorOptions shape' dtype device
+  , shape' ~ DiagShape index shape
+  , KnownNat index
+  ) => Apply' DiagSpec ((Proxy index, (Proxy device, (Proxy dtype, Proxy shape))), IO ()) (IO ()) where
+  apply' DiagSpec (_, agg) = agg >> do
+    let t = ones @shape @dtype @device
+    foldMap (\tri -> checkDynamicTensorAttributes $ diag @index tri t) [Upper, Lower]
```
tscholak, line 711: very good!

hasktorch/test/Torch/Typed/FunctionalSpec.hs
```diff
@@ -981,6 +993,21 @@ spec' device =
           hfoldrM @IO Transpose2DSpec () (hattach cpu   (hproduct allDTypes (Proxy @'[2, 3] :. HNil)))
         Device { deviceType = CUDA, deviceIndex = 0 } ->
           hfoldrM @IO Transpose2DSpec () (hattach cuda0 (hproduct allDTypes (Proxy @'[2, 3] :. HNil)))
+      it "diag" $ do
+        let
+          shapes1 = Proxy @'[0] :. Proxy @'[1] :. Proxy @'[3] :. HNil
+          shapes2 = Proxy @'[2, 3] :. HNil
+          shapes2' = Proxy @'[0, 0] :. Proxy @'[0, 1]  :. Proxy @'[1, 0] :. HNil -- only offset 0 has upper and lower defined
```
tscholak, line 1000: if certain offsets are not allowed with certain shapes, then you may want to add this to your type family above, e.g. 
```hs
type family DiagShape (index :: Nat) (shape :: [Nat]) :: [Nat] where
   ...
   DiagShape 0 '[0, n] = ... ok!
   DiagShape _ '[0, n] = TypeError "boom!"
   DiagShape 0 '[n, 0] = ... ok!
   DiagShape _ '[n, 0] = TypeError "boom!"
   ...
```

tscholak: or maybe there's a more general rule that holds for all `'[m, n]` :)

mcwitt: Thanks for the tip. I think this should be addressed generally by 4523321.

hasktorch/test/Torch/Typed/FunctionalSpec.hs
```diff
@@ -1017,6 +1031,22 @@ spec' device =
...
+          tris = Proxy @'Upper :. Proxy @'Lower :. HNil
+          indexes = Proxy @0 :. Proxy @1 :. HNil
+          indexes' = Proxy @0 :. HNil
+        case device of
+          Device { deviceType = CPU,  deviceIndex = 0 } -> do
+            hfoldrM @IO DiagSpec () (hproduct (hproduct tris indexes)  (hattach cpu   (hproduct standardDTypes standardShapes)))
+            hfoldrM @IO DiagSpec () (hproduct (hproduct tris indexes)  (hattach cpu   (hproduct standardDTypes vectorShapes)))
+            hfoldrM @IO DiagSpec () (hproduct (hproduct tris indexes') (hattach cpu   (hproduct standardDTypes emptyShapes)))
+          Device { deviceType = CUDA, deviceIndex = 0 } -> do
+            hfoldrM @IO DiagSpec () (hproduct (hproduct tris indexes)  (hattach cuda0 (hproduct standardDTypes standardShapes)))
```
tscholak, line 1047: you may be able to add `Half` precision here

tscholak: thanks!

PR #417, opened 2020-07-03
Allow full domain of offset values in diag* functions

### LINKED ISSUE #416: Can only express subset of allowed offsets in `diagEmbed`, `diagflat`, `diagonal`
In `Torch.Functional`, [diagEmbed][], [diagflat][], and [diagonal][] each take a value of type [Offset][] to specify the desired diagonal; `Offset` has 3 values (`OffsetMain`, `OffsetAbove`, `OffsetBelow`).

On the other hand, the PyTorch interface (e.g. [diag_embed][]) allows an integer offset (i.e. diagonals other than -1, 0, 1).

In the same module, [diag][] is using the `Diag` wrapper around `Int`s. Maybe we should transition the others to this convention?

[diagEmbed]: https://github.com/hasktorch/hasktorch/blob/fa0e9c2d4d0d06228001b5bd317bb563ec6120ae/hasktorch/src/Torch/Functional.hs#L1245-L1251
[diagflat]: https://github.com/hasktorch/hasktorch/blob/fa0e9c2d4d0d06228001b5bd317bb563ec6120ae/hasktorch/src/Torch/Functional.hs#L1269-L1273
[diagonal]: https://github.com/hasktorch/hasktorch/blob/fa0e9c2d4d0d06228001b5bd317bb563ec6120ae/hasktorch/src/Torch/Functional.hs#L1277-L1284
[Offset]: https://github.com/hasktorch/hasktorch/blob/fa0e9c2d4d0d06228001b5bd317bb563ec6120ae/hasktorch/src/Torch/Functional.hs#L1253
[diag_embed]: https://pytorch.org/docs/stable/torch.html?highlight=diag_embed#torch.diag_embed
[diag]: https://github.com/hasktorch/hasktorch/blob/fa0e9c2d4d0d06228001b5bd317bb563ec6120ae/hasktorch/src/Torch/Functional.hs#L1238-L1242

mcwitt:
Fixes #416.

I also went ahead and removed the `Offset` type and instance as there weren't any other uses.

### DIFF
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -1244,11 +1244,11 @@ diag (Diag index) t = unsafePerformIO $ (cast2 ATen.tensor_diag_l) t index
 -- 
 diagEmbed
   :: Diag -- ^ offset
-  -> Int -- ^ dim1
-  -> Int -- ^ dim2
+  -> Dim -- ^ dim1
+  -> Dim -- ^ dim2
   -> Tensor -- ^ self
   -> Tensor
-diagEmbed (Diag offset) dim1 dim2 t = unsafePerformIO $ (cast4 ATen.diag_embed_tlll) t offset dim1 dim2
+diagEmbed (Diag offset) (Dim dim1) (Dim dim2) t = unsafePerformIO $ (cast4 ATen.diag_embed_tlll) t offset dim1 dim2
 
 -- | If input is a vector (1-D tensor), then returns a 2-D square tensor with the elements of input as the diagonal.
 -- If input is a tensor with more than one dimension, then returns a 2-D tensor with diagonal elements equal to a flattened input.
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -1266,12 +1266,11 @@ diagflat (Diag offset) t = unsafePerformIO $ (cast2 ATen.diagflat_tl) t offset
 -- Applying diagEmbed to the output of this function with the same arguments yields a diagonal matrix with the diagonal entries of the input. However, diagEmbed has different default dimensions, so those need to be explicitly specified.
 diagonal
   :: Diag -- ^ offset
-  -> Dimname -- ^ outdim
-  -> Dimname -- ^ dim1
-  -> Dimname -- ^ dim2
+  -> Dim -- ^ dim1
+  -> Dim -- ^ dim2
   -> Tensor -- ^ input
   -> Tensor -- ^ output
-diagonal (Diag offset) outdim dim1 dim2 t = unsafePerformIO $ (cast5 ATen.diagonal_tnnnl) t outdim dim1 dim2 offset
+diagonal (Diag offset) (Dim dim1) (Dim dim2) t = unsafePerformIO $ (cast4 ATen.diagonal_tlll) t offset dim1 dim2
 
 
 -- | Returns True if all elements in the tensor are True, False otherwise.
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -1243,12 +1243,12 @@ diag (Diag index) t = unsafePerformIO $ (cast2 ATen.tensor_diag_l) t index
 
 -- 
 diagEmbed
-  :: Offset -- ^ offset
+  :: Diag -- ^ offset
   -> Int -- ^ dim1
   -> Int -- ^ dim2
   -> Tensor -- ^ self
   -> Tensor
-diagEmbed offset dim1 dim2 t = unsafePerformIO $ (cast4 ATen.diag_embed_tlll) t offset dim1 dim2
+diagEmbed (Diag offset) dim1 dim2 t = unsafePerformIO $ (cast4 ATen.diag_embed_tlll) t offset dim1 dim2
 
 data Offset = OffsetMain | OffsetAbove | OffsetBelow deriving Show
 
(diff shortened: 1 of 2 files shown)


tscholak:
can you remind me what `Diag` is? a newtype of `Int`?

mcwitt:
> can you remind me what `Diag` is? a newtype of `Int`?

@tscholak yep, well actually it's `data Diag = Diag Int` (but it seems like it could just be a `newtype`). It's already being used as the index argument for [`diag`](https://github.com/hasktorch/hasktorch/blob/fa0e9c2d4d0d06228001b5bd317bb563ec6120ae/hasktorch/src/Torch/Functional.hs#L1239), so it seemed like the right choice in these cases too for consistency.

tscholak:
Can I ask you to add a few tests? It would make the contribution much stronger :)

hasktorch/src/Torch/Functional.hs
```diff
@@ -1266,12 +1266,11 @@ diagflat (Diag offset) t = unsafePerformIO $ (cast2 ATen.diagflat_tl) t offset
...
   :: Diag -- ^ offset
-  -> Dimname -- ^ outdim
-  -> Dimname -- ^ dim1
-  -> Dimname -- ^ dim2
+  -> Dim -- ^ dim1
+  -> Dim -- ^ dim2
   -> Tensor -- ^ input
   -> Tensor -- ^ output
-diagonal (Diag offset) outdim dim1 dim2 t = unsafePerformIO $ (cast5 ATen.diagonal_tnnnl) t outdim dim1 dim2 offset
+diagonal (Diag offset) (Dim dim1) (Dim dim2) t = unsafePerformIO $ (cast4 ATen.diagonal_tlll) t offset dim1 dim2
```
tscholak, line 1273: @junjihashimoto does `Dimname` refer here to our API for torch's new named dimensions?

tscholak: @mcwitt I pinged Junji here because I'm not sure if `Dimname` -> `Dim` is justified here.

hasktorch/test/FunctionalSpec.hs
```diff
@@ -106,6 +106,22 @@ spec = do
...
+    shape (diagflat (Diag 0) t1) `shouldBe` [3, 3]
+    shape (diagflat (Diag 1) t1) `shouldBe` [4, 4]
+    let t2 = ones' [2, 2]
+    shape (diagflat (Diag 0) t2) `shouldBe` [4, 4]
+  it "diagonal" $ do
+    let t1 = ones' [3, 3]
+    shape (diagonal (Diag 0) (Dim 0) (Dim 1) t1) `shouldBe` [3]
+    shape (diagonal (Diag 1) (Dim 0) (Dim 1) t1) `shouldBe` [2]
+    let t2 = ones' [2, 5, 4, 2]
+    shape (diagonal (Diag (-1)) (Dim 1) (Dim 2) t2) `shouldBe` [2, 2, 4]
```
tscholak, line 124: thanks for adding tests!

mcwitt:
I was a bit confused about the signature for `diagonal`, since it doesn't match up with [PyTorch API](https://pytorch.org/docs/master/generated/torch.diagonal.html). For now (4521cc7) I changed it to wrap [`diagonal_tlll`](https://github.com/hasktorch/hasktorch/blob/82da6bceccd51f8ff0a6eb1ac3c64938941870ef/libtorch-ffi/src/Torch/Internal/Unmanaged/Native.hs#L4608) for consistency with PyTorch.

hasktorch/src/Torch/Functional.hs
mcwitt: @tscholak yeah, I was unsure of that. See https://github.com/hasktorch/hasktorch/pull/417#issuecomment-653648499

tscholak:
ok! I think we need to think about how we want to support both named and unnamed dimensions in our API.

mcwitt:
> ok! I think we need to think about how we want to support both named and unnamed dimensions in our API.
It can't be randomly one or the other.

Makes sense! I can go ahead and add back the named version in this PR (but it seems like we should always support unnamed as well). Currently there are quite a few other functions in `Torch.Functional` taking `Dim`s, but looks like only one other taking `Dimname`s (`indexCopyWithDimname`).

mcwitt:
I didn't add back the version of `diagonal` accepting `Dimname` arguments for now; my thinking is that the change in this PR makes the existing API more consistent (i.e. most functions accept `Dim`s), and that it could make more sense to add "named" versions all at once and consistently in a future diff. I'm happy to add back the named version of `diagonal` in this PR if there's disagreement, though.

mcwitt:
@tscholak fyi, I added two more small commits (d29a155, fe19907) to fix the inconsistency where `diagEmbed` was taking dimension values as `Int`s, while the other diag* functions took `Dim`s. Now they all take `Dim`s.

From a quick skim it looks like `Torch.Functional` uses `Dim`s in most places, but there are a few functions that still use unwrapped `Int`s (`indexCopy`, `logsumexp`). I'm thinking those are best left for a future PR if we decide they should change.

tscholak:
@mcwitt thank you! I'm waiting for the tests to finish and @junjihashimoto to comment, otherwise this is ready to be merged

hasktorch/src/Torch/Functional.hs
junjihashimoto: > @junjihashimoto does Dimname refer here to our API for torch's new named dimensions?

Yes.　The previous API(diagonal) refered `Dimname`.
diagonal_tnnnl 's n means `Dimname`.

junjihashimoto: I think it's good to replace Dimname into Dim.
If we have a time, we can prepare a type-class called DimLike to support both.

junjihashimoto:
@mcwitt @tscholak Thx!
LGTM!

PR #421, opened 2020-07-08
Typed diagEmbed, diagflat, diagonal

mcwitt:
This PR adds typed versions of [`diagEmbed`](https://pytorch.org/docs/stable/torch.html?highlight=diag_embed#torch.diag_embed), [`diagflat`](https://pytorch.org/docs/stable/torch.html?highlight=diagflat#torch.diagflat), and [`diagonal`](https://pytorch.org/docs/stable/torch.html?highlight=diagonal#torch.diagonal).

~~This also introduces a new type/kind for signed dimensions:~~
```haskell
data Dim = PDim Nat | NDim Nat
```
~~This is useful for example in `diag_embed`, where you often want to the diagonal embedded in the last 2 dimensions (the PyTorch defaults are -2, -1). In retrospect I'm not sure it's a good idea to introduce this in this PR, since it will create some inconsistency with other functions that don't support signed dims yet. I'm happy to defer judgement on this and possibly remove the signed dimensions / save for a future PR.~~

**EDIT**: I decided to remove the signed dimension type from this PR to cut down on the complexity. I think it'd make sense to revisit this in the future after some discussion about the best interface to support natural, negative, and named dimensions (for example the `DimLike` type class proposed by @junjihashimoto https://github.com/hasktorch/hasktorch/pull/417#discussion_r451541354)

#### TODO
- [x] Eliminate some of the general-purpose types / type families introduced in this PR, or move them to `Torch.Typed.Aux` or `Torch.Typed.Tensor`
  - ~`Dim`, `KnownDim`, `UnDim`, `UnDimImpl`, `CmpDim`~ (removed these)
  - ~`Drop`~, ~`Take`~, ~`Last`~
  - ~`NatProd`~
- [x] Improve error messages
  - ~checking `dim1` comes before `dim2`: `CmpDim shape dim1 dim2 ~ 'LT`~
  - ~checking that input is at least 2D: `2 <= ListLength shape`~
- [x] Add haddocks / doctests
- [x] ~Fix overloading of "NDim" as both "number of dimensions" and "negative dimension"~ (removed `NDim`)

### DIFF
--- a/hasktorch/src/Torch/Typed/Functional.hs
+++ b/hasktorch/src/Torch/Typed/Functional.hs
@@ -2279,8 +2279,40 @@ instance (KnownNat n) => KnownDim (PDim n) where
 instance (KnownNat n) => KnownDim (NDim n) where
   dimVal = - natValI @n
 
+-- TODO: eliminate or move to 'Torch.Typed.Aux': UnDim, CmpDim, Drop, Take, Last
+type family UnDim (dim :: Dim) (last :: Nat) :: Nat where
+  UnDim (PDim dim) _ = dim
+  UnDim (NDim dim) last = last - dim
+
+type family CmpDim (shape :: [Nat]) (dim :: Dim) (dim' :: Dim) where
+  CmpDim shape dim dim' = CmpNat (UnDim dim (ListLength shape)) (UnDim dim' (ListLength shape))
+
+type family Drop (n :: Nat) (xs :: [a]) :: [a] where
+  Drop 0 xs = xs
+  Drop _ '[] = '[]
+  Drop n (x ': xs) = Drop (n - 1) xs
+
+type family Take (n :: Nat) (xs :: [a]) :: [a] where
+  Take 0 xs = '[]
+  Take _ '[] = '[]
+  Take n (x ': xs) = x ': Take (n - 1) xs
+
+type family Last (xs :: [a]) :: a where
+  Last (x ': '[]) = x
+  Last (x ': xs) = Last xs
+
+type family DiagEmbedShapeImpl' (shape :: [Nat]) (n :: Nat) (dim1 :: Nat) (dim2 :: Nat) :: [Nat] where
+  DiagEmbedShapeImpl' shape n dim1 dim2 =
+    Take dim1 shape
+      ++ (n ': Take (dim2 - dim1 - 1) (Drop dim1 shape))
+      ++ (n ': Drop dim2 shape)
+
+type family DiagEmbedShapeImpl (dim1 :: Dim) (dim2 :: Dim) (shape :: [Nat]) (n :: Nat) (outdims :: Nat) :: [Nat] where
+  DiagEmbedShapeImpl dim1 dim2 shape n outdims = DiagEmbedShapeImpl' shape n (UnDim dim1 outdims) (UnDim dim2 outdims)
+
 type family DiagEmbedShape (index :: Nat) (dim1 :: Dim) (dim2 :: Dim) (shape :: [Nat]) :: [Nat] where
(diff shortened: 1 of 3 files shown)


hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -2267,14 +2270,155 @@ det
...
+  Drop n (x ': xs) = Drop (n - 1) xs
+
+type family Take (n :: Nat) (xs :: [a]) :: [a] where
+  Take 0 xs = '[]
+  Take _ '[] = '[]
+  Take n (x ': xs) = x ': Take (n - 1) xs
+
+type family Last (xs :: [a]) :: a where
+  Last (x ': '[]) = x
+  Last (x ': xs) = Last xs
```
tscholak, line 2318: @mcwitt awesome, and I'd gladly add these, but I believe those type families already exist elsewhere. check Aux.hs or Tensor.hs

mcwitt: @tscholak gotcha, I skimmed `Torch.Typed.Aux` and didn't immediately see these, so left it as a todo to clean this up. I'll definitely give this another pass before marking ready for review.

tscholak:
looks better and better, great work @mcwitt !

junjihashimoto:

hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -2267,14 +2270,125 @@ det
...
+type family DiagEmbedShapeImpl (dim1 :: Nat) (dim2 :: Nat) (shape :: [Nat]) (n :: Nat) :: [Nat] where
+  DiagEmbedShapeImpl dim1 dim2 shape n = Insert dim1 n (Insert (dim2 - 1) n (Init shape))
+
+type family DiagEmbedShape (index :: Nat) (dim1 :: Nat) (dim2 :: Nat) (shape :: [Nat]) :: [Nat] where
+  DiagEmbedShape index dim1 dim2 shape = DiagEmbedShapeImpl dim1 dim2 shape (Last shape + index)
+
+-- | diagEmbed
+--
+-- >>> dtype &&& shape $ diagEmbed @0 @0 @1 Upper (ones :: CPUTensor 'D.Float '[2,3])
+-- (Float,[2,3,3])
```
junjihashimoto, line 2295: image
CI's error occurs here.
Should we change it to `(Float,[3,3,2])`?

mcwitt: oops, let me check! I missed that this one started failing

mcwitt: Ah, ok, I think the issue was that I incorrectly translated dimensions -2, -1 to 0, 1 (instead of the correct 1, 2) when I removed the dimension type I had introduced earlier. This should be fixed in 057a1d5.

mcwitt:
@tscholak thanks for the encouragement and reviews! 😃

hasktorch/src/Torch/Typed/Functional.hs
junjihashimoto: Thank you for your updating.
I am not familiar with negative dimention.
I think I now understand how to use it and why -2 becomes 0 ( `dim of input tensor` - 2 = 0).

tscholak:
I can't get `nix-linux` to build, it runs out of memory. I'll merge anyway, because @junjihashimoto has already some compilation time improvements lined up.

junjihashimoto:
@tscholak @mcwitt 
It is 10 times faster in an environment with a lot of cores and memory, but it requires adjustment of the number of parallels in both the number of cores and memory.

junjihashimoto:
My local environment is a chromebook with 8 cores and 8GB of memory, but it failed.

PR #425, opened 2020-07-15
Streamed datasets and a CSV reader

andredaprato:
- [x] Datastream typeclass which uses streaming to yield batches rather than a monadic action (like it is now in Pipeline.hs)
- [x] CSV dataset for reading parsing CSV data
- [x] Example usage of CSV dataset on Iris dataset
- [x] Additional dataset utilities 
- [x] Port more examples to use dataloader


junjihashimoto:
Awesome!

examples/mnist-mlp/Main.hs
```diff
@@ -47,39 +53,25 @@ mlp MLP{..} input =
...
-            pure $ replaceParameters state newParam
-    pure trained
-    where
-        spec = MLPSpec 784 64 32 10
-        dataDim = 784
-        numIters = 3000
-        batchSize = 256
-        optimizer = GD
-
+trainLoop model optimizer = FoldM step model done
```
junjihashimoto, line 56: Could you add this type? Is it `FoldM IO MLP MLP`?

hasktorch/src/Torch/Data/StreamedPipeline.hs
```diff
@@ -0,0 +1,216 @@
...
+withOne ::
+  (MonadBaseControl IO m) =>
+  Buffer a ->
+  (Output a -> m l) ->
+  (Input a -> m r) ->
+  m r
+withOne buffer fOutput fInput =
+  liftedBracket
+    (liftBase $ spawn' buffer)
+    ( \(_, _, seal) -> do liftBase $ print "here"
```
junjihashimoto, line 182: In development?

examples/typed-transformer/Main.hs
```diff
@@ -245,46 +259,59 @@ learning
...
+  -- let trainingData = TransformerData { length = trainingLen , }
+  -- let trainingData = 
+  let collatedTrain :: CollatedDataset m (TransformerData seqLen) [Maybe Int] (input, target)
+      collatedTrain = CollatedDataset { set = trainingData, chunkSize = natValI @batchSize, collateFn = collation }
+      collatedEval :: CollatedDataset m (TransformerData seqLen) [Maybe Int] (input, target)
+      collatedEval = CollatedDataset { set = evaluationData, chunkSize = natValI @batchSize, collateFn = collation }
+  in for (each [1 .. numEpochs]) $ \_epoch -> do
+    (model', optim') <- lift $ foldOverWith' collatedTrain id $ training @workerDevices @modelDevice @dataDevice @dtype @model @models @optim @input @inputs @target @targets @inputTargets @losses @parameters' @gradients @parameters @tensors learningRate (model, optim) 
+    evalLoss' <- lift $ foldOverWith' collatedEval id $ evaluation @workerDevices @modelDevice @dataDevice @numEmbeds @batchSize @seqLen @dtype @model @models @input @inputs @output @outputs @target model'
```
junjihashimoto, line 283: I think foldOverWith is better to export numWorkers.

examples/iris-classification/Main.hs
```diff
@@ -0,0 +1,119 @@
...
+
+  -- inputs <- makeListT @_ @_ @_ @[Iris] defaultDataloaderOpts irisTrain (Select $ yield ()) 
+  -- transformed <- pMap inputs Main.toTensors 2
+
+  foldM (\model epoch -> do
+            inputs <- makeListT @[Iris] defaultDataloaderOpts irisTrain (Select $ yield ()) 
+            transformed <- pMap inputs Main.toTensors 2
+            -- liftIO $ print epoch
+            trainLoop model optimizer $ enumerate transformed
+        ) init [1]
```
junjihashimoto, line 112: I think dataloader has some waits when there are multiple epochs.

Is it possible to put the dataloader in an outer loop like this?
`replicate numEpoch dataload >-> transformer >-> training`
Does outer loop has any disadvantages?

hasktorch/src/Torch/Data/StreamedPipeline.hs
junjihashimoto: https://hackage.haskell.org/package/pipes-concurrency-2.0.12/docs/src/Pipes.Concurrent.html#spawn%27
seal implementation is here.
I will deepen my understanding later.

andredaprato: This should be fixed in the current version.

examples/iris-classification/Main.hs
andredaprato: I'm not sure how we can avoid this, but yes there are some waits, in that the threads are killed and generated again on each epoch. I think the code you have written would still have this problem, I'll think more about this to see if it could be avoided but it shouldn't be too big of a bottleneck on real datasets.

PR #419, opened 2020-07-08
Reduce compilation time

junjihashimoto:
libtorch-ffi's compilation time takes 10-20ms.
To speedup compilation time, I did the following.

- Splite modules of both Native and Tensor 
- Remove unused imports and tweak dependencies of modules.
- Replace large c++ headers like ATen/ATen with small headers like ATen/Tensor
- Reduce generated codes of inline-c-cpp(https://github.com/fpco/inline-c/pull/112)

Before 
```
# 56 cores
$ time cabal v2-build libtorch-ffi -j --ghc-options "-j +RTS -A128m -n2m -RTS"
real    6m2.483s
user    20m30.915s
sys     1m49.478s

# 56 cores
$ time cabal v2-build libtorch-ffi --ghc-options=-j
real    10m1.120s
user    130m37.817s
sys     35m52.595s

$ time cabal v2-build libtorch-ffi
real    15m16.017s
user    14m54.103s
sys     0m49.186s

```

After
```
# 56 cores
$ time cabal v2-build libtorch-ffi -j --ghc-options "-j +RTS -A128m -n2m -RTS"
...
real    1m27.280s
user    19m40.053s
sys     1m45.472s

# 8cores
$ time cabal v2-build libtorch-ffi -j8 --ghc-options "-j8 +RTS -A128m -n2m -RTS"
...
real    2m27.052s
user    14m14.518s
sys     0m52.018s

# 4cores
$ time cabal v2-build libtorch-ffi -j4 --ghc-options "-j4 +RTS -A128m -n2m -RTS"
real    4m3.585s
user    13m50.770s
sys     0m50.311s

$ time cabal v2-build libtorch-ffi
real    14m12.646s
user    13m54.727s
sys     0m54.099s
```

This PR adds ghc-options  "-j +RTS -A128m -n2m -RTS" to cabal.project file.
So you do not need to specify the options via command-line-argument.

PR #427, opened 2020-07-23
Automatic adjustment of used resources

junjihashimoto:
To reduce compilation time, we try https://github.com/hasktorch/hasktorch/pull/419.
After that, OSKill has come to happen frequently on CI.
I've tried the following to solve it:

- Automatically adjust the parallel number from the spec of system memory and CPU.
- Split Torch.Typed.FunctionalSpec and Torch.Typed.TensorSpec to reduce allocated memory.
- Set the upper limit of memory used by the compiler

I feel that the adjustment of the parallel number is not working well on nix,
because OSkill happened instread of heap-error with setting the upper limit of memory.


setup-cabal.sh
```diff
@@ -14,6 +14,22 @@ sed -e 's/inline-c ==.*,/inline-c ==0.9.0.0,/g' \
...
+  "Linux")
+      TOTAL_MEM_GB=`grep MemTotal /proc/meminfo | awk '{print int($2/1024/1024)}'`
+      NUM_CPU=$(nproc --all)
+    ;;
+esac
+
+USED_MEM_GB=`echo $TOTAL_MEM_GB | awk '{print int(($1 + 1) / 2)}'`
+USED_NUM_CPU=`echo $NUM_CPU | awk '{print int(($1 + 1) / 2)}'`
+USED_NUM_CPU=`echo $USED_MEM_GB $USED_NUM_CPU | awk '{if($1<x$2) {print $1} else {print $2}}'`
+USED_MEM_GB=`echo $USED_NUM_CPU | awk '{print int($1 * 4 / 3)"G"}'`
```
junjihashimoto, line 31: This pr tries conservative allocation of resources. 
It uses half of the cores and 1.3GB of memory per core.
I feel a little wrong. For now, I want to try it with CI.

junjihashimoto:
image

junjihashimoto:
image
It is 1GB instead of 1.3GB.

junjihashimoto:
When using cabal, CI passes.
When using nix, CI fails.
Both uses 'ghc-options: -j1 +RTS -A128m -n2m -M2G -RTS ' which means the parallel number is 1 and the maximum memory value is 2 GB.
The system has 7GB of memory, so it should pass even if two other processes run.
Why does only nix fail?

junjihashimoto:
When this commit still can't pass, I will consider splitting `Torch.Typed.FunctionalSpec`.

hasktorch/hasktorch.cabal
```diff
@@ -138,9 +138,12 @@ test-suite spec
...
                     , PipelineSpec
                     , Torch.Typed.AuxSpec
-                    , Torch.Typed.TensorSpec
+                    , Torch.Typed.TensorSpec0
+                    , Torch.Typed.TensorSpec1
                     , Torch.Typed.FactoriesSpec
-                    , Torch.Typed.FunctionalSpec
+                    , Torch.Typed.FunctionalSpec0
+                    , Torch.Typed.FunctionalSpec1
+                    , Torch.Typed.FunctionalSpec2
```
junjihashimoto, line 146: I don't like this decomposition very much, but it is necessary to reduce memory at compile time.

.github/workflows/nix-macos.yml
```diff
@@ -27,8 +27,8 @@ jobs:
       - run: cachix use iohk
       - run: |
           nix-build -j 2 -A haskellPackages.libtorch-ffi.checks.spec
-          nix-build -j 2 -A haskellPackages.hasktorch.checks.spec
-          nix-build -j 2 -A haskellPackages.examples.checks.spec
+          nix-build -j 1 -A haskellPackages.hasktorch.checks.spec
+          nix-build -j 1 -A haskellPackages.examples.checks.spec
```
junjihashimoto, line 31: Before disassembling FunctionalSpec, OSkill happened  with `-j1`. `-j1`  intends to reduce memory. I'm not sure why OSkill happened.

setup-cabal.sh
```diff
@@ -14,6 +14,23 @@ sed -e 's/inline-c ==.*,/inline-c ==0.9.0.0,/g' \
...
+      TOTAL_MEM_GB=`grep MemTotal /proc/meminfo | awk '{print int($2/1024/1024)}'`
+      NUM_CPU=$(nproc --all)
+    ;;
+esac
+
+USED_MEM_GB=`echo $TOTAL_MEM_GB | awk '{print int(($1 + 1) / 2)}'`
+USED_NUM_CPU=`echo $NUM_CPU | awk '{print int(($1 + 1) / 2)}'`
+USED_NUM_CPU=`echo $USED_MEM_GB $USED_NUM_CPU | awk '{if($1<x$2) {print $1} else {print $2}}'`
+USED_MEM_GB=`echo $USED_NUM_CPU | awk '{print ($1)"G"}'`
+USED_MEMX2_GB=`echo $USED_NUM_CPU | awk '{print ($1 * 2)"G"}'`
```
junjihashimoto, line 32: Hasktorch-package seems to need 1.5GB of memory per core. (This line sets 2GB memory per core for margin.)

hasktorch/src/Torch/Tensor.hs
```diff
@@ -423,7 +423,7 @@ float_opts = withDType Float defaultOpts
 withTensor :: Tensor -> (Ptr () -> IO a) -> IO a
 withTensor t fn = cast t $ \t' -> withForeignPtr t' $ \tensor_ptr -> Unmanaged.tensor_data_ptr tensor_ptr >>= fn
 
-instance (Reifies a DType, Storable a) => TensorLike a where
+instance {-# OVERLAPPING #-} (Reifies a DType, Storable a) => TensorLike a where
```
junjihashimoto, line 426: This fixes a compilation error of distill-example.

junjihashimoto:
@tscholak Thx!

PR #432, opened 2020-07-26
Csv cleanup, Torch.Tensor fix

andredaprato:
- [x] Makes some small fixes to the csv dataset which makes the constraints more usable.
- [x] Removes the IncoherentInstances extension from Torch.Tensor that I accidently added in my previous pr (this fixes Austin's 
      regularization example). 
- [x] Add a functional dependency to datastream.

### DIFF
--- a/hasktorch/src/Torch/Data/StreamedPipeline.hs
+++ b/hasktorch/src/Torch/Data/StreamedPipeline.hs
@@ -22,6 +22,8 @@ module Torch.Data.StreamedPipeline
     Datastream (..),
     MonadBase (..),
     MonadBaseControl (..),
+    module Pipes,
+    module Pipes.Safe
   )
 where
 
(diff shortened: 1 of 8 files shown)


austinvhuang:
Instead of importing `one_hot` from `Internal`, can you add `oneHot` to `Torch.Functional` with standard minor tweaks of the user-facing versions (camel case, curry-friendly config-parameters first and input tensor last)?

hasktorch/src/Torch/Data/StreamedPipeline.hs
```diff
@@ -43,7 +46,7 @@ import qualified Pipes.Safe as Safe
 import           Torch.Typed
 
 
-class (MonadBase IO m) => Datastream m seed dataset batch where
+class (MonadBase IO m) => Datastream m seed dataset batch | dataset -> batch where
```
tscholak, line 49: thanks!

andredaprato:
A named dataset lets you choose the names of the columns in the csv you want to decode.

tscholak:
@andredaprato merge?

andredaprato:
Should be good to go unless @austinvhuang is interested in seeing anything changed.

tscholak:
I'll merge it for now.

PR #437, opened 2020-08-03
have grad return Gradients

tycho01:


tycho01:
ohh I just realized we'd intentionally created a `grad` `grad'` distinction over this. closing then.

PR #438, opened 2020-08-05
add stylish-haskell file

Cmdv:
just adding [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) file.

It's set to default which seems nice but can be changed to preferred style.

I've done a single file to show the results and it looks like it's only really effecting the imports and flags.

I'm happy to go over more files if need be 😸


.stylish-haskell.yaml
```diff
@@ -0,0 +1,260 @@
...
+  - trailing_whitespace: {}
+
+  # Squash multiple spaces between the left and right hand sides of some
+  # elements into single spaces. Basically, this undoes the effect of
+  # simple_align but is a bit less conservative.
+  # - squash: {}
+
+# A common setting is the number of columns (parts of) code will be wrapped
+# to. Different steps take this into account. Default: 80.
+columns: 80
```
tscholak, line 215: 80 is really low for this code. I wonder how it is enforced. Your experiments with Torch.Typed.NN.Transformer didn't break lines, did they?

hasktorch/src/Torch/Distributions/Bernoulli.hs
```diff
@@ -33,7 +33,7 @@ instance Distribution Bernoulli where
     sample d = D.bernoulliIO' . F.expand (probs d) False . extendedShape d
     logProb d value = F.mulScalar (-1 :: Int) (bce' (logits d) value)
     entropy d = bce' (logits d) $ probs d
-    enumerateSupport d doExpand = 
+    enumerateSupport d doExpand =
         (if doExpand then \t -> F.expand t False ([-1] <> batchShape d) else id) values
         where
             values = D.reshape ([-1] <> replicate (length $ batchShape d) 1) $ D.asTensor [0.0, 1.0 :: Float]
```
tscholak, line 39: so far so good :) happy to try this on more!

.stylish-haskell.yaml
Cmdv: I can up it, I just used the default leaving open for everyones preference.
I've gone with 2 spaces rather than tabs other than that feel free to let me know preferences.

hasktorch/src/Torch/Distributions/Bernoulli.hs
Cmdv: lots more files incoming!!

.stylish-haskell.yaml
Cmdv: not it didn't seem to or on the files that I'm working on now so not sure what's happening there!!

Cmdv:
Well this make it look like a massive contribution!! when really it was just repeated key strokes and fixing the odd issues 😂

hasktorch/src/Torch/Distributions/Bernoulli.hs
tscholak: You’re on your way to becoming the stuff of folklore and legends, your soul being forever immortalized.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
```diff
@@ -200,14 +200,14 @@ instance A.Parameterized (LSTMLayer inputSize hiddenSize directionality dtype de
...
-  (device :: (D.DeviceType, Nat))
-  = LSTMLayerStackSpec deriving (Show, Eq)
+data LSTMLayerStackSpec (inputSize :: Nat) (hiddenSize ::
+                                         Nat) (numLayers :: Nat) (directionality ::
+                                                                    RNNDirectionality) (dtype ::
+                                                                                          D.DType) (device
+                                                                                                      ::
+                                                                                                      (D.DeviceType,
+                                                                                                       Nat)) = LSTMLayerStackSpec
+    deriving (Show, Eq)
```
junjihashimoto, line 211: Can we arrange the lines?

junjihashimoto:
Thank you for setting and applying.

.stylish-haskell.yaml
```diff
@@ -0,0 +1,266 @@
...
+      #
+      # LANGUAGE pragma redundancy checking is disabled because most of our
+      # pragmas are set at the project level.
+      remove_redundant: false
+
+  - records:
+      deriving: 2
+      equals: 'indent 2'
+      field_comment: 2
+      first_field: 'same_line'
```
junjihashimoto, line 203: I think it is good, because these settings shows following codes.

```haskell
-- equals: "indent 2", first_field: "same_line"
data Foo a
  = Foo { a :: Int
        , a2 :: String
          -- ^ some haddock
        }
  | Bar { b :: a
        }
  deriving (Eq, Show)
  deriving (ToJSON) via Bar Foo
```
https://github.com/jaspervdj/stylish-haskell


But there is the follow codes in this PR.

```
data DType = Bool
    | UInt8
    | Int8
    | Int16
    | Int32
    | Int64
    | Half
    | Float
    | Double
    deriving (Eq, Show)
```

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
tscholak: Yes, this looks horrible :(

junjihashimoto:
`Torch.Typed.Functional` and `Torch.Typed.Tensor` looks good.

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
Cmdv: oops my bad, sometimes it does crazy stuff didn't spot this. Will change

.stylish-haskell.yaml
Cmdv: I couldn't work that out and kept playing with it, seems the whole `-records` is a bit of a lie. I'll check the issues in their repo see if anyone else has come across this. :)

experimental/dataloader-cifar10/Main.hs
```diff
@@ -1,46 +1,48 @@
...
 
-data MLPSpec = MLPSpec {
-    inputFeatures :: Int,
-    hiddenFeatures0 :: Int,
-    hiddenFeatures1 :: Int,
-    outputFeatures :: Int
-    } deriving (Show, Eq)
+data MLPSpec = MLPSpec
+    { inputFeatures :: Int
+    , hiddenFeatures0 :: Int
```
tscholak, line 24: sometimes it moves the comma to the front (like here), sometimes it doesn't (like in examples/vae/Main.hs). can you investigate, please?

Cmdv:
I've not continued with this as had a bit more trouble with my nix setup on osx, I might do the sum types manually as in the tests for stylish haskell it seems to do exactly what I've asked it to do:

https://github.com/jaspervdj/stylish-haskell/blob/b8a731eb948b98019b8663c6fc653d2c930df2b1/tests/Language/Haskell/Stylish/Tests.hs#L43-L64

It's a weird one might look into stylish-haskell but it's a little strange to have records do the same think for data types that don't actually have any records. Will see 😄

tscholak:
I was reviewing this file by file, but it proves to be challenging since it's so much code.

Cmdv:
yeah I did bombarded it!! maybe I do different PR's for different folders split it up a bit?

Cmdv:
Going to close this in favour of Ormolu.

What I'll do is incremental PR's so there isn't 159 files in one PR 😅

hasktorch/src/Torch/Typed/NN/Recurrent/LSTM.hs
austinvhuang: Is this the ormolu default? do you have to manually change it back per instance or is there a flag that can keep it from doing this?

PR #439, opened 2020-08-06
initial tensorboard port

KiaraGrouwstra:
this is a watered-down port of the Tensorboard functionality from `haskell-torch` (specifically, commenting functionality related to e.g. images to focus on the easy bits first).
I hope I haven't messed up any of the license-y bits (@abarbu?).
the whole protobuff stuff needed [`protoc` installed](https://github.com/google/proto-lens/blob/master/docs/installing-protoc.md), so I guess that still needs to be incorporated in our dev-opsy stuff (CI / Nix / Docker)...


junjihashimoto:
I would like to merge if this pr have a test for writing tensorboard logs.

abarbu:
@tycho01  Cool! Would be nice to share things like this more officially so we don't have to duplicate efforts.

tycho01:
@abarbu agreed! I imagine that might also address the others' concern here about licensing stuff.

on finding a common abstraction tho, I'm still kinda trying to figure out how that might look.

another concern I heard here was about the protoc dependency -- it'd be nice if such a common dependency could have kind of handled those bits in such a way that the consuming frameworks would no longer have to like rerun those bits... but my knowledge of the protocol buffers ecosystem is a bit lacking for that.

PR #440, opened 2020-08-07
New untyped bindings from 1.6, clean up RNN implementations

austinvhuang:
- Add bindings to several new 1.6 functions for untyped.
- fix untyped rnn implementations (dimension and initialization issues)
- add rnn layer bindings


tscholak:
this looks very good! thanks for implementing all the missing ops along the way

PR #441, opened 2020-08-09
Change get-deps.sh to brew-install

junjihashimoto:
This PR changes get-deps.sh to brew-install for macos.
When we use this, we do not have to set extra library settings like `DYLD_LIBRARY_PATH`.

### DIFF
--- a/.github/workflows/stack-macos.yaml
+++ b/.github/workflows/stack-macos.yaml
@@ -19,14 +19,16 @@ jobs:
         wget -qO- https://get.haskellstack.org/ | sh || true
         clang --version
         stack --version
-        pushd deps/ ; ./get-deps.sh -a cpu -c ;popd 
+        brew tap hasktorch/libtorch-prebuild https://github.com/hasktorch/homebrew-libtorch-prebuild || true
+        brew install libtorch-prebuild || true
+        #pushd deps/ ; ./get-deps.sh -a cpu -c ;popd
     - name: Build
       run: |
-        . setenv
+        #. setenv
         stack build
     - name: Test
       run: |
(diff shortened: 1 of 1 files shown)


austinvhuang:
Thanks for looking into this @junjihashimoto !

I'm not sure how homebrew works under the hood. Do we ensure the homebrew tap is in sync with the master branch? 

Should we assume/ask all mac users to use homebrew or would get-deps.sh still be supported? If the former, may want to consider how universal homebrew adoption is among mac users, if the latter, we should factor in some maintenance overhead of supporting both.

Not to make things even more complicated, but sometimes conda gets used to package system dependencies of non-python packages (iirc nimtorch used it - see [here](https://github.com/sinkingsugar/nimtorch#super-easy-using-conda) and [here](https://anaconda.org/fragcolor/nimtorch)). A large fraction of people using hasktorch would have conda through being familiar with the data science / ml stacks. Maybe it's not the right solution but might be worth checking into.

junjihashimoto:
@austinvhuang 
> Do we ensure the homebrew tap is in sync with the master branch?
Yes.
There are several ways to install the library of specific version.
https://stackoverflow.com/questions/3987683/homebrew-install-specific-version-of-formula/9832084

junjihashimoto:
I want users using high-level-api to stop using `get-deps.sh`. It's nice for users using code generation or low-level-api, but it is not useful for users using high-level-api.

I haven't tried conda so I'm not sure which one is better, brew or conda.
I think it is better to choose the one that is easier to set up for xcode.

austinvhuang:
Okay let's give it a try 👍

PR #442, opened 2020-08-09
Remove pytorch-repo

junjihashimoto:
For stack's bug, hasktorch's repo can not be cloned on stack.
We already have the bug-fix on the master branch of stack, but I'm not sure when it will be released.
This PR stops the pytorch submodule and embeds the tag used in get-deps.sh.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "deps/pytorch"]
-	path = deps/pytorch
-	url = https://github.com/pytorch/pytorch.git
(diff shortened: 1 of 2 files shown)


tscholak:
@junjihashimoto do we keep it now that the new stack is released?

junjihashimoto:
I've just found something very disappointing.
 stack-2.3.3 uses patry-0.4.* not pantry-0.5.1.1 which is fixed for relative links.
https://github.com/commercialhaskell/stack/tree/v2.3.3
https://github.com/commercialhaskell/stack/commit/f36409cb29494a61efdd71b42438808864e413b5

junjihashimoto:
I think we should keep this.

PR #443, opened 2020-08-11
Map Style Datasets

andredaprato:
This adds support for indexable fixed size datasets, similar to pytorch's map-style datasets. The API for running an epoch over a map style dataset is mostly the same as using a streaming style dataset with two differences. The first is a `collate` function, which definies how each worker will take a sequence of samples to transform them into a minibatch (our mnist implementation batches within the dataset retreival so we just use `id` for collation in this case). The second difference is using a `sampler` which allows you to specify different indexes than purely sequential iteration.  I also added an option to specify the buffer size for both streamed and map-style datasets, since one would want to adjust this based on the number of threads. 

Additionally, there was a disparity between the typed and untyped MNIST image retrieval functions, the typed API returns pure tensors, while the untyped API returned tensors IO, I adjusted the untyped version to return pure tensors as well. 

FYI, I think something might be up with the typed optimizers, the static-mnist examples should be converging but they aren't. This was happening before changing them to use the new data loading utilities.


junjihashimoto:
The optimizer hasn't converged since testing in May.
https://github.com/hasktorch/hasktorch/wiki/Benchmark-comparison-with-pytorch

tscholak:
For the loss issues, I think we should run the typed and the untyped optimizers side-by-side and compare.

tscholak:
> When I use more than one CPU, I still want that shuffling is deterministic. That means that the order in which the examples become available for training is the same whether or not I’m using multiple CPUs.

examples/static-xor-mlp/Main.hs
```diff
@@ -78,41 +82,45 @@ xor t = (1 - (1 - a) * (1 - b)) * (1 - (a * b))
...
          , RandDTypeIsValid device 'Float
          , ComparisonDTypeIsValid device 'Float
-         ) => Dataset IO Xor (Tensor device 'Float '[batchSize, 2]) where
-  getBatch _ _ =  toDType @'Float @'Bool .
-                  gt (toDevice @device (0.5 :: CPUTensor 'Float '[]))
-                  <$> rand @'[batchSize, 2] @'Float @device
-  numIters = iters 
-
+         ) => Datastream IO () (Xor device batchSize) (Tensor device 'Float '[batchSize, 2]) where
+  --TODO: has to be datastream because getItem is an IO action
```
tscholak, line 92: let's talk about that :)

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -11,145 +12,79 @@
...
+makeListT :: forall sampler dataset sample batch m b. (Dataset dataset sample, MonadBaseControl IO m) =>
+  MapStyleOptions -> dataset -> Sampler m -> CollateFn sample batch m -> ContT b m (ListT m (batch, Int))
+makeListT MapStyleOptions{..} dataset sampler collateFn = runWithBuffer bufferSize streamBatches
+  where
+    streamBatches output = forConcurrently_ [0..numWorkers - 1]  (runWorker output)
+    runWorker output = runEffect . yieldItem output
+    -- yieldItem output workerId = sampler (indices workerId) (size dataset) >-> pipeItems  >-> collateFn >->  toOutput' output
+    yieldItem output workerId = collateFn (sampler (indices workerId) (size dataset) >-> pipeItems) >->  toOutput' output
+    pipeItems = for cat (yield . getItem dataset)
+    indices workerId = (workerId * (size dataset `div` numWorkers), (workerId + 1) * (size dataset `div` numWorkers) - 1)
```
tscholak, line 77: if possible, I'd like batching and pulling from the dataset to be separated.

andredaprato: Frankly, I don't really even know why I approached it like this. It's not how it is done in the streaming interface, and I agree it shouldn't be done here. I think I was trying to make it look like pytorch or something.

andredaprato:
Right, currently this doesn't yield deterministically with more than CPU. I thought more about your suggestion of using a `Map` of `TVar (Maybe Value)` and this makes a lot of sense for achieving a deterministic ordering.

andredaprato:
I think this should be ready for another review @tscholak

hasktorch/src/Torch/Data/Internal.hs
```diff
@@ -0,0 +1,81 @@
...
+import Control.Monad.Trans.Control 
+import Pipes
+import Pipes.Concurrent
+import Control.Exception.Safe (finally, bracket)
+import Control.Concurrent.Async.Lifted (concurrently)
+import Control.Monad (when)
+import qualified Pipes.Prelude as P
+import Control.Monad.Cont (ContT(ContT))
+
+runWithBuffer :: forall batch m b .
```
tscholak, line 17:
```suggestion
runWithBuffer :: forall a m b .
```
it's not necessarily a batch, is it?

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -11,145 +14,127 @@
...
+                                       , numWorkers :: Int
+                                       , shuffle :: Sample
+                                       }
+
+mapStyleOpts numWorkers = MapStyleOptions { bufferSize = numWorkers
+                                          , numWorkers = numWorkers
+                                          , shuffle = Sequential
+                                          }
+
+data Sample = Sequential | Shuffle StdGen
```
tscholak, line 61:
```suggestion
data Sample = Sequential | RandomGen g => Shuffle g
```
?
I'm actually not sure if this valid syntax for ADTs.

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -11,145 +14,127 @@
...
+
+---------------------- Workflow --------------------
+-- - make a new map of keys to TVars of samples, possibly shuffled keys, tracking which keys have been sampled
+-- - create a TQueue of keys (using pipes-concurrency wrapper)
+-- - fork off workers which all pull from the TQueue and sample that key using the dataset,
+--   then update the TVar associated with that key
+-- have a worker waiting for each successive key to be updated in the list of (key, TVar)
+
+makeListT :: forall m dataset k sample r .
+  (MonadIO m, MonadBaseControl IO m,  Dataset m dataset k sample) =>
```
tscholak, line 75: can you make a lifted version of `atomically` that works with `MonadBaseControl IO m`?

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -11,145 +14,127 @@
...
+  (keyOutput, keyInput, seal) <- liftIO $ spawn' unbounded
+
+  let retreiveSet = liftIO $ keyTVarSet $ keys @m dataset
+  keyTVarSet <- case shuffle of
+    Sequential -> retreiveSet
+    Shuffle g -> fst . fisherYates g <$> retreiveSet
+   
+  -- fill the queue with each key and associated TVar then seal it
+  keyQueue keyOutput $ keyTVarSet
+  liftIO $ atomically seal
```
tscholak, line 89: is this working as it is supposed to?

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -11,145 +14,127 @@
...
+  (MonadIO m, MonadBaseControl IO m,  Dataset m dataset k sample) =>
+  MapStyleOptions ->
+  dataset ->
+  ContT r m  (ListT m (sample, Int))
+makeListT MapStyleOptions{..} dataset = do
+  (keyOutput, keyInput, seal) <- liftIO $ spawn' unbounded
+
+  let retreiveSet = liftIO $ keyTVarSet $ keys @m dataset
+  keyTVarSet <- case shuffle of
+    Sequential -> retreiveSet
```
tscholak, line 84:
```suggestion
    Sequential -> retrieveSet
```
:)

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -11,145 +14,127 @@
...
+  m ()
+awaitNextItem tvars output  = runEffect $ each tvars >-> readNextItem >-> toOutput' output
+  where readNextItem  = forever $ do
+          (key, tvar) <- await
+          item <- liftIO $ atomically $ do
+            val <- readTVar tvar 
+            case val of
+              Nothing -> retry
+              Just item -> writeTVar tvar Nothing >> pure item -- reset the tvar once we get the sample out of it to save memory
+          yield item
```
tscholak, line 119: during our call this morning, Adam remarked that it is possible that when one of the workers above stalls, the other workers will continue filling up the tvars, potentially until memory runs out, because the downstream consumer here is blocked.
You said this can be resolved easily, but I am missing this here.
A possible solution would be to restrict the number of keys that are pushed to the `keyInput` in such a way that it can never go beyond some `n` keys after the last unfulfilled tvar.

hasktorch/src/Torch/Data/Pipeline.hs
```diff
@@ -11,145 +14,127 @@
...
+  MapStyleOptions ->
+  dataset ->
+  ContT r m  (ListT m (sample, Int))
+makeListT MapStyleOptions{..} dataset = do
+  (keyOutput, keyInput, seal) <- liftIO $ spawn' unbounded
+
+  let retreiveSet = liftIO $ keyTVarSet $ keys @m dataset
+  keyTVarSet <- case shuffle of
+    Sequential -> retreiveSet
+    Shuffle g -> fst . fisherYates g <$> retreiveSet
```
tscholak, line 85: It could be nice to return the new random generator in the same continuation so that it is available for the next epoch.
can you make the return type
```
ContT r m (ListT m (sample, Int), Sample)
```
?

hasktorch/src/Torch/Data/StreamedPipeline.hs
```diff
@@ -75,37 +75,24 @@ pmap' n f  prod = ContT $ \cont ->
       (\output -> runEffect $ enumerate prod >-> f >-> toOutput output)
       (\input -> cont . Select $ fromInput input)
 
+
 makeListT ::
   forall batch m dataset seed b r.
   (Datastream m seed dataset batch, MonadBaseControl IO m, MonadBase IO m) =>
```
tscholak, line 81:
```suggestion
  (Datastream m seed dataset a, MonadBaseControl IO m, MonadBase IO m) =>
```
again, I'm not sure it's necessary for these to be batches. What do you think?

hasktorch/src/Torch/Typed/Vision.hs
```diff
@@ -14,16 +18,36 @@ import qualified Data.ByteString.Lazy          as BS.Lazy
...
+  Dataset m (Mnist device batchSize) Int (Tensor device 'D.Float '[batchSize, 784], Tensor device 'D.Int64 '[batchSize]) where
+
+  getItem Mnist{..} ix =  
+    let
+      batchSize = natValI @batchSize
+      -- indexes = [ix * batchSize .. (ix+1) * batchSize - 1]
+      indexes = [ix * batchSize .. (ix+1) * batchSize - 1]
+      imgs =  getImages @batchSize mnistData indexes
+      labels =  getLabels @batchSize mnistData indexes
+    in pure (toDevice @device imgs, toDevice @device labels)
```
tscholak, line 48: nice!

hasktorch/src/Torch/Typed/Vision.hs
```diff
@@ -14,16 +18,36 @@ import qualified Data.ByteString.Lazy          as BS.Lazy
...
+
+data Mnist (device :: (D.DeviceType, Nat) ) (batchSize :: Nat) = Mnist { mnistData :: MnistData }
+
+instance (KnownNat batchSize, KnownDevice device, Applicative m ) =>
+  Dataset m (Mnist device batchSize) Int (Tensor device 'D.Float '[batchSize, 784], Tensor device 'D.Int64 '[batchSize]) where
+
+  getItem Mnist{..} ix =  
+    let
+      batchSize = natValI @batchSize
+      -- indexes = [ix * batchSize .. (ix+1) * batchSize - 1]
```
tscholak, line 44: remove please

hasktorch/src/Torch/Typed/Vision.hs
```diff
@@ -14,16 +18,36 @@ import qualified Data.ByteString.Lazy          as BS.Lazy
...
+      indexes = [ix * batchSize .. (ix+1) * batchSize - 1]
+      imgs =  getImages @batchSize mnistData indexes
+      labels =  getLabels @batchSize mnistData indexes
+    in pure (toDevice @device imgs, toDevice @device labels)
+
+  keys Mnist{..} = fromList [ 0 .. Torch.Typed.Vision.length mnistData `Prelude.div` (natValI @batchSize) - 1]
```
tscholak, line 50: good!

hasktorch/src/Torch/Vision.hs
```diff
@@ -62,11 +64,21 @@ instance (MonadPlus m, MonadBase IO m) => Datastream m Int Mnist (Tensor, Tensor
...
+      where numIters = I.length mnistData `Prelude.div` batchSize
             
+instance Applicative m => Dataset m Mnist Int (Tensor, Tensor) where
+  getItem Mnist{..} ix =  
+    let
+      indexes = [ix * batchSize .. (ix+1) * batchSize - 1]
+      imgs = getImages' batchSize 784 mnistData indexes
+      labels = getLabels' batchSize mnistData indexes
+    in pure (imgs, labels)
+  -- size Mnist{..} = I.length mnistData `Prelude.div` batchSize
```
tscholak, line 79: please remove

hasktorch/src/Torch/Vision.hs
```diff
@@ -77,8 +89,8 @@ getImages' ::
   -> Int -- dimensionality of the data
   -> I.MnistData -- mnist data representation
   -> [Int] -- indices of the dataset
-  -> IO Tensor
-getImages' n dataDim mnist imageIdxs = do
+  -> Tensor
+getImages' n dataDim mnist imageIdxs = unsafePerformIO $ do
```
tscholak, line 93: ok

hasktorch/test/PipelineSpec.hs
```diff
@@ -9,65 +10,80 @@ import Control.Monad
...
+data MockData = MockData 
+data ConcurrentData = ConcurrentData 
+data ShuffleSet = ShuffleSet
 
-instance Dataset IO MockData Int where
-  getBatch (MockData iters) _ = threadDelay 10000 >> pure iters 
-  numIters (MockData iters)  = iters
+-- | Yields 4 items, each item taking 10000 milliseconds to compute
+instance Dataset IO ConcurrentData Int Int where
+  getItem ConcurrentData  _ = threadDelay 10000 >> pure 0
```
tscholak, line 35:
```suggestion
  getItem ConcurrentData k = threadDelay 10000 >> pure k
```
that way you can check that the actually come in the right order

hasktorch/test/PipelineSpec.hs
```diff
@@ -9,65 +10,80 @@ import Control.Monad
...
+  let options = (mapStyleOpts 4) { shuffle = Shuffle $ mkStdGen 123 }
+  let optionsDiff = (mapStyleOpts 4) { shuffle = Shuffle $ mkStdGen 50 }
+
+  datasets <- replicateM 100 $ runContT (makeListT options ShuffleSet) $ P.toListM . enumerate 
+  differentOrder <- runContT (makeListT optionsDiff ShuffleSet) $ P.toListM . enumerate 
 
-takeBatchThenTimeout :: Int -> Int -> Int -> IO Int
-takeBatchThenTimeout timeout _ input =  threadDelay timeout >> pure input
+  all (\elems -> elems == head datasets) datasets `shouldBe` True
+  head datasets == differentOrder `shouldBe` False
```
tscholak, line 77: ok, these are good properties to test, nice!

hasktorch/src/Torch/Data/Pipeline.hs
andredaprato: Good idea!

hasktorch/src/Torch/Data/Pipeline.hs
andredaprato: Yes, at least I think so. It allows the queue to be read from in order to update the main list of TVars, but not written to again.

hasktorch/src/Torch/Data/Pipeline.hs
andredaprato: I changed it to a GADT with this constraint

hasktorch/src/Torch/Data/Internal.hs
andredaprato: Nope, fixed.

hasktorch/src/Torch/Data/StreamedPipeline.hs
andredaprato: I agree, I changed the parameter to `sample` to be in line with the map style datasets.

PR #444, opened 2020-08-14
Bump haskell nix

tscholak:
- bump haskell.nix, hackage, stackage
- ghc 8.8.4
- remove ghcide from shell
- add haskell-language-server to shell


Cmdv:
look forward to trying this out as had terrible problems with getting ghcide with hasktorch 👍

PR #450, opened 2020-08-28
revamp Parameterized typeclass and add misc additional functionality

tscholak:
I've put everything non-experimental in #447 in this PR because these things are ready to be merged and make live easier for everyone.

The biggest change is the new `Torch.Typed.Parameter.Parameterized` type class:
```haskell
class
  Parameterized (f :: Type)
  where
  type Parameters f :: [Type]
  type Parameters f = GParameters (Rep f)
  flattenParameters :: f -> HList (Parameters f)
  default flattenParameters ::
    (Generic f, GParameterized (Rep f), Parameters f ~ GParameters (Rep f)) =>
    f ->
    HList (Parameters f)
  flattenParameters f = gFlattenParameters (from f)
  replaceParameters :: f -> HList (Parameters f) -> f
  default replaceParameters ::
    (Generic f, GParameterized (Rep f), Parameters f ~ GParameters (Rep f)) =>
    f ->
    HList (Parameters f) ->
    f
  replaceParameters f as = to (gReplaceParameters (from f) as)
```
The old one was:
```haskell
class
  Parameterized
    (f :: Type)
    (as :: [Type])
    | f -> as where
  flattenParameters :: f -> HList as
  replaceParameters :: f -> HList as -> f
```
The new one doesn't have functional dependencies, has one type parameter less, uses default signatures to delegate implementations to generics, and provides the `Parameters` type family that computes the list of parameters for a given model ADT.

You can use `DeriveAnyClass` with the `deriving` syntax now to ask for a `Parameterized` instance for any model ADT that has a `Generic` instance:
```haskell
{-# LANGUAGE DeriveAnyClass #-}

...

data MyModel = MyModel (Generic, Parameterized)
```
If this compiles, you can use `Parameters MyModel` to ask GHC for the list of parameters of your model. Within ghci, this would be:
```
> :kind! Parameters MyModel
'[]
```
or something to that effect. (This example has no parameters, therefore the empty type-level list.)


tscholak:
hmm, doctests are failing with stack.

could be this issue:
```
/home/runner/work/hasktorch/hasktorch/hasktorch/src/Torch/Typed/NN/Transformer.hs:177:10: error:
    Illegal `..' notation for constructor ‘MultiheadAttentionSpec’
      The constructor has no labelled fields
    |
177 |   sample MultiheadAttentionSpec {..} =
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

junjihashimoto:
Very nice. Thank you very much.

There are newtype Parameter and type class Parameters, one is like a value, and the other is a function, so I feel that we need to be careful when coding. Both of them are type level, so they start with capital letters, so I feel like I can't help it.

junjihashimoto (approved):

PR #453, opened 2020-09-03
Add performGC in runStep

junjihashimoto:
PerformGC is executed every time in runStep.
performGC is blocking.(When it is non-moving-gc, performGC is not blocking.) It is useful to detect memory leaks.


tycho01 (approved):

tscholak:
I’m putting performGC in all my training loops, but with little to no effect :/

junjihashimoto:
Thank you for your feedback.
It is good there are no side effects.
There may be a leak elsewhere.

PR #456, opened 2020-09-06
Add pure functional dropout

junjihashimoto:
This PR provides pure functional dropout.
dropout implementation is here.
When training, it uses _fused_dropout. _fused_dropout can use local generater.
https://github.com/pytorch/pytorch/blob/1.6/aten/src/ATen/native/Dropout.cpp
https://github.com/pytorch/pytorch/blob/1.6/aten/src/ATen/native/native_functions.yaml#L167

### DIFF
--- a/hasktorch/src/Torch/Random.hs
+++ b/hasktorch/src/Torch/Random.hs
@@ -13,6 +13,7 @@ module Torch.Random
   , randint'
   , normal
   , normal'
+  , dropout
   ) where
 
 import Torch.Internal.Cast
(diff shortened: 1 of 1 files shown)


hasktorch/src/Torch/Random.hs
```diff
@@ -105,3 +131,12 @@ normal mean std size opts gen = generatorFactory (LibTorch.normal_ddlGo (realToF
...
 normal' :: Double -> Double -> [Int] -> Generator -> (Tensor,Generator)
 normal' mean std size gen = normal mean std size defaultOpts gen
+
+-- | Pure functional dropout
+dropout
+  :: Double -- ^ dropout probability
+  -> Maybe Generator -- ^ generator
+  -> Tensor -- ^ input
+  -> (Tensor,Generator) -- ^ output
+dropout p Nothing input = cast3 ATen.dropout_tdb input p False
```
tscholak, line 141: Hi @junjihashimoto, this looks great, thank you for working on this.
This line here is odd, though. I don't understand how this can return a `Generator` when `Nothing` is passed into it. Isn't `ATen.dropout_tdb` the one that returns `IO Tensor`?

junjihashimoto: Thank you for your pointing it out. If the input generator is Nothing, the output generator should also return Nothing or just a Tensor(not the Tuple).

Issue #457, opened 2020-09-07
Turning randomness on and off at the type level


mcwitt:
#### Problem

Currently the `HasForward` typeclass splits the implementation of the forward pass into 2 functions:
- `forward :: f -> a -> b`, for a  forward pass that is a pure function, and
- `forwardStoch :: f -> a -> IO b`, for a forward pass with randomness ("stochastic", e.g. dropout)

For most models, these functions contain the same information, meaning that one can be easily inferred from the other. For example, we can go from `forward` to `forwardStoch` by using `pure`, and from `forwardStoch` to `forward` using `unsafePerformIO`:

```haskell
forwardStoch = (pure .) . forward
forward = (unsafePerformIO .) . forwardStoch
```

For theses cases, half of the implementation is mechanical and it seems natural that we could infer it from a type-level encoding.

#### Example
Consider a 2-layer net with dropout:

``` haskell
instance
  HasForward
    (TwoLayerDropout numIn numOut numHidden)
    (Tensor Device DType '[batchSize, numIn])
    (Tensor Device DType '[batchSize, numOut])
  where
  forward (TwoLayerDropout linear1 linear2) =
    forward linear2
      . forward (Dropout 0.2)
      . tanh
      . forward linear1
  forwardStoch (TwoLayerDropout linear1 linear2) =
    forwardStoch linear2
      <=< forwardStoch (Dropout 0.2)
      <=< pure . tanh
      <=< forwardStoch linear1
```

Notice how we can go from `forward` to `forwardStoch` mechanically by composing pure layers with `pure`, and changing the call from `forward` to `forwardStoch` for stochastic layers.

#### Proposed solutions

Summarizing some ideas from discussion with @tscholak and @junjihashimoto so far. We haven't landed on a concrete design just yet.

1. Define a new monad (e.g. `Stoch`) parameterized by a type-level boolean flag to indicate whether randomness should be turned on or off. The definition could look something like
    ``` haskell
    newtype Stoch (randomnessFlag :: Bool) a = Stoch { runStoch :: IO a }
    ```
    The `HasForward` class could be modified to have only a single `forward` like
    ``` haskell
    forward :: f -> a -> Stoch b
    ```

2. Avoid monadic context; use pure functions with explicit generator state `g`, e.g.
    ``` haskell
    forward :: f -> a -> g -> (b, g)
    ```
    This would give users the option of keeping the random state explicit, or making it implicit with a state monad.

#### Other examples
[...]

PR #459, opened 2020-09-13
Forward unification

mcwitt:
Starting to implement suggestions in #457


hasktorch/src/Torch/NN.hs
```diff
@@ -41,73 +41,62 @@ nextParameter = do
...
-  forwardStoch f a = to <$> gForwardStoch (from f) (from a)
+    B f a
+  forward f a = gForward (from f) (from a)
 
-class GHasForward (f :: Type -> Type) (a :: Type -> Type) (b :: Type -> Type) | f a -> b where
-  gForward :: forall c c' c''. f c -> a c' -> b c''
-  gForwardStoch :: forall c c' c''. f c -> a c' -> IO (b c)
+class GHasForward (f :: Type -> Type) (a :: Type -> Type) where
+  type GB f a :: Type
+  gForward :: forall c c'. f c -> a c' -> GB f a
```
tscholak, line 62: Idea: 
```suggestion
  gForward :: forall c c'. f c -> a c' -> Rep (GB f a)
```
Have you tried that?
I think that way you wouldn't need to work with `()`, `(,)`, and `Either` below.

tscholak (changes requested):
Hi @mcwitt, looks good, thanks! :)
I've got an idea for an improvement, though, do you want to see if you can make it work?
Cheers, Torsten

hasktorch/src/Torch/NN.hs
mcwitt: Thanks, I will try it! I think this is why I was trying to make `GB f a :: Type -> Type` before; I hadn't thought of this way. Still barely understand Generics, but slowly improving :smiley:

tscholak: I have been thinking about this some more in the last half an hour, and I'm not so sure any more about my suggestion.
The issue seems to be that the output type is only defined modulo isomorphism.
I am actually not sure why the compiler accepted the fundeps in `master`, because the output type doesn't seem unambiguously defined jointly by `f` and `a`. I suspect that it will complains once one wants to define instances.

I thus think your draft solution is fine. You've got to commit to specific output types, and these may as well be `()`, `(,)`, and `Either`.

mcwitt: When I do it this way I'm getting confused about how to implement the type family for e.g. the trivial instance:
```haskell
instance GHasForward U1 U1 where
  type GB U1 U1 = _  -- what goes here?
  gForward U1 U1 = U1
```

tscholak:
```Haskell
type GB U1 U1 = _  -- what goes here?
```
Yes, exactly, I realized that, too. You could write `()` of course but that wouldn't improve things.

tscholak: @mcwitt It's late, and I need to sleep over this, but I think we should rethink all of these generic instances. It's probably better to start again from what we are trying to accomplish here.
One thing I came up with just now is Generic1, that is, generics of type constructors...
We could then assume that the type constructor for model, input, and output is the same...

tscholak: Ok, I'm more awake now, let's try to go back to what we want to achieve here:

I'm envisioning automatic derivation capabilities for the following situations:

1) sum types:
```haskell
data Model ~ Either ModelA ModelB  -- using '~' here to indicate that this can be any ADT isomorphic to 'Either'
data In ~ Either InA InB
data Out ~ Either OutA OutB

instance
  (HasForward ModelA InA, OutA ~ B ModelA InA, HasForward ModelB InB, OutB ~ B ModelB InB) =>
  HasForward Model In where
  type B Model In = Maybe Out
  forward (Left modelA) (Left inA) = Just . Left $ forward modelA inA
  forward (Right modelB) (Right inB) = Just . Right $ forward modelB inB
  forward _ _ = Nothing
```
To achieve this, I think we should not only demand that `In` and `Out` are isomorphic, but also that they are the same type constructors. That is, if `In a b` is `Either a b`, then `Out a b` must also be `Either a b`. If `In a b` is
```haskell
data In a b = This { unThis :: a } | That { unThat :: b }
```
then
```haskell
type Out a b = In a b
```
Note that `Model` can be any other ADT that is isomorphic with `Either`. It could be `Either ModelA ModelB` or it could be something else like
```
data OneOfTwoModels a b = ThisModel { unThisModel :: a } | ThatModel {unThatModel :: b }
```
but it mustn't be
```
data OneOfThreeModels a b c = ThisModel { unThisModel :: a } | ThatModel {unThatModel :: b } | TheirModel {unTheirModel :: c }
```

2) product types:
```haskell
data Model ~ (ModelA, ModelB)  -- using '~' here to indicate that this can be any ADT isomorphic to '(,)'
data In ~ (InA, InB)
data Out ~ (OutA, OutB)

instance
  (HasForward ModelA InA, OutA ~ B ModelA InA, HasForward ModelB InB, OutB ~ B ModelB InB) =>
  HasForward Model In where
  type B Model In = Out
  forward (modelA, modelB) (inA, inB) = (forward modelA inA, forward modelB inB)
```
Again, to achieve this, I think we should not only demand that `In` and `Out` are isomorphic, but also that they are the same type constructors. That is, if `In a b` is `(a, b)`, then `Out a b` must also be `(a, b)`. If `In a b` is
```haskell
data In a b = Both { this :: a, that :: b }
```
then
```haskell
type Out a b = In a b
```
Similar considerations as above apply to `Model`.

3) composite types (something that has a `Compose f g` instance where `f` is a functor)
We'll see if this is useful, but consider:
```Haskell
data MaybeModel ~ Maybe Model
data MaybeIn ~ Maybe In
data MaybeOut ~ Maybe Out

instance
  (HasForward Model In, Out ~ B Model In) =>
  HasForward MaybeModel MaybeIn where
  type B Model In = MaybeOut
  forward (Just model) (Just in) = Just $ forward model in
  forward _ _ = Nothing
```

4) lesser variations or combinations of the above
Like:
```haskell
data Model ~ Either ModelA ModelB  -- using '~' here to indicate that this can be any ADT isomorphic to 'Either'
data In = ...
data Out = ...

instance
  (HasForward ModelA In, Out ~ B ModelA In, HasForward ModelB In, Out ~ B ModelB In) =>
  HasForward Model In where
  type B Model In = Out
  forward (Left modelA) in = forward modelA in
  forward (Right modelB) in = forward modelB in
```
and so on. There are many such combinations.

I am quite confident that all of this can be achieved with generics, and I think we should look into `Generic1` (https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-Generics.html#g:16), because `Generic1` makes it possible to reason about type constructors like `Either a b`.

tscholak: About non-determinism/stochasticity in this context:

If we have `StochOut = G -> (Out, G)` with `G` some random state ADT, we may need to wire the instances differently. Note also that `(->) a b` doesn't have a `Generic` instance (AFAIK).

Let's just look at product types for now. We will have:
```haskell
data Model ~ (ModelA, ModelB)  -- using '~' here to indicate that this can be any ADT isomorphic to '(,)'
data In ~ (InA, InB)
data StochOutA ~ G -> (OutA, G)
data StochOutA ~ G -> (OutB, G)
data StochOut ~ G -> ((OutA, OutB), G)

instance
  (HasForward ModelA InA, StochOutA ~ B ModelA InA, HasForward ModelB InB, StochOutB ~ B ModelB InB) =>
  HasForward Model In where
  type B Model In = StochOut
  forward (modelA, modelB) (inA, inB) =
    \g -> case (forward modelA inA) g of (outA, g') -> case (forward modelB inB) g' of (outB, g'') -> ((outA, outB), g'')
```

tscholak: Ok, I think I know now how to get the above dispatching working for type class instances. We can use a closed typed family of the following form to compute an auxiliary type tag:
```Haskell
data ModelRandomness = Deterministic | Stochastic

type family ModelRandomnessR (out :: Type) :: ModelRandomness where
  ModelRandomnessR ((->) g (_, g)) = 'Stochastic  -- not sure yet if this pattern match is syntactically correct, we'll see...
  ModelRandomNessR _ = 'Deterministic
```
Once we have this, we can use the `ModelRandomness` type-level flag in the following way (for example):
```Haskell
class HasForwardProduct (modelARandomness :: ModelRandomness) (modelBRandomness :: ModelRandomness) f1 a1 f2 a2 where
  type BProduct f1 a1 f2 a2 :: Type
  forwardProduct :: f1 -> a1 -> f2 -> a2 -> BProduct f1 a1 f2 a2

instance
  (HasForward ModelA InA, OutA ~ B ModelA InA, HasForward ModelB InB, OutB ~ B ModelB inB) =>
  HasForwardProduct 'Deterministic 'Deterministic ModelA InA ModelB inB where
  type BProduct ModelA InA ModelB inB = (OutA, OutB)
  forwardProduct modelA inA modelB inB = (forward modelA inA, forward modelB inB)

instance
  (HasForward ModelA InA, OutA ~ B ModelA InA, HasForward ModelB InB, StochOutB ~ B ModelB inB) =>
  HasForwardProduct 'Stochastic 'Deterministic ModelA InA ModelB inB where
  type BProduct ModelA InA ModelB inB = G -> ((OutA, OutB), G)
  forwardProduct modelA inA modelB inB =
    \g -> case (forward modelA inA) g of (outA, g') -> ((outA, forward modelB inB), g')

...
```
plus an instance for `HasForward` for the product type that dispatches to `HasForwardProduct` by means of `ModelRandomnessR`. Will write an example tomorrow.

tscholak: And here's how the dispatching would work:
```Haskell
instance
  ( ModelRandomnessR OutA ~ modelARandomness,
    ModelRandomnessR OutB ~ modelBRandomness,
    HasForwardProduct modelARandomness modelBRandomness ModelA InA ModelB InB
  ) =>
  HasForward Model In where
  type B Model In = BProduct (ModelRandomnessR OutA) (ModelRandomnessR OutB) ModelA InA ModelB InB
  forward (modelA, modelB) (inA, inB) = forwardProduct (Proxy :: Proxy modelARandomness) (Proxy :: Proxy modelBRandomness) modelA inA modelB inB
```

tscholak: I think this should work, provided that the pattern match for the `->` type constructor in the `ModelRandomnessR` type family compiles.

@mcwitt can you see if this makes sense?

If so, then what remains to be done is to generalize these concrete implementations with generics. That is, instead of having `Either`, `(,)`, `Compose`, etc., we need to work with `:+:`, `:*:`, `:.:` and so on. I also don't know if we need `Generic1` or not, but I suspect we do.

mcwitt: Awesome! I'll have some time this evening to try to work through it, and will let you know if I have any questions.

tscholak: btw, I tried `ModelRandomnessR`, and it actually works:
```Haskell
data ModelRandomness = Deterministic | Stochastic

data G = G

type family ModelRandomnessR (out :: Type) :: ModelRandomness where
  ModelRandomnessR ((->) G (_, G)) = 'Stochastic
  ModelRandomnessR _ = 'Deterministic
```

```Haskell
> :kind! ModelRandomnessR (G -> (Tensor, G))
ModelRandomnessR (G -> (Tensor, G)) :: ModelRandomness
= 'Stochastic
> :kind! ModelRandomnessR Tensor
ModelRandomnessR Tensor :: ModelRandomness
= 'Deterministic
> :kind! ModelRandomnessR (Int -> (Tensor, Int))
ModelRandomnessR (Int -> (Tensor, Int)) :: ModelRandomness
= 'Deterministic
```

mcwitt:
@tscholak, thanks for writing out your thoughts in detail, this is extremely helpful. Following your suggestion I'm planning to start implementing instances for concrete sums and products.

I got a start on this, but it's been slow going as I still have a lot to learn about working at the type level. I ran into a couple of issues while trying to implement your `HasForwardProduct` class, which I'll mention here to keep a log of my progress.

1. In the examples you use TitleCased for things like `ModelA`, `InA`, etc., but I think these are type variables and therefore should start lowercase (`modelA`, `inA`, etc.)? (if I'm wrong about this one, I must pretty confused indeed :sweat_smile:)

2. Changing type variables to lowercase, I get:
    > • Type variables ‘outA’, ‘outB’ are mentioned in the RHS,
    but not bound on the LHS of the family instance
• In the type instance declaration for ‘BProduct’

    I got this to compile by eliminating the constraint `outA ~ B modelA inA` and just replacing `outA` with `B modelA inA` in the RHS of the type family definition:
    ```diff
     instance
    -  (HasForward modelA inA, outA ~ B modelA inA, HasForward modelB inB, outB ~ B modelB inB) =>
    +  (HasForward modelA inA, HasForward modelB inB) =>
   HasForwardProduct 'Deterministic 'Deterministic modelA inA modelB inB where
    -  type BProduct 'Deterministic 'Deterministic ModelA InA ModelB InB = (OutA, OutB)
    +  type BProduct 'Deterministic 'Deterministic ModelA InA ModelB InB = (B modelA inA, B modelB inB)
       forwardProduct _ _ modelA inA modelB inB = (forward modelA inA, forward modelB inB)
    ```

3. Where I'm currently stuck: implementing the mixed case `HasForwardProduct 'Stochastic 'Deterministic`, I am unsure of the syntax for a polymorphic function type on the RHS of a type family definition. E.g. the compiler complains that the type variable `g` is out of scope in this definition:
    ```haskell
      type BProduct 'Deterministic 'Deterministic modelA inA modelB inB = g -> ((B modelA inA, B modelB inB), g)
    ```
    Changing the RHS to `forall g. g -> ((B modelA inA, B modelB inB), g)` doesn't help either, with ghc complaining
    > • Illegal polymorphic type:
    forall g. g -> ((B modelA inA, B modelB inB), g)
• In the type instance declaration for ‘BProduct’

tscholak:
Hi @mcwitt, I'm glad that the stuff I wrote is helpful to you. I think it's amazing that you are trying so hard to make this work, thanks for that!

The three things you mentioned aren't that bad, I went through similar things when I worked last on this. Let me tell you a few things about each one of them:

1. Yes, you should use lower-case type variables. The convention in Haskell is upper case for type constructors and lower-case for (type) variables. In the examples I wrote I assumed we had a set of concrete model, input, and output types to work with so that it is more clear when the generic instances should fire. In the library, as you found, it of course makes only sense to leave these variable.

2. Yes, this is a limitation of type families. The workaround is exactly as you described. Very good!

3. Whereas I elected to choose concrete types for the model, I for some reason settled on a type variable for the random state. That was a bad choice, because we should have a concrete type constructor for that already somewhere. You can move on by using a skolem type, `data G` (no right-hand side), to push this aside and move on :)

tscholak:
Once `HasForwardProduct` compiles, try to see if you can make a special `HasForwardSum` class, too :)

hasktorch/src/Torch/NN.hs
```diff
@@ -41,73 +43,114 @@ nextParameter = do
...
+  forwardSum' _ _ (Left (modelA, inA)) = \g -> (Left $ forward modelA inA, g)
+  forwardSum' _ _ (Right (modelB, inB)) = \g -> let (outA, g') = forward modelB inB g in (Right outA, g')
+
+--
+-- Fully-stochastic instances
+--
+
+instance (HasForward modelA inA, HasForward modelB inB) => HasForwardProduct 'Stochastic 'Stochastic modelA inA modelB inB where
+  type BProduct 'Stochastic 'Stochastic modelA inA modelB inB = G -> ((B modelA inA, B modelB inB), G)
+  forwardProduct _ _ modelA inA modelB inB = do
```
tscholak, line 134: It's fine to use the state monad internally, but I don't think the types line up without calling runState before returning

hasktorch/src/Torch/NN.hs
```diff
@@ -41,73 +43,114 @@ nextParameter = do
...
+data ModelRandomness = Deterministic | Stochastic
 
-instance
-  (GHasForward f a b) =>
-  GHasForward (M1 i t f) (M1 i t' a) (M1 i t' b)
-  where
-  gForward (M1 f) (M1 a) = M1 $ gForward f a
-  gForwardStoch (M1 f) (M1 a) = M1 <$> gForwardStoch f a
+type family ModelRandomnessR (out :: Type) :: ModelRandomness where
+  ModelRandomnessR ((->) g (_, g)) = 'Stochastic
```
tscholak, line 53: Until the dust settles and we've figured out what random generator state to use, you may want to use the concrete `G` data type here, too. Otherwise, this will match also on, say, `Int -> (_, Int)`.

To make this even better, we could write another type family that deconstructs the passed in type constructor by constructor and checks that `G` doesn't appear any other way than the above in it. If it does, we return a `TypeError`, and GHC will tell the user they this ain't good 😊

mcwitt: Added a check in 04b4175 (it can probably be improved!)

tscholak:
Hey @mcwitt, I saw your last commit, https://github.com/hasktorch/hasktorch/pull/459/commits/5b761410e6e015a5b4b7481e1f6488dd9fcf6fcc.
Looks great!
Do you want to try something cool?
It's actually possible to recursively unpack a data type.
we could actually make a type family that tests whether or not a type, e.g. `G`, is present at any leave of a nested type constructor:
```Haskell
type family Contains (f :: k) (a :: Type) :: Constraint where
  Contains a a = ()
  Contains (f a) a = ()
  Contains (f g) a = (Contains f a, Contains g a)
  Contains _ a = TypeError (Text "Type doesn't contain " :<>: ShowType a :<>: ".")
```
(I just hacked this together, maybe it's not entirely correct)
Can you see if you can make a type family `ContainsNot f a` that returns a `TypeError` if `a` does not appear in `f`?
If you had that, you could use it to make sure that `G` doesn't appear in any other way than as `G -> (f, G)`.

mcwitt:
> It's actually possible to recursively unpack a data type.

@tscholak ah, I see what you were getting at now. This is a really cool idea, thanks for the tip!

tscholak:
Yes, your last edit is exactly in line with what I had thought of.

This is of course just a suggestion.

Let me know what you want to look at next, I think we can turn to Generics again soon.

hasktorch/src/Torch/NN.hs
```diff
@@ -54,15 +56,31 @@ data ModelRandomness = Deterministic | Stochastic
...
+          ( Text "For stochastic models, 'b' must not contain 'Generator' in "
+              :<>: Text "'forward :: f -> a -> Generator -> (b, Generator)'."
+          )
       )
+  CheckStochasticOutType _ =
+    ( TypeError
+        ( Text "Stochastic models must have a forward pass of the form "
+            :<>: Text "'forward :: f -> a -> Generator -> (b, Generator)'."
+        )
+    )
```
mcwitt, line 80: @tscholak I wonder if you know if there's a way to use `Contains` here as a constraint, like in your original example, instead of a type family returning `Bool`. E.g. something analogous to a pattern guard, like
```haskell
type family CheckStochasticOutType (f :: k) :: ModelRandomness where
  CheckStochasticOutType (b, G) | Not (Contains b G) = -- ...
```
would be really cool (but this doesn't work, and I haven't found anything analogous yet).

tscholak: hi @mcwitt, yeah, guards would be cool, but that doesn't work.

I think you may just need the `Bool` kind:
```
type family Contains (f :: k) (a :: Type) :: Bool where
  Contains a a = 'True
  Contains (f a) a = 'True
  Contains (f g) a = If (Contains f a) 'True (Contains g a)
  Contains _ a = 'False

type family ModelRandomnessR (out :: Type) :: ModelRandomness where
  ModelRandomnessR ((->) G (a, G)) = If (Contains a G) (TypeError ...) 'Stochastic
  ModelRandomnessR a = If (Contains a G) (TypeError ...) 'Deterministic
```

tscholak: btw, did you find `||` and `If` in http://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Type-Bool.html or did you make it yourself? 
I think there's an `If` in `Torch.Typed.Aux` that we should probably remove

mcwitt: > btw, did you find Not, ||, and If

Yep, these are from `Data.Type.Bool` (I just found out about this, and remember using `If` from `Torch.Typed.Aux` in a previous PR.) I'll make an issue to remove from `Torch.Typed.Aux`.

mcwitt:
The next thing I'm wanting to add is a `HasForward` instance that dispatches to `HasForwardProduct`, like https://github.com/hasktorch/hasktorch/pull/459#discussion_r489446749. But while I'd like to write the analogous thing with type variables instead of concrete types, i.e.
``` haskell
instance
  ( modelARandomness ~ ModelRandomnessR outA,
    modelBRandomness ~ ModelRandomnessR outB,
    HasForwardProduct modelARandomness modelBRandomness modelA inA modelB inB
  ) =>
  HasForward model input
  where
  type B model input = BProduct (ModelRandomnessR outA) (ModelRandomnessR outB) modelA inA modelB inB
  forward (modelA, modelB) (inA, inB) = forwardProduct (Proxy :: Proxy modelARandomness) (Proxy :: Proxy modelBRandomness) modelA inA modelB inB
```
this unfortunately doesn't work, with GHC complaining that `in{A,B}`, `out{A,B}`, `model{A,B}` are unbound on the LHS of the type family instance. I solved a similar issue above (https://github.com/hasktorch/hasktorch/pull/459#issuecomment-693983200), but in this case we don't have much to work with on the LHS (just `model`, `input`).

I'm not sure how to proceed at the moment, but I'll keep thinking about this. Maybe the way to get this working with general types is through Generics?

tscholak:
Hi @mcwitt! yes, this seems to be a bit problematic.

I haven't tried this, but could you try this:
```
instance
  ( modelARandomness ~ ModelRandomnessR outA,
    modelBRandomness ~ ModelRandomnessR outB,
    HasForwardProduct modelARandomness modelBRandomness modelA inA modelB inB
  ) =>
  HasForward (modelA, modelB) (inA, inB)
  where
  type B (modelA, modelB) (inA, inB) = BProduct (ModelRandomnessR outA) (ModelRandomnessR outB) modelA inA modelB inB
  forward (modelA, modelB) (inA, inB) = forwardProduct (Proxy :: Proxy modelARandomness) (Proxy :: Proxy modelBRandomness) modelA inA modelB inB
```
or something to this effect? otherwise, there's really nothing that tells the compiler how `model` and `model{A,B}` are related.

mcwitt:
@tscholak thanks for the suggestion; I tried it and it seemed to work (modulo needing to enable `ScopedTypeVariables`, but maybe there's a way to do it without).

While doing this, I realized I had forgotten to update the signature for `forward` in `HasForward` to `f -> a -> B f a`. Fixing this led me to realize there is a problem with the way I defined the stochastic `HasForwardProduct` and `HasForwardSum` instances. I think something like c710d18 is needed to avoid the following

>    • Couldn't match expected type ‘G -> (B modelB inB, G)’
                  with actual type ‘B modelB inB’
    • The function ‘forward’ is applied to three arguments,
      but its type ‘modelB -> inB -> B modelB inB’ has only two
      In the expression: forward modelB inB g
      In a pattern binding: (outB, g') = forward modelB inB g

Unfortunately, this doesn't work and generates the following error

> • Expected kind ‘(*, b0)’, but ‘B modelA inA’ has kind ‘*’  
• In the first argument of ‘Fst’, namely ‘(B modelA inA)’   
  In the type ‘G -> ((Fst (B modelA inA), B modelB inB), G)’
  In the type instance declaration for ‘BProduct’ (lsp)     

It makes sense that this is not obviously a safe thing to do; although we know that if `modelARandomness ~ 'Stochastic`, then `outA ~ G -> (b, G)`, but of course GHC does not know this. This feels similar to problems in normal term-level programming that are solved by passing "evidence parameters" rather than booleans. I wonder if something like that could apply here? (Of course, I may still just be missing something obvious).

Mainly writing this to let you know I'm still thinking about this, but would appreciate your thoughts if you get a chance!

Mostly just writing this to let you know I'm still working on this, but would also appreciate any suggestions if you have them.

tscholak:
Hey @mcwitt. I fumbled around last night and this morning, and I think I figured it out.
I made a draft of an initial gradually typed hasktorch API, and I put the exploration in there because that was the most convenient thing to do. You should be able to use it with only small modifications:

https://github.com/hasktorch/hasktorch/blob/208ddbf654f94aabb53862e0674dd5b09d345b28/hasktorch/src/Torch/GraduallyTyped/NN.hs#L84-L111

tscholak:
semi-relevant curiosity: https://twitter.com/iceland_jack/status/1310689993397096448?s=21

mcwitt:
Thanks for getting me unstuck! The solution isn't so bad. Signing off for tonight; next I will tackle `HasForward` instances and return to generics.

tscholak:
Thanks @mcwitt, the changes look great!

hasktorch/src/Torch/NN.hs
```diff
@@ -41,73 +46,170 @@ nextParameter = do
...
+    a2 ->
+    Proxy b2 ->
+    BProduct modelARandomness modelBRandomness f1 a1 b1 f2 a2 b2
+
+class HasForwardSum (modelARandomness :: ModelRandomness) (modelBRandomness :: ModelRandomness) f1 a1 b1 f2 a2 b2 where
+  type BSum modelARandomness modelBRandomness f1 a1 b1 f2 a2 b2 :: Type
+  forwardSum ::
+    Proxy modelARandomness ->
+    Proxy modelBRandomness ->
+    Either (f1, a1) (f2, a2) ->
```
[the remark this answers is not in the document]
mcwitt: Thanks, yes, I see that now. I wasn't seeing the full picture, and this way seemed nicer in isolation.

tscholak:
nicely done @mcwitt !

tscholak:
How is it going? It looks like you're a bit stuck

mcwitt:
> How is it going? It looks like you're a bit stuck

@tscholak yeah, where I left off I realized I needed a better understanding of Generics (especially `Generics1`) to make progress. I've been reading up on this. BTW, finding Andreas Loh's talk from the last ZuriHac extremely helpful (both the talk and lecture notes in the [repo](https://github.com/well-typed/gp-zurihac-2020)).

I'm planning to work on it more this weekend and probably make more slow incremental progress from there. In the meantime, absolutely no hard feelings if it makes sense to subsume some or all of this work in your gradually-typed work. I'm happy to collaborate with you there as well :smiley:

tscholak:
I also found @kosmikus ' Zurihac talk extremely good!
Btw, I looked a bit at the https://github.com/well-typed/generics-sop package, and because of its Sum-of-Products paradigm it may be a better fit for this.

tscholak:
With regards to gradual typing, I'd think what you're doing is largely independent of that.
I only did some early exploration for this PR in that branch, because I didn't want to send you off chasing wild geese.

mcwitt:
Hey @tscholak, I'm getting a little bogged down in this, and think I might be trying to do too much at once for my comfort level. I've been going back to the drawing board so to speak, and had the idea to try to break the changes into two smaller units:

1. **Remove the output type parameter `b` and functional dependency from the `HasForward` type class, replace with associated type family `type B f a :: Type` (leaving `forwardStochastic` in place for the moment).** Here I'm getting stuck on the following: the associated type family `GB` needs to return a concrete type (rather than a representation), because type families are non-injective and we cannot recover the type parameter with something like `type B f a = UnRep (GB (Rep f) (Rep a))`. But then I am not sure how to implement `GB` without choosing an explicit representation in terms of `(,)` or `Either`. I think you had already been thinking about this and had a strategy in mind ([here](https://github.com/hasktorch/hasktorch/pull/459#discussion_r487916090)), but admittedly I'm not sure how we would implement this. Would this involve unwrapping / rewrapping the Generics metadata wrappers? For example, I'm stuck on implementing this: https://github.com/mcwitt/hasktorch/blob/f66aff7d17500d6c4fbcf149626b516971d73dce/hasktorch/src/Torch/NN.hs#L85

2. **Remove `forwardStochastic` in favor of stochastic models having `type B f a = (G -> (Output, G))` and dispatching mechanism described by @tscholak earlier in the PR comments.** Here I had made some progress on the implementation for concrete types, following your examples, but got a bit tripped up when trying to extend to Generics. I think clearing up the first question will give me more confidence moving forward.

Would it make sense to split these into 2 separate PRs?

tscholak:
Hi @mcwitt!
The progress you have made is fantastic! Thanks so much!
The problems you describe are indeed familiar to me, and I think it would make sense to split the work up along those lines.
However, I would suggest a different split. I would prefer if you kept the work on stochastic vs not stochastic but removed or commented all the Generic bits instead.
That’s because what you have without the Generics is actually usable and provides enough insight for others to define their own HasForward instances. Compared to that, the generalization to isomorphic ADTs is really just a cherry on top, because it only saves some boilerplate.
What do you think?

mcwitt:
@tscholak I think this makes sense to me. Because the amount of stuff in this draft is getting pretty overwhelming, I'm thinking to close this and open a series of more focused PRs, starting with the elimination of `forwardStoch` as you suggest.

Here's the first one: https://github.com/hasktorch/hasktorch/pull/477

PR #463, opened 2020-09-20
Improve getting started doc

sami-badawi:
Improve Nix getting started documentation to include build and running unit and a mnist example


austinvhuang:
Thanks!

PR #466, opened 2020-09-27
Stack nix documentation

sami-badawi:
Added more specific documentation about using Stack to build in nix-shell.
Also took cpu as the default device for examples static-mnist-cnn and static-mnist-mlp so they don't fail when called without arguments.


tscholak (approved):
@sami-badawi looks great, thank you for helping!

PR #469, opened 2020-09-30
Windows documentation

sami-badawi:
Documents how to run Hasktorch under Windows Subsystem for Linux 2 and a few troubleshooting tips

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,16 @@ You can launch a Nix shell via
 $ nix-shell
 ```
 
+We also support Stack with Nix, see below.
+
+Note that this shell is configured to use the CPU backend only.
+In order to benefit from any CUDA-capable hardware acceleration your computer may provide (sorry macOS users),
+launch the Nix shell instead with:
+
+```sh
+$ nix-shell --arg cudaSupport true --argstr cudaMajorVersion 10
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,16 +131,21 @@ $ cd ..
...
-launch the Nix shell instead with:
+Hasktorch mainly works on Windows Subsystem for Linux 2. 
+Tested on Debian and Ubuntu.
+It is not quite as roubust, but these extra command gets it to work:
 
 ```sh
-$ nix-shell --arg cudaSupport true --argstr cudaMajorVersion 10
+$ nix-shell
+$ cabal update
+$ haskell-language-server
```
tscholak, line 143: is this starting the server as a background process?
will vscode be able to connect to it?
I'm asking because I'm used to vscode (specifically, the Haskell extension) starting the haskell language server.

tscholak:
Thank you @sami-badawi! Looks good, just one question, see above :)

README.md
sami-badawi: No this will run a compile, but it takes a long time and you can find and fix problem this way. Otherwise you are opening VS Code to a sea of red underlines.

Issue #467, opened 2020-09-28
CppOptimizer runStep ignores its learning rate parameter


RichardWarfield:
I'm not sure how to fix this, or whether it should be fixed given the learning rate is provided in the <Optimizer>Options record.

```haskell
  runStep paramState optState lossValue lr = do
    performGC
    unsafeStep paramState optState lossValue
```

junjihashimoto:
The learning rate is included in not optmizer's base class but child class.
https://github.com/pytorch/pytorch/blob/master/torch/csrc/api/include/torch/optim/optimizer.h
https://github.com/pytorch/pytorch/blob/master/torch/csrc/api/include/torch/optim/sgd.h#L26
So I think we should change step-function's definition or use upcast to update lr on unsafeStep.

RichardWarfield:
I was thinking maybe just don't provide an `Optimizer` instance for CppOptimizerState and instead have a standalone `runStep` function (or have it be part of the CppOptimizer class)?  As it stands it seems like an awkward fit since two of the three class members of `Optimizer` are undefined.  Also I've been trying to write a typed version for `CppOptimizer` and this seems like an even more awkward fit since things like the `gradients` type don't show up anywhere.

junjihashimoto:
Sorry for lack of explanation. 
There are 2 class, one is Haskell, another is C++.
Above my using class is c++'s one, and we may be able to add learning rate parameter for runStep function.

You know, Haskell's Optimizer's step and runStep' are not implemented.
The step function is not used except for runStep.

The gradient of `runStep'` can not be implemented, because c++'s step function can use loss only.
https://github.com/pytorch/pytorch/blob/master/torch/csrc/api/include/torch/optim/optimizer.h#L107

Haskell
```
class Optimizer optimizer where
    step :: LearningRate -> Gradients -> [Tensor] -> optimizer -> ([Tensor], optimizer)
    runStep :: (Parameterized model) => model -> optimizer -> Loss -> LearningRate -> IO (model, optimizer)
    runStep' :: (Parameterized model) =>model -> optimizer -> Gradients -> LearningRate -> IO (model, optimizer)

instance {-# OVERLAPS #-} CppOptimizer option => Optim.Optimizer (CppOptimizerState option) where
  step = error  "step is not implemented for CppOptimizer."
  runStep paramState optState lossValue lr = do
    performGC
    unsafeStep paramState optState lossValue
  runStep' = error  "runStep' is not implemented for CppOptimizer."
```

[...]

PR #471, opened 2020-10-06
Add Torch.Typed.Optim.CppOptim and associated example

RichardWarfield:
A typed version of Torch.Optim.CppOptim.  I have also added a partial reimplementation of the `optimizers-cpp` example under `optimizers-cpp-typed` to demonstrate usage.

### DIFF
--- a/hasktorch/src/Torch/Typed/Optim/CppOptim.hs
+++ b/hasktorch/src/Torch/Typed/Optim/CppOptim.hs
@@ -1,0 +1,141 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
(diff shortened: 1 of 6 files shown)


tscholak:
Awesome @RichardWarfield, thank you very much! Looking forward to reviewing it!

junjihashimoto:
LGTM!

junjihashimoto (dismissed):
Thx!

hasktorch/src/Torch/Typed/Optim/CppOptim.hs
```diff
@@ -0,0 +1,141 @@
...
+      initParams'= hmap' ToDependent $ flattenParameters model
+
+instance CppOptimizer SGDOptions where
+  initOptimizer  opt@SGDOptions{..} model = do
+    v <- cast6 LibTorch.sgd sgdLr sgdMomentum sgdDampening sgdWeightDecay sgdNesterov initParams'
+    return $ CppOptimizerState opt v
+    where
+      initParams'= hmap' ToDependent $ flattenParameters model
+
+runStep :: (CppOptimizer option, Parameterized model
```
junjihashimoto, line 130: This typed runStep is not the same as Torch.Typed.Optim's one.
Can we use type class for runStep?
If there are any problems on design, there is no need to use it.

junjihashimoto: If it doesn't seem to be easy to fix, I think it's okay to merge this pr.

RichardWarfield: The only difference should be the learningRate parameter which I excluded for the reasons discussed in #467.  The learning rate is part of the `options` field of the `optim` argument already.

tscholak (approved):
I agree with @junjihashimoto.
I see this as an incremental milestone. I think we will figure out better abstractions in the future.
Thank you very much @RichardWarfield!

PR #473, opened 2020-10-09
bump dependencies

tscholak:
@mcwitt I also added hls 0.5 :)


tscholak:
jupyterlab is broken. it needs additional work, anyway. how about we remove this check for now?

tscholak:
closed in favour of #474

PR #474, opened 2020-10-09
GHC 8.10 with nix again

tscholak:


junjihashimoto:
Thx!

junjihashimoto:
How about not upgrading stack of macos for the time being? Because it seems to take time.

junjihashimoto:
Thx! macos using stack will be fixed later.

PR #477, opened 2020-10-22
Forward unification 1: add output type family, HasForward instances

mcwitt:
Breaking up https://github.com/hasktorch/hasktorch/pull/459 into a series of more manageable chunks.

In this PR the goals are to:

1. Update `HasForward`, replacing output type parameter `b` with a type family and removing `fowardStoch`. Stochastic vs. deterministic will now be determined by the associated type family for each instance, with types of the form `Generator -> (a, Generator)` stochastic.

2. Add `HasForwardProduct`, `HasForwardSum` instances for products and sums. Eventually (not in this PR), we can dispatch to `forwardProduct` and `forwardSum` via a generic implementation of `forward`.

3. Get everything to compile! This means updating all existing `HasForward` instances.

### `HasForward` instance progress

- [x] Linear (but this is ugly; better to introduce intermediate type family?)
- [x] Normalization
- [x] Dropout
- [x] Sparse
- [ ] HasForward (stuck on this; `outputSize` must appear on LHS of type family def; now that I think about it, can't the output type in general depend on the expected type in the context?)
- [ ] DataParallel
- etc.


hasktorch/src/Torch/Typed/NN/Convolution.hs
```diff
@@ -84,10 +91,16 @@ instance
...
-  HasForward (Conv1d inputChannelSize outputChannelSize kernelSize dtype device) (Tensor device dtype '[batchSize, inputChannelSize, inputSize], Proxy stride, Proxy padding) (Tensor device dtype '[batchSize, outputChannelSize, outputSize])
+  HasForward
+    (Conv1d inputChannelSize outputChannelSize kernelSize dtype device)
+    (Tensor device dtype '[batchSize, inputChannelSize, inputSize], Proxy stride, Proxy padding)
   where
+  type
+    Output
+      (Conv1d inputChannelSize outputChannelSize kernelSize dtype device)
+      (Tensor device dtype '[batchSize, inputChannelSize, inputSize], Proxy stride, Proxy padding) =
+      Tensor device dtype '[batchSize, outputChannelSize, outputSize]
```
mcwitt, line 102: This fails to compile because `outputSize` is unknown here. This worries me a bit that we have cases where the output type is polymorphic and only reified in context; so it's not possible in general to replace the type parameter `b` with a type family. But I might be missing something.

mcwitt: Sorry, that was a confusing comment, and possibly misused some overloaded terminology; my essential point of confusion is this: what happens when the output type (e.g. the dimensions of the output tensor) are polymorphic (and therefore context-dependent)? For example, in `Conv1d` before this PR, `forward` is polymorphic in `inputSize` and `outputSize`, and we use the constraint family `ConvSideCheck` to constrain to valid dimensions. But in this case it seems like there might be many possible output shapes consistent with the input shape (this is where I think I'm wrong... still need to understand `ConvSideCheck`). In any case, it seems like we'd need to rewrite the constraint into a type family from input to output. I'll take a look at this next.

mcwitt: Resolved in Slack conversation! I don't think my concern above is valid; there should not be any cases where `forward` is polymorphic in the output type (e.g. allowing different shapes depending on context)

hasktorch/src/Torch/Typed/NN/Dropout.hs
```diff
@@ -34,9 +35,9 @@ dropoutForward ::
   IO (Tensor device dtype shape)
 dropoutForward Dropout {..} dropoutTrain = dropout dropoutProb dropoutTrain
 
-instance HasForward Dropout (Tensor device dtype shape) (Tensor device dtype shape) where
+instance HasForward Dropout (Tensor device dtype shape) where
+  type Output Dropout (Tensor device dtype shape) = Tensor device dtype shape
   forward dropout input = unsafePerformIO $ dropoutForward dropout False input
```
tscholak, line 40: alright, I guess here we are block because there isn't a dropout implementation yet that supports generators

tscholak:
@mcwitt This looks great!

hasktorch/src/Torch/Typed/NN/Transformer.hs
```diff
@@ -581,10 +581,16 @@ instance
...
+  HasForward
+    (TransformerLM numAttnLayers numHeads ffnDim paddingIdx numEmbeds embedDim dtype device)
+    (Tensor device 'D.Int64 '[batchSize, seqLen])
   where
+  type
+    Output
+      (TransformerLM numAttnLayers numHeads ffnDim paddingIdx numEmbeds embedDim dtype device)
+      (Tensor device 'D.Int64 '[batchSize, seqLen]) =
+      Tensor device dtype '[batchSize, seqLen, numEmbeds]
   forward model input = unsafePerformIO $ transformerLM model False input
```
tscholak, line 593: same here. I hope we can get a dropout soon that takes a generator as argument. otherwise, we can't train these models with dropout.

hasktorch/src/Torch/Typed/NN/Dropout.hs
mcwitt: Yes, makes sense. My initial goal was just to get everything to compile, but makes sense that this is not very useful until we can augment the output type to make it stochastic.

tscholak:
hey, do you want to revisit this for `Torch.GraduallyTyped`?
it would go here: https://github.com/hasktorch/hasktorch/blob/master/experimental/gradually-typed/src/Torch/GraduallyTyped/NN/Class.hs#L45

hasktorch/src/Torch/Typed/NN/Transformer.hs
tscholak: this still isn't resolved :/

mcwitt:
> hey, do you want to revisit this for Torch.GraduallyTyped?
@tscholak yep, I'm pretty excited to try this out!

hasktorch/src/Torch/Typed/NN/Transformer.hs
mcwitt: Yeah, too bad. When I find some time I'm hoping to try to reproduce the test failure in Junji's PR and see if I can make any progress..

PR #479, opened 2020-10-30
Add stack-ghc88.yaml to keep supporting ghc-88

junjihashimoto:
To support ghc-810 for macos, we need to upgrade Cabal package of stack command.
https://github.com/haskell/cabal/issues/6421

Since it takes time to upgrade, we will continue to support ghc88 of macos for the time being.


tscholak (approved):
thank you Junji, this was a team effort! 🙌

junjihashimoto:
Thank you, too.

PR #493, opened 2020-11-25
Add makeIndependent with requires_grad argument

junjihashimoto:
Add a requires_grad argument to make a tensor of requires_grad=False, because running_mean argument of batchNorm function requires a tensor with requires_grad = False.


tscholak (approved):
LGTM.
This seems to be a band-aid to a more fundamental problem with the semantics around `makeIndependent` and how it is implemented in the API. Let's consider alternatives later.

junjihashimoto:
Thx!
Pytorch module has parameter of requires_grad=true and buffer of requires_grad=false,
so It may be good to add a buffer type for untyped api.
https://pytorch.org/docs/stable/generated/torch.nn.Module.html

PR #500, opened 2020-12-16
Add ToDevice typeclass for untyped api

junjihashimoto:
This pr adds ToDevice typeclass to provide a general way to move the data from CPU to CUDA.

### DIFF
--- a/hasktorch/src/Torch/NN.hs
+++ b/hasktorch/src/Torch/NN.hs
@@ -28,6 +28,7 @@ import qualified Torch.Internal.Managed.Type.Tensor as ATen
 import Torch.Scalar
 import Torch.Tensor
 import Torch.TensorFactories (ones', randIO', randnIO', zeros')
+import Torch.Device
 
 type Parameter = IndependentTensor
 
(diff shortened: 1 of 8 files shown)


hasktorch/src/Torch/NN.hs
```diff
@@ -117,40 +118,57 @@ class Parameterized f where
...
         then f'
         else error "Some parameters in a call to replaceParameters haven't been consumed!"
 
+instance Parameterized a => ToDevice a where
+  toDevice = replaceDevice
+
 instance Parameterized Tensor where
   flattenParameters _ = []
   _replaceParameters = return
+  replaceDevice = _toDevice
```
junjihashimoto, line 141: replaceDevice is similar to _replaceParameters.
Tensor type is not a parameter, but it is the target of the move.
Parameter type is a parameter and the target of the move.

tscholak (approved):
LGTM, thanks @junjihashimoto.

I guess I will implement yet another version of this for the gradually typed API...

junjihashimoto:
Thank you for your reviewing!

PR #504, opened 2020-12-19
Introduce alternative version of `stack` using `vector-sized`

wz1000:
Using `vecStack` instead of `stack` cut GHC memory usage and compile time by > 80% for my application.

### DIFF
--- a/hasktorch/hasktorch.cabal
+++ b/hasktorch/hasktorch.cabal
@@ -126,6 +126,7 @@ library
                     , data-default-class
                     , containers
                     , inline-c
+                    , vector-sized
 
  default-extensions:  Strict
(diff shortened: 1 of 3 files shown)


tscholak (approved):
Cool, thanks 🙏

junjihashimoto:
Is the improvement at compile time or at runtime? Or both?

wz1000:
> Is the improvement at compile time or at runtime? Or both?

Mostly compile time, though there might be a minor runtime improvement depending on how you use `stack`. I suspect this to be largely dwarfed by the cost of actually running/training the model in most cases.

junjihashimoto:
Thx!

PR #505, opened 2020-12-20
Add replaceTensor to swap floating point type

junjihashimoto:
This PR enables training in half type.
Half type is only supported on cuda.

### DIFF
--- a/hasktorch/src/Torch/NN.hs
+++ b/hasktorch/src/Torch/NN.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DefaultSignatures #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
@@ -10,7 +11,6 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
-{-# LANGUAGE ConstraintKinds #-}
 
 module Torch.NN where
 
@@ -21,8 +21,8 @@ import Data.Kind
 import GHC.Generics
 import System.IO.Unsafe (unsafePerformIO)
 import Torch.Autograd
-import Torch.Device
 import Torch.DType
+import Torch.Device
 import Torch.Functional
 import Torch.Initializers
 import Torch.Internal.Cast (cast3)
@@ -42,16 +42,20 @@ type Parameterized = GTraversable Parameter
 instance GTraversable Parameter Bool where
   gflatten _ = []
   gupdate = pure
+
 instance GTraversable Parameter Int where
   gflatten _ = []
   gupdate = pure
+
 instance GTraversable Parameter Float where
   gflatten _ = []
   gupdate = pure
+
 instance GTraversable Parameter Double where
   gflatten _ = []
   gupdate = pure
-instance GTraversable Parameter (a->a) where
+
+instance GTraversable Parameter (a -> a) where
   gflatten _ = []
   gupdate = pure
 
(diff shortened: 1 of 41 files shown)


hasktorch/src/Torch/NN.hs
```diff
@@ -118,12 +119,12 @@ class Parameterized f where
   default _replaceParameters :: (Generic f, GParameterized (Rep f)) => f -> ParamStream f
   _replaceParameters f = to <$> _gReplaceParameters (from f)
 
-  replaceDevice :: Device -> f -> f
-  default replaceDevice :: (Generic f, GParameterized (Rep f)) => Device -> f -> f
-  replaceDevice dev f = to $ gReplaceDevice dev (from f)
+  replaceTensor :: (Tensor -> Tensor) -> f -> f
```
tscholak, line 122: this is a nice idea. can you make this dependent on the generic representation if `f`, please?
this way, one could use that to do more intentional model weight surgery.

junjihashimoto: I think there is a demand to match and replace specific data, but I'm not sure how to make it without creating a custom class and instance for each application. I wish there was something like try-function for to-function of ghc.generic.

junjihashimoto: We can do it with template haskell, but it's not what we expect.

junjihashimoto: Oh, I misunderstood it.

junjihashimoto: I misunderstood a little, please wait a moment.

junjihashimoto:
@tscholak I've added GTraversal type class to support for generic replacement.
We will need to change Paramerized instances.

hasktorch/src/Torch/Traversable.hs
```diff
@@ -0,0 +1,117 @@
...
+  gflatten' (M1 x) = gflatten' x
+  gupdate' (M1 x) = do
+    x' <- gupdate' x
+    return $ M1 x'
+
+instance GTraversable a a where
+  gflatten = pure
+  gupdate _ = gpop
+
+instance {-# OVERLAPS #-} (Generic f, GGTraversable a (Rep f)) => GTraversable a f
```
tscholak, line 68: I'm not sure about this. How is the user experience if
1. No Generic instance exists  for a given f?
2. No Generic instance can be automatically derived for a given f?
3. A user wishes to implement a custom GTraversable instance for a given f?

junjihashimoto: > 1. No Generic instance exists for a given f?

f can become a GTraversable instance without a Generic instance.

> 2. No Generic instance can be automatically derived for a given f?

We can use a custom Generic instance for a given f.

> 3. A user wishes to implement a custom GTraversable instance for a given f?

A user can implement a custom GTraversable instance of f without a Generic instance.
A user can not implement a custom GTraversable instance of f with a Generic instance.

Oh, we can not use both a custom Generic instance and a custom GTraversable instance for a given f.

junjihashimoto: I will make some examples for the instances.

hasktorch/test/NNSpec.hs
```diff
@@ -16,6 +17,67 @@ import Torch.Autograd
...
+  }
+  deriving (Show, Generic)
+
+instance GTraversable Parameter Customlinear where
+  gflatten Customlinear {..} = [customWeight, customBias, customBias] -- These duplicated biases are for testing!!
+  gupdate _ = do
+    customWeight <- gpop
+    customBias <- gpop
+    _ <- gpop
+    return $ Customlinear {..}
```
junjihashimoto, line 38: > 1. No Generic instance exists for a given f?
> 2. No Generic instance can be automatically derived for a given f?

We can implement a custom GTraversable instance with or without a generic instance.

> 3. A user wishes to implement a custom GTraversable instance for a given f?

I think so.
I think a user wishes to use a generic GTraversable instance for the time being.

junjihashimoto: An additional general instance is not allowed. 
The following code causes  a duplicated instance error.
```
instance {-# OVERLAPS #-} (Scalar f) => GTraversable a f
```

PR #506, opened 2020-12-29
Update niv dependencies

Cmdv:
Running update of the niv dependencies, which should trigger new cachix.

Things still compiling on my end but will report back once done.

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -5,10 +5,10 @@
         "homepage": "",
         "owner": "input-output-hk",
         "repo": "hackage.nix",
-        "rev": "ed53458ea6b3bb1139685659ca6d0d857515d757",
-        "sha256": "0swzynsjp8yyarvz8sq23ras864dqsd323yb0rlwmjxnhqan5b19",
+        "rev": "93e877fa2744aa44d2eed632dbbf051cf5a362d0",
+        "sha256": "1cly319aq0pa199sb6jvvpa0q3w7nxq11c3qc9lv0ms5vdrzjgig",
         "type": "tarball",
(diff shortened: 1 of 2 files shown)


junjihashimoto:
Thx!

junjihashimoto:
It may be killed by the OS for not enough memory.

junjihashimoto:
CI uses two processes. https://github.com/hasktorch/hasktorch/blob/master/.github/workflows/nix-linux.yml#L52

It may be better to use one.

Cmdv:
ah yeah cool I'll change that now see if that's any better. Locally think it's rebuilding GHC 🤦‍♀️

Cmdv:
damn seems to still be failing 😞

nix/sources.json
```diff
@@ -101,10 +101,10 @@
         "homepage": "https://pytorch.org/",
         "owner": "stites",
         "repo": "pytorch-world",
-        "rev": "e40fc2133b169b72eb8d772f98b1484dc1d3903f",
+        "rev": "e08419a2f2b9ca177903c9595858b0dc386fc4dd",
```
tscholak, line 104: @Cmdv what commit is this? are you sure this is a newer version?

junjihashimoto: There is no need to change, but both are almost the same.

Cmdv: I literally did `niv update` and let niv change all the revs

tscholak: ok cool, thanks :)

tscholak (approved):
LGTM, let's see if the CI agrees

tscholak:
closing since #510 has been merged

PR #512, opened 2021-01-19
Dump all libtorch-objects to find objects of memory-leak.

junjihashimoto:
`dumpLibtorchObjects` shows all libtorch-objects older than the argument age.
When it is called, age increases by one.

### DIFF
--- a/examples/mnist-mlp/Main.hs
+++ b/examples/mnist-mlp/Main.hs
@@ -20,6 +20,7 @@ import Torch
 import Torch.Serialize
 import Torch.Typed.Vision (initMnist)
 import Torch.Internal.GC
+import System.Mem (performGC)
 import qualified Torch.Vision as V
 import Prelude hiding (exp)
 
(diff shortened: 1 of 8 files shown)


examples/mnist-mlp/Main.hs
```diff
@@ -91,6 +93,8 @@ main = do
   model <- foldLoop init 5 $ \model _ ->
     runContT (streamFromMap (datasetOpts 2) trainMnist) $ trainLoop localDevice model optimizer . fst
 
+  performGC
+  dumpLibtorchObjects 1
```
austinvhuang, line 97: I would remove the code from the example before merging since this probably confuses new users who are just looking to see an mnist example for the first time.

austinvhuang: if you want to keep this for testing/diagnostics, would copy the example to the experimental directory as its own program.

austinvhuang:
Thanks @junjihashimoto ! I think it's good to merge except we should leave this sort of testing/experimentation outside the examples directory. Could create a separate mnist memory test under `experimental/` if you want to keep it for further testing. Other than that, this looks pretty handy.

examples/mnist-mlp/Main.hs
junjihashimoto: Thank you for your reviewing.
You are right. It should be removed on the example.

junjihashimoto:
Thx, too!
I remove it from the mnist example. 
It's a debugging function, and I don't always try it, so I think it's okay not to put it in experimental.

PR #513, opened 2021-01-20
[WIP] Use jemalloc as the default

junjihashimoto:
To prevent memory-leak of libtorch, this pr sets jemalloc as the default.


tscholak (approved):
LGTM, thanks!

tscholak:
there's an issue with some of the nix builds:
```
/nix/store/i2j8ys3w3ss10r67pvln1dirk21zlm27-jemalloc-5.2.1/lib/libjemalloc.so.2: cannot allocate memory in static TLS block
```

tscholak:
there may be a flag, `--disable-initial-exec-tls`, that provides a fix, see https://github.com/NixOS/nixpkgs/issues/54183

tscholak:
this flag should be passed to the `jemalloc` derivation:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/libraries/jemalloc/common.nix#L9

tscholak:
additional relevant information about this flag: https://github.com/NixOS/nixpkgs/pull/82711#issuecomment-602495037

tscholak:
if we want to try this, we would have to add this little nixpkgs overlay:
```nix
(self: super: {
  jemalloc = pkgs.jemalloc.override {
    disableInitExecTls = true;
  };
})
```

junjihashimoto:
Thank you for your advice.
While building hasktorch-package, the following segmentation fault occurs on the local pc.
```
Building library for hasktorch-0.2.0.0..
/nix/store/frh7ir9rcv19b3ym33ck64s813yr3xrr-stdenv-linux/setup: line 1303:   139 Segmentation fault      $SETUP_HS build lib:hasktorch -j$(($NIX_BUILD_CORES > 4 ? 4 : $NIX_BUILD_CORES))
builder for '/nix/store/cqsr9xvwcwr4val8g9z04rcjf9y0d55c-hasktorch-lib-hasktorch-0.2.0.0.drv' failed with exit code 139
```

junjihashimoto:
I think the order of the links may cause this segmentation fault.
I should run cabal-build to verify the order with `LD_DEBUG=bindings`.

junjihashimoto:
GHC throws the segmentation fault.
I'm thinking of another solution.

https://news.ycombinator.com/item?id=24232809
> I also found a simple solution: call malloc_trim() after a GC. This reduces memory usage by 70%.

Issue #520, opened 2021-02-07
asTensor cannot create empty tensors


tscholak:
```
λ> Torch.Tensor.asTensor ([[]]  :: [[Int]])
Tensor Int64 [1] [ 4123162245250837044]
```
expected:
```
λ> Torch.Tensor.asTensor ([[]]  :: [[Int]])
Tensor Int64 [1,0] [[]]
```

PR #521, opened 2021-02-08
Fix empty tensors

junjihashimoto:
See https://github.com/hasktorch/hasktorch/issues/520 .


tscholak (commented):
nice, this was quick!

junjihashimoto:
I'm not sure how to use an empty tensor different from zero tensor.
But it seems to behave like pytorch.

junjihashimoto:
pytorch : 
image

hasktorch:
image

tscholak:
Interesting, I need to check if I have implemented broadcasting correctly

junjihashimoto:
Typed tensor works correctly too.
image

PR #524, opened 2021-02-18
Add functorized shapes

junjihashimoto:


tscholak:
this is looking more and more interesting, but I'm not sure if we want to experiment on this in the `Torch.Typed` API. Do you want to work with me on `Torch.GraduallyTyped`?

junjihashimoto:
I'm not sure if we want to do it, too.  But if it is possible to develop shapes with functor on `Torch.GraduallyTyped`, I want to do it.
On the other hand, I am wondering if an extensible definition is possible.
Is it difficult to use following notations/examples like a monad?
Type inference may be difficult or something impossible.

```
Tensor () --Untyped Tensor
Tensor (Vector n Float) -- 1 dimensional typed Tensor, and device is not specified.
Tensor [Float] -- Typed tensor of undecided size.
Tensor (Cpu (Vector n Float)) --Tensor on CPU
Tensor (Cpu (RequiredGrad (Vector n Float))) --Tensor with RequirementsGrad on CPU
```

junjihashimoto:
Apart from the tensor definition, this PR still has compilation errors, probably because type inference isn't working.

tscholak:
hm, this is intriguing.
Some issue I see with this is that of nesting and ordering. if all of these constructors are `Type -> Type`, then you can compose them in a bunch of ways that don't make sense or cannot be supported, like a tensor that's half on cpu and half on gpu.

hasktorch/src/Torch/Typed/Tensor.hs
```diff
@@ -120,9 +126,11 @@ instance (KnownNat n) => KnownDevice '( 'D.CPU, n) where
 instance (KnownNat n) => KnownDevice '( 'D.CUDA, n) where
   deviceVal = D.Device D.CUDA (natValInt16 @n)
 
-data Tensor (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (shape :: [Nat]) where
+data Tensor (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (shape :: [Type->Type]) where
   UnsafeMkTensor :: forall device dtype shape. {toDynamic :: D.Tensor} -> Tensor device dtype shape
 
+type Tensor' device dtype shape = Tensor device dtype (FromNats shape)
+
```
junjihashimoto, line 133: This PR defines old tensor(Tensor') using [Nat] as type synonym.
It occurs an illegal type synonym family application error in a LSTM implementation .

PR #526, opened 2021-02-27
update stack-macos to 8.10.4

tscholak:


tscholak:
This fails because of:

2021-02-27T01:13:15.4966160Z libtorch-ffi                     > error: invalid argument '-std=c++14' not allowed with 'C'
2021-02-27T01:13:15.4981840Z libtorch-ffi                     > `gcc' failed in phase `C Compiler'. (Exit code: 1)

tscholak:
@junjihashimoto I think this an old problem. How did we use to solve this?

austinvhuang:
Maybe try setting the c++ compiler to g++ instead of gcc?

tscholak:
So it turns out that, on macOS, libtorch-ffi also doesn't build with cabal in the nix-shell.
The workaround is to activate the gcc flag,
```
cabal build -f gcc
```
Is there any reason why we do not want this flag to be always on on macOS?
If we do, then we could simplify this section in `libtorch-ffi.cabal`:
```
 if os(darwin)
  ld-options: -Wl,-keep_dwarf_unwind
  cpp-options: -DENABLE_DUMMY_MALLOC_TRIM
  if !flag(gcc)
    ghc-options:       -optc-std=c++14 -optc-xc++
    cxx-options:       -std=c++14
```

junjihashimoto:
Before upgrading inline-c-cpp, the build of inline-c-cpp failed on macos.
This pr have passed the compilation of inline-c-cpp.

junjihashimoto:
We can't always use gcc on macos. When we use xcode, the compiler is llvm not gcc.

junjihashimoto:
https://hackage.haskell.org/package/inline-c-cpp-0.4.0.3/src/inline-c-cpp.cabal
image

To build libtorch-ffi of llvm on macos, we should set `-optcxx-std=c++11` like inline-c-cpp.

junjihashimoto:
I'm sorry it's complicated.
This option is not needed for gcc, because c++14 is the default.

tscholak:
still doesn't work :(

junjihashimoto:
Thank you for upgrading ghc!

PR #532, opened 2021-03-13
Remove dependency on singletons

mcwitt:
The 3.0 release of `singletons` split up into 3 packages so `Data.Singletons.Prelude` is now in `singletons-base`. I think this is the relevant [PR](https://github.com/goldfirere/singletons/pull/463).

This PR ~adds an upper bound to the hasktorch cabal file to fix for now~ removes the dependence on singletons, instead defining `Product` in `Torch.Typed.Aux`.

### DIFF
--- a/hasktorch/hasktorch.cabal
+++ b/hasktorch/hasktorch.cabal
@@ -103,7 +103,7 @@ library
                     , safe-exceptions
                     , random
                     , reflection
-                    , singletons
+                    , singletons < 3
                     , stm
                     , JuicyPixels
                     , vector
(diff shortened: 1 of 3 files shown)


tscholak:
Hey Matt, I didn't know we depended on singletons, can you confirm that we actually need it?

tscholak:
perfect, thanks!

mcwitt:
> Hey Matt, I didn't know we depended on singletons, can you confirm that we actually need it?

Good call, it looks like we were only depending on singletons in one place for `Product`. I added a product implementation to `Torch.Typed.Aux` and removed the singletons dependency in d4e34bc.

tscholak (approved):
thanks @mcwitt!

junjihashimoto:
Thx!

PR #533, opened 2021-03-14
Update dependencies, remove HLS from custom-tools

mcwitt:
Since haskell-language-server 1.0 is now in Hackage, we no longer need to add it as a custom tool.

Also removed `nixpkgs-2009` from `sources.json`, because it seemed like this was unused (?)


tscholak:
thank you! this will now take forever to build ;)

mcwitt:
> thank you! this will now take forever to build ;)

At what stage does cachix get updated with the new artifacts?

tscholak:
> At what stage does cachix get updated with the new artifacts?
Only at the very end :/

tscholak:
@mcwitt when I get the chance, I will try to build the `nix-macos` derivation locally and push it to cachix. not sure what's going on with `nix-jupyterlab`

mcwitt:
> not sure what's going on with nix-jupyterlab
@tscholak looks to me like it ran out of memory (error 137 at the end). Is there a way we can bump that up?

tscholak:
quick update: I have to use a newer nixpkgs (nixpkgs-unstable probably) because otherwise I think Big Sur won't be supported.

tscholak:
or, alternatively, I build gcc on another machine with an older version of macOS, and push to cachix from there.

tscholak:
@mcwitt can you merge https://github.com/hasktorch/hasktorch/tree/mcwitt-update-deps into your branch please?

tscholak:
@mcwitt I've got hls 1.0 working again on Mac OS Big Sur with ghc 8.10.4. Please merge https://github.com/hasktorch/hasktorch/tree/mcwitt-update-deps into this branch at your earliest convenience so that we can wrap this up.

mcwitt:
@tscholak done!

PR #536, opened 2021-03-23
Added named tensor

junjihashimoto:
I implemented 3 different tensors to define named-tensor as follows.

- This PR with `Unnamed`-typeclass.
- https://github.com/hasktorch/hasktorch/pull/537
- https://github.com/hasktorch/hasktorch/pull/524

The method other than this PR is a lot of modification of the existing code.
When considering a way to gradually update an existing implementation to a new tensor (Named Tensor), it is better to put a layer that abstracts the `Unnamed Tensor` instead of modifying the definition of `Tensor` directly. 
So I'm making `Unnamed-typeclass` as the layer.
In addition to this, we can use multiple Tensor definitions by using that layer.

I'm thinking broadcast for the namedtensor.

### DIFF
--- a/hasktorch/src/Torch/Typed/Factories.hs
+++ b/hasktorch/src/Torch/Typed/Factories.hs
@@ -36,6 +36,18 @@ import Torch.Typed.Aux
 import Torch.Typed.Tensor
 import Prelude hiding (sin)
 
+instance Semigroup (Tensor device dtype shape) where
+  (<>) a b = UnsafeMkTensor $ D.add (toDynamic a) (toDynamic b)
+
+instance (TensorOptions shape dtype device) => Monoid (Tensor device dtype shape) where
+  mempty = zeros
+
+instance Semigroup (NamedTensor device dtype shape) where
+  (<>) a b = fromUnnamed . UnsafeMkTensor $ D.add (toDynamic a) (toDynamic b)
+
+instance (TensorOptions shape' dtype device, shape' ~ ToNats shape) => Monoid (NamedTensor device dtype shape) where
+  mempty = fromUnnamed zeros
+
 zeros ::
   forall shape dtype device.
   (TensorOptions shape dtype device) =>
--- a/hasktorch/src/Torch/Typed/Factories.hs
+++ b/hasktorch/src/Torch/Typed/Factories.hs
@@ -35,18 +35,13 @@ import qualified Torch.TensorOptions as D
 import Torch.Typed.Aux
 import Torch.Typed.Tensor
 import Prelude hiding (sin)
+import Data.Default.Class
 
-instance Semigroup (Tensor device dtype shape) where
-  (<>) a b = UnsafeMkTensor $ D.add (toDynamic a) (toDynamic b)
+instance (TensorOptions shape dtype device) => Default (Tensor device dtype shape) where
+  def = zeros
 
-instance (TensorOptions shape dtype device) => Monoid (Tensor device dtype shape) where
-  mempty = zeros
-
-instance Semigroup (NamedTensor device dtype shape) where
-  (<>) a b = fromUnnamed . UnsafeMkTensor $ D.add (toDynamic a) (toDynamic b)
-
-instance (TensorOptions shape' dtype device, shape' ~ ToNats shape) => Monoid (NamedTensor device dtype shape) where
-  mempty = fromUnnamed zeros
+instance (TensorOptions shape' dtype device, shape' ~ ToNats shape) => Default (NamedTensor device dtype shape) where
+  def = fromUnnamed zeros
 
 zeros ::
   forall shape dtype device.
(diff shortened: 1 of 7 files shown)


hasktorch/src/Torch/Typed/Functional.hs
```diff
@@ -654,13 +654,15 @@ pow exponent input = unsafePerformIO $ ATen.cast2 ATen.Managed.pow_tt input expo
...
+   Unnamed t, device ~ (UTDevice t), dtype ~ (UTDType t), shape ~ (UTShape t)
+  ) =>
   -- | input
-  Tensor device dtype shape ->
+  t ->
   -- | output
-  Tensor device dtype shape
-relu input = unsafePerformIO $ ATen.cast1 ATen.Managed.relu_t input
+  t
+relu input = unWrap $ unsafePerformIO $ ATen.cast1 ATen.Managed.relu_t (Wrap input)
```
junjihashimoto, line 665: This modification does not effect current implementation like static-mnist-cnn of example-directory.

hasktorch/src/Torch/Typed/Factories.hs
```diff
@@ -36,6 +36,18 @@ import Torch.Typed.Aux
 import Torch.Typed.Tensor
 import Prelude hiding (sin)
 
+instance Semigroup (Tensor device dtype shape) where
+  (<>) a b = UnsafeMkTensor $ D.add (toDynamic a) (toDynamic b)
+
+instance (TensorOptions shape dtype device) => Monoid (Tensor device dtype shape) where
```
tscholak, line 42: why do you need this concrete monoid for Tensor and NamedTensor?
Perhaps consider https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Monoid.html#v:Sum or https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Monoid.html#v:Product from base.

hasktorch/src/Torch/Typed/Tensor.hs
```diff
@@ -599,3 +665,43 @@ toDouble t = D.asValue . toDynamic . toCPU $ t
...
 toBool t = D.asValue . toDynamic . toCPU $ t
+
+--------------------------------------------------------------------------------
+-- NamedTensor
+--------------------------------------------------------------------------------
+
+data NamedTensor (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (shape :: Shape) where
+  FromTensor :: forall device dtype shape' shape. shape ~ ToNats shape' => Tensor device dtype shape ->  NamedTensor device dtype shape'
+
+instance Unnamed (NamedTensor device dtype shape) where
```
tscholak, line 676: do you have an instance for `Tensor device dtype shape`, too?

hasktorch/src/Torch/Typed/Tensor.hs
```diff
@@ -599,3 +665,43 @@ toDouble t = D.asValue . toDynamic . toCPU $ t
...
+  type UTShape (NamedTensor device dtype shape) = ToNats shape
+  type UTDevice (NamedTensor device dtype shape) = device
+  type UTDType (NamedTensor device dtype shape) = dtype
+  toUnnamed (FromTensor t) = t
+  fromUnnamed = FromTensor
+  toDynamic (FromTensor (UnsafeMkTensor t)) = t
+
+instance (KnownDevice device) => Num (NamedTensor device dtype shape) where
```
tscholak, line 684: you may be able to use the `Num` instance for `Tensor device dtype shape` here

tscholak (approved):
LGTM, well done!

hasktorch/src/Torch/Typed/Tensor.hs
junjihashimoto: Yes. [This code](https://github.com/hasktorch/hasktorch/pull/536/files/2ab534287bd778c9325ead060f1a878fde0453da#diff-5757ffa202fb7ecd16a6aa1c8217948a432d109d10ae364fdb8963f6e2303d80R162) is the instance.

hasktorch/src/Torch/Typed/Factories.hs
junjihashimoto: I need a default constructor like zeros for this code.
https://github.com/hasktorch/hasktorch/pull/536/files/2ab534287bd778c9325ead060f1a878fde0453da#diff-ed8556e187a95c50a009496221d2e54ca9fe54ceafd52bbabd8470f10c1677fbR69-R79

Is it better to have an instance of Data.Default?

hasktorch/src/Torch/Typed/Tensor.hs
```diff
@@ -682,21 +683,21 @@ instance Unnamed (NamedTensor device dtype shape) where
...
-  abs t = fromUnnamed $ UnsafeMkTensor $ abs $ toDynamic t
-  signum t = fromUnnamed $ UnsafeMkTensor $ signum $ toDynamic t
-  fromInteger i = fromUnnamed $ UnsafeMkTensor . D.toDevice (deviceVal @device) . D.asTensor @Int $ fromInteger @Int i
+  (+) a b = fromUnnamed $ toUnnamed a + toUnnamed b
+  (-) a b = fromUnnamed $ toUnnamed a - toUnnamed b
+  (*) a b = fromUnnamed $ toUnnamed a * toUnnamed b
+  negate = fromUnnamed . negate . toUnnamed
+  abs = fromUnnamed . abs . toUnnamed
+  signum = fromUnnamed . signum . toUnnamed
+  fromInteger = fromUnnamed . fromInteger
```
junjihashimoto: I've tried using an instance of Num. It looks good.

junjihashimoto:
@tscholak 
Thank you for reviewing!

PR #537, opened 2021-03-23
Change the default tensor to named tensor

junjihashimoto:
This PR is just trial codes for named tensor.

### DIFF
--- a/hasktorch/src/Torch/Typed/Tensor.hs
+++ b/hasktorch/src/Torch/Typed/Tensor.hs
@@ -31,6 +31,7 @@ import Data.Kind
 import Data.Maybe
 import Data.Proxy
 import Data.Reflection
+import Data.Type.Equality
 import Foreign.ForeignPtr
 import Foreign.Storable
 import GHC.Exts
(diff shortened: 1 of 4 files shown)


hasktorch/src/Torch/Typed/Tensor.hs
```diff
@@ -120,12 +127,13 @@ instance (KnownNat n) => KnownDevice '( 'D.CPU, n) where
 instance (KnownNat n) => KnownDevice '( 'D.CUDA, n) where
   deviceVal = D.Device D.CUDA (natValInt16 @n)
 
-data Tensor (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (shape :: [Nat]) where
+
+data Tensor (device :: (D.DeviceType, Nat)) (dtype :: D.DType) (shape :: Shape) where
   UnsafeMkTensor :: forall device dtype shape. {toDynamic :: D.Tensor} -> Tensor device dtype shape
```
junjihashimoto, line 133: This pr replaces  tensor's shape( [Nat] ) with  [Type->Type].
The codes become harder to understand the code.
I think the boundary between Named Tensor and non-Named Tensor is vague.

PR #540, opened 2021-04-07
Add rename

junjihashimoto:
Low level api does not have `rename` function of named-tensor.
This PR adds the function.
https://pytorch.org/docs/stable/named_tensor.html

The way to add the function is adhoc.
https://github.com/hasktorch/hasktorch/blob/master/spec/cppclass/gen.sh
I should stop using this `gen.sh` and use `Declarations.yaml`.


tscholak (approved):
LGTM, thanks for the prompt response AND implementation!

junjihashimoto:
Thank you for reviewing!

PR #543, opened 2021-04-13
Add mnist-mixed-precision example to hie.yaml

jul1u5:
[mnist-mixed-precision example](examples/mnist-mixed-precision) was missing from [hie.yaml](hie.yaml)


junjihashimoto:
Thx!

PR #554, opened 2021-05-19
Change libstdc++ to stdenv.cc.cc in libtorch-nix

junjihashimoto:
This is related to https://github.com/hasktorch/tokenizers/pull/4.
libtorch-nix uses libcxx derivation. libcxx derivation is llvm's c++ library.
Normally we should use stdenv.cc.cc, so fix it.

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -53,10 +53,10 @@
         "homepage": "https://pytorch.org/",
         "owner": "hasktorch",
         "repo": "libtorch-nix",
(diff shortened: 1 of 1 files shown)


tscholak:
Do you need me to test this on Mac?

junjihashimoto:
@tscholak 
CI tests this on Mac, so I think you do not need to test it.
But when we merge tokenizer, we should test it on both mac and linux .

PR #557, opened 2021-05-26
Add a HasForward instance of ScriptModule

junjihashimoto:
At first, this pr removes ScriptModule's forward function to avoid forward function collisions.
Then it adds the instance of HasForward for ScriptModule.


austinvhuang:
Thanks @junjihashimoto! LGTM

junjihashimoto:
Thx!

PR #558, opened 2021-05-31
Fix duplicated definitions for both save and load

junjihashimoto:


austinvhuang:
LGTM! Will update https://github.com/hasktorch/hasktorch/pull/553 when this is merged.

PR #569, opened 2021-06-20
updated authors to `hasktorch contributor team` instead of `austin huang`

austinvhuang:
Mainly for attribution on the tutorial page https://hasktorch.github.io/tutorial/index.html


tscholak:
I think this is a great idea that recognizes the contributions of many! Thank you Austin!

PR #570, opened 2021-06-22
Add python-style slice operator

junjihashimoto:
This PR adds the slice operator of python-style with ~~`pySlice`~~ `slice` quote.

When you get values in python looks like this,

```python
tensor[0,0:3]
```

hasktorch uses ~~pySlice~~ slice to get the value with `!`  and ~~`pySlice`~~ `slice` as follows:

```
tensor ! [slice|0,0:3|]
```


austinvhuang:
This looks good @junjihashimoto ! Could you pick a shorter QQ prefix? Maybe just "slice"? (was originally thinking just a single character "s" but that might be too cryptic looking).

junjihashimoto:
@austinvhuang 
Renamed existing slice-function to sliceDim and renamed pySlice to slice.

PR #571, opened 2021-06-24
SingKind instances

tscholak:


jul1u5:
Looks great! I fixed the failing doctests, it would be nice if CI could catch those.

PR #576, opened 2021-07-04
Add lens for index

junjihashimoto:
To convert index into lens, add lslice function of template haskell.
For example, you can write getter and setter for tensor as follows.
These are the odd element accessors of the tensor.

```haskell
-- getter 
tensor ^. [lslice|1::2|]
-- setter
tensor & [lslice|1::2|] .~ 2
```


junjihashimoto:
I want to use this pr at https://github.com/hasktorch/hasktorch-serving-models-skeleton, so merge it.

PR #578, opened 2021-07-06
make forward monadic and support MonadThrow

tscholak:
I've made the gradually typed `HasForward` class monadic in the return type. I propagated that change through all `Torch.GraduallyTyped` modules and was able to remove a bunch of `unsafePerformIO` calls because of it.


jul1u5 (approved):
LGTM! I fixed failing doctests, I think they got merged in from the previous PR.

tscholak:
I'm getting segfaults, unfortunately. something is very wrong

PR #584, opened 2021-07-14
Use spec/Declarations.yaml to generate tensor-methods

junjihashimoto:
Current codegen uses gen.sh to generate method-functions of tensor.
https://github.com/hasktorch/hasktorch/blob/4e96f7a0688db1ec6f4d66510ff85241ede107dc/spec/cppclass/gen.sh
`Declarations.yaml` has the list of method-functions.
Before supporting libtorch-1.9.0, this PR modifies codegen to stop `gen.sh` and use `Declarations.yaml`.


tscholak (approved):
Amazing 🤩 
I hope it doesn't break anything downstream. Feel free to merge 😊

junjihashimoto:
Thank you for reviewing!

PR #585, opened 2021-07-15
Bump libtorch to 1.9

junjihashimoto:
- Add the new type of `at::ArrayRef<at::Scalar>`  in codegen.
- Fix the parser of `Declaration.yaml`. (Just remove the field of `matches_jit_signature`.)
- Add some APIs to blacklists of codegen due to duplicate definitions of suffix of functions and ambiguous overload of c++.
- Fix the arguments of `Torch.Internal.Unmanaged.Module.dumpToStr`
- Fix the doctests of `Torch.Typed.Functional` for deprecated APIs.

### DIFF
--- a/.github/workflows/cabal-macos.yaml
+++ b/.github/workflows/cabal-macos.yaml
@@ -18,7 +18,7 @@ jobs:
         brew install ghc@8.10 || true
         brew install cabal-install || true
         brew tap hasktorch/libtorch-prebuild https://github.com/hasktorch/homebrew-libtorch-prebuild || true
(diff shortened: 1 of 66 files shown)


tscholak:
I love that `nix-linux` is the only build that works so far :p

junjihashimoto:
Why do we get this error only on macos?
image

tscholak:
That's funky, I don't know

tscholak (approved):
Fantastic, thanks for this amazing work!

junjihashimoto:
Thank you for reviewing!

PR #588, opened 2021-07-23
Expose in Functional and add layers for all conv-nd and convTranspose-nd 

Reep236:
Typed and Untyped convolution and transposed convolution functions and layers. Specs added in Untyped to match the existing conv2d test. Would like to add a Torch.GraduallyTyped.NN.Convolution, but I haven't looked into or worked with GraduallyTyped more than in passing. Typed and Untyped seem to behave and test as expected so far which makes sense given they were an expansion built off of existing scaffolding. 

Also includes `InstanceNorm`. I followed the convention set by `BatchNorm` of not having any Parameterized instance, but I could see it being helpful to have one for working with full models that use the layers, especially with serialization and deserialization. 

Hope this all looks good, and I'd be glad for any feedback or suggestions!

### DIFF
--- a/hasktorch/src/Torch/Functional.hs
+++ b/hasktorch/src/Torch/Functional.hs
@@ -1852,6 +1852,19 @@ conv1d weight bias stride padding dilation groups input =
       dilation
       groups
 
+conv1d' :: 
+  -- | weight
+  Tensor ->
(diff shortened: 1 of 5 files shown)


junjihashimoto:
LGTM!

PR #592, opened 2021-07-28
Indexing/slicing

jul1u5:


experimental/gradually-typed/src/Torch/GraduallyTyped/Shape/Type.hs
```diff
@@ -350,13 +350,11 @@ instance SingKind (Shape [Dim (Name Symbol) (Size Nat)]) where
   toSing (Checked shape) = withSomeSing shape $ SomeSing . SShape
 
 pattern (:|:) ::
```
tscholak, line 352: could you move this pattern to the Prelude?

experimental/gradually-typed/test/IndexingSpec.hs
```diff
@@ -0,0 +1,93 @@
...
+    x <- tensor ! SIndices (SEllipsis :|: SSliceAt (SIndex @2) :|: SEllipsis :|: SNil)
+
+    dimSize <$> getDims x `shouldBe` [2, 2]
+    fromTensor @((Int, Int), (Int, Int)) x `shouldBe` ((2, 5), (8, 11))
+
+  it "slices with SSliceFrom" $ do
+    x <- tensor ! SIndices (SSliceAll :|: SSliceFrom (SIndex @1) :|: SNil)
+
+    dimSize <$> getDims x `shouldBe` [2, 1, 3]
+    fromTensor @(SV.Vector 2 (SV.Vector 1 (SV.Vector 3 Int))) x
```
tscholak, line 69: quick question. can you write:
```suggestion
    fromTensor @(SV.Vector _ (SV.Vector _ (SV.Vector _ Int))) x
```
with `PartialTypeSignatures` enabled?

tscholak (approved):
Amazing! I tracked your changes, and it looks good to me!
If anything pops up later, we can revise then.
Let's merge this in as soon as the tests pass.

PR #593, opened 2021-07-29
Fix to cuda tensor

tscholak:
I had to comment the `TensorLike` instance for `Tensor`, because it would allow for switching gradients off or on from outside of `IO` which mustn't be allowed.

### DIFF
--- a/experimental/gradually-typed/src/Torch/GraduallyTyped/Tensor/Type.hs
+++ b/experimental/gradually-typed/src/Torch/GraduallyTyped/Tensor/Type.hs
@@ -1395,8 +1395,10 @@ fromTensorRaw ::
   (TensorLike a dType dims, TensorLikeRaw a, SGetDims dims) =>
   Tensor gradient layout device ('DataType dType) ('Shape dims) ->
   a
-fromTensorRaw t = unsafePerformIO $
-  withTensor t $ \ptr -> tensorPeekElemOff ptr 0 (fromInteger . dimSize <$> getDims t)
+fromTensorRaw t =
+  unsafePerformIO $
+    sSetDevice (SDevice SCPU) t
+      >>= flip withTensor (\ptr -> tensorPeekElemOff ptr 0 $ fromInteger . dimSize <$> getDims t)
 
 instance TensorLike Bool 'Bool '[] where
   sToTensor = sToTensorRaw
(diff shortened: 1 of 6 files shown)


experimental/gradually-typed/src/Torch/GraduallyTyped/Tensor/Type.hs
```diff
@@ -242,26 +242,16 @@ withGradient ::
...
-
-instance SSetGradient ('Gradient 'WithoutGradient) where
-  sSetGradient (SGradient SWithoutGradient) tensor =
-    cast2 ATen.tensor_set_requires_grad_b tensor False
+-- | Turn gradient computations off or on for a tensor.
+sSetGradient ::
+  forall gradient gradient' layout device dataType shape.
+  SGradient gradient ->
+  Tensor gradient' layout device dataType shape ->
+  IO (Tensor gradient layout device dataType shape)
```
jul1u5, line 250: Shouldn't the return type be wrapped in `MonadThrow` instead of `IO` here?

jul1u5 (commented):
Oh, I missed this. Maybe we should wrap the return type of `fromTensor` with `MonadThrow`? I could make a PR for this.

tscholak:
It has to be IO, unfortunately. (Or a special yet-to-be-defined TorchMonad that evaluates to IO.) The problem is that turning on or off gradients for an existing tensor changes the global state of the torch runtime. As a side effect, that runtime builds the compute graph in order to be able to backpropagate and compute gradients when requested. That messes with referential transparency.

jul1u5 (approved):
Oh, ok, that makes sense. I think we can merge this.

experimental/gradually-typed/src/Torch/GraduallyTyped/Tensor/Type.hs
```diff
@@ -1404,17 +1386,19 @@ sToTensorRaw gradient' layout device x = do
       t <- UnsafeTensor <$> cast2 ATen.empty_lo dims' opts
       withTensor t $ \ptr ->
         tensorPokeElemOff ptr 0 dims' x
-      pure t
+      sSetDevice device t
```
tscholak, line 1389: @jul1u5 fyi

experimental/gradually-typed/src/Torch/GraduallyTyped/Tensor/Type.hs
```diff
@@ -1404,17 +1386,19 @@ sToTensorRaw gradient' layout device x = do
...
 fromTensorRaw ::
   forall gradient layout device a dType dims.
   (TensorLike a dType dims, TensorLikeRaw a, SGetDims dims) =>
   Tensor gradient layout device ('DataType dType) ('Shape dims) ->
   a
-fromTensorRaw t = unsafePerformIO $
-  withTensor t $ \ptr -> tensorPeekElemOff ptr 0 (fromInteger . dimSize <$> getDims t)
+fromTensorRaw t =
+  unsafePerformIO $
+    sSetDevice (SDevice SCPU) t
```
tscholak, line 1400: @jul1u5 fyi

PR #597, opened 2021-07-30
add initial two-layer-network example

tscholak:
I have started porting last-year's two-layer-network example to the gradually typed API.
There are some issues with the type signature of the optimizer's `step` function. I can't use a generator with it.


experimental/gradually-typed/src/Torch/GraduallyTyped/Examples/TwoLayerNetwork.hs
```diff
@@ -0,0 +1,407 @@
...
+      ( let inputTensor = sToTensor (SGradient SWithoutGradient) (SLayout SDense) device input
+            expectedOutputTensor = sToTensor (SGradient SWithoutGradient) (SLayout SDense) device expectedOutput
+            model = flip evalStateT stateDict $ fromStateDict modelSpec mempty
+         in (,,) <$> inputTensor <*> expectedOutputTensor <*> model
+      )
+      >>>= ( \(inputTensor, expectedOutputTensor, model) ->
+               ireturn inputTensor
+                 >>>= IxStateT . forward model
+                 >>>= ilift . (`mseLoss` expectedOutputTensor)
+           )
```
junjihashimoto, line 277: In this case, what is IxStateT's state? Is it Generator?
Could we write this funciton just like this?
```
computeLoss device modelSpec stateDict input expectedOutput =
let inputTensor = sToTensor (SGradient SWithoutGradient) (SLayout SDense) device input
     expectedOutputTensor = sToTensor (SGradient SWithoutGradient) (SLayout SDense) device expectedOutput
     model = flip evalStateT stateDict $ fromStateDict modelSpec mempty
in mseLoss <$> forward model inputTensor <*> pure  expectedOutputTensor
```

experimental/gradually-typed/src/Torch/GraduallyTyped/Examples/TwoLayerNetwork.hs
```diff
@@ -0,0 +1,407 @@
...
+     in a * maxLearningRate
+  | numWarmupEpochs < epoch && epoch < numEpochs - numCooldownEpochs =
+    let a :: Double =
+          fromIntegral (numEpochs - numCooldownEpochs - epoch)
+            / fromIntegral (numEpochs - numCooldownEpochs - numWarmupEpochs)
+     in a * maxLearningRate + (1 - a) * finalLearningRate
+  | otherwise = finalLearningRate
+
+runTwoLayerNetworkExample :: IO ()
+runTwoLayerNetworkExample = do
```
junjihashimoto, line 327: I think this is the main function of two layer network, and runTwoLayerNetworkExample is not called by test.
Should we use ghci to call it?

experimental/gradually-typed/src/Torch/GraduallyTyped/Examples/TwoLayerNetwork.hs
```diff
@@ -0,0 +1,407 @@
...
+    beta1 :: Double,
+    -- | beta2
+    beta2 :: Double,
+    -- | epsilon
+    epsilon :: Double,
+    -- | weight decay
+    weightDecay :: Double,
+    -- | use amsgrad
+    amsgrad :: Bool
+  }
```
junjihashimoto, line 200: Would you like to create Torch.GraduallyTyped.Optim later and include it?

junjihashimoto (approved):
Awesome!
I'm glad to see this example.

experimental/gradually-typed/src/Torch/GraduallyTyped/Examples/TwoLayerNetwork.hs
```diff
@@ -0,0 +1,407 @@
...
+train seed device modelSpec stateDict learningRate examples = do
+  g <- sMkGenerator device seed
+  optim <- mkAdam defaultAdamOptions {learningRate} (Map.elems stateDict)
+  let step g' (x, y, _iter) = do
+        let f :: ForeignPtr ATen.TensorList -> IO (ForeignPtr ATen.Tensor)
+            f tensorList = do
+              stateDict' :: StateDict <- ATen.uncast tensorList (pure . Map.fromList . zip (Map.keys stateDict))
+              UnsafeTensor t <- fst <$> computeLoss device modelSpec stateDict' x y g'
+              pure t
+        _loss :: ForeignPtr ATen.Tensor <- ATen.cast2 ATen.step optim f
```
junjihashimoto, line 303: I haven't tested this step function very much. Please let me know if there is anything.

junjihashimoto: Instead, `unsafeStep` has been tested.

experimental/gradually-typed/src/Torch/GraduallyTyped/Examples/TwoLayerNetwork.hs
tscholak: Yes, at some point

experimental/gradually-typed/src/Torch/GraduallyTyped/Examples/TwoLayerNetwork.hs
tscholak: I pushed the first program that worked. I've since rewritten this part. Yes, this can be simplified.

I'm reaching for the indexed state monad IxStateT by default because the type of the generator state can change. Forward functions can move the generator to a different device.

experimental/gradually-typed/src/Torch/GraduallyTyped/Examples/TwoLayerNetwork.hs
tscholak: There's a lot going on in step. Calling the loss function, calculating the gradients, 
assuming that there's only one parameter group, running the optimizer step, returning the loss, etc.

This is fine now, but there will be times when it is not. There are features that we will need to support for training transformers that are not supported with this, including:

* gradient clipping (improves stability)
* gradient checkpointing (trades space for compute)
* gradient accumulation (increases effective batch size)

It may therefore be necessary to provide a more atomic API.

experimental/gradually-typed/src/Torch/GraduallyTyped/Examples/TwoLayerNetwork.hs
tscholak: Perhaps. It would need to be optimized for speed, this runs quite slowly currently.

It's also not clear where this code should live long-term. Perhaps I'll move it to experimental/gradually-typed/examples/two-layer-network and make an executable out of it.

I've added this code mostly to get a feel for how training will look like, and what abstractions will be useful.

There will be more examples. @jul1u5 wants to do one on MNIST, and I want to do one with transformers, likely BERT and T5.

experimental/gradually-typed/src/Torch/GraduallyTyped/Examples/TwoLayerNetwork.hs
tscholak: I'm not sure I get unsafeStep. It doesn't seem to have a good signature given that the optimizer mutates the parameter tensor list.

tscholak: I also can't imagine that any of this is thread-safe. If you have two optimizers updating the same tensor list, this will certainly give you garbage

tscholak:
@junjihashimoto @jul1u5 the code is still unusable because of the issue with the `StdTuple` conversion. However, the other changes I've made are useful and should go in, so I'm merging this.

PR #599, opened 2021-08-01
Change generator-definition

junjihashimoto:


tscholak (approved):
LGTM!

hasktorch/src/Torch/Random.hs
```diff
@@ -33,53 +34,68 @@ import qualified Torch.Internal.Type as ATen
...
+    generatorIsCpu gen = unsafePerformIO $ cast1 ATen.generator_is_cpu gen
+    
+    generatorIsCuda :: ForeignPtr ATen.Generator -> Bool
+    generatorIsCuda gen = unsafePerformIO $ cast1 ATen.generator_is_cuda gen
+    
+    generatorDevice :: ForeignPtr ATen.Generator -> Int
+    generatorDevice gen = unsafePerformIO $ cast1 ATen.generator_get_device gen
+    
+    generatorSeed :: ForeignPtr ATen.Generator -> Word64
+    generatorSeed gen = fromIntegral (unsafePerformIO $ cast1 ATen.generator_get_device gen :: Int64)
```
junjihashimoto, line 97: If these functions throw some exceptions, above `atomically` may be broken, but I think these functions  always succeed.

junjihashimoto: We can not call IO function in `atomically`, so I use `unsafePerformIO`.

PR #601, opened 2021-08-02
Add the wrapper of stepWithGenerator

junjihashimoto:


hasktorch/src/Torch/Optim/CppOptim.hs
```diff
@@ -43,6 +44,21 @@ data CppOptimizerState option = CppOptimizerState option CppOptimizerRef
...
+  -> IO (Tensor, ForeignPtr ATen.Generator)
+stepWithGenerator o@(CppOptimizerState _ ref) generator loss = do
+  (v, nextGenerator) <- cast3 LibTorch.stepWithGenerator ref generator loss'
+  return (v, nextGenerator)
+  where
+    loss' :: ForeignPtr ATen.TensorList -> ForeignPtr ATen.Generator -> IO (ForeignPtr (ATen.StdTuple '(ATen.Tensor,ATen.Generator)))
+    loss' params gen = do
+      (v :: Tensor,gen') <- uncast params $ \params' -> loss params' gen
+      v' <- cast v pure :: IO (ForeignPtr ATen.Tensor)
+      cast (v',gen') pure
```
tscholak, line 60: This will crash because that instance isn't fully implemented. Can it be?

tscholak (changes requested):

hasktorch/src/Torch/Optim/CppOptim.hs
junjihashimoto: Probably yes. 
https://github.com/hasktorch/hasktorch/pull/601/commits/021afe6e592abb1fb1ae3183aacd32c2831cf023

libtorch-ffi/src/Torch/Internal/Cast.hs
```diff
@@ -189,7 +189,10 @@ instance (Castable a a', Castable b b', Castable c c',
 --------------------------------------------------------------------------------
 
 instance (CppTuple2 c, Castable a (A c), Castable b (B c)) => Castable (a,b) c where
-  cast _ _ = error "Attempted to cast a 2-tuple from Haskell to C++, this is not supported."
+  cast (t0',t1') f = do
+    cast t0' $ \t0 -> 
+      cast t1' $ \t1 -> 
+        makeTuple2 (t0,t1) >>= f
```
junjihashimoto, line 195: Add cast-function to create a StdTuple from a tuple of pointers.

libtorch-ffi/src/Torch/Internal/Unmanaged/Type/Tuple.hs
```diff
@@ -172,3 +172,4 @@ instance CppTuple2 (Ptr (StdTuple '(Tensor,Generator))) where
   type B (Ptr (StdTuple '(Tensor,Generator))) = Ptr Generator
   get0 v = [C.throwBlock| at::Tensor* { return new at::Tensor(std::get<0>(*$(std::tuple<at::Tensor,at::Generator>* v)));}|]
   get1 v = [C.throwBlock| at::Generator* { return new at::Generator(std::get<1>(*$(std::tuple<at::Tensor,at::Generator>* v)));}|]
+  makeTuple2 (a,b) = [C.throwBlock| std::tuple<at::Tensor,at::Generator>* { return new std::tuple<at::Tensor,at::Generator>(std::make_tuple(*$(at::Tensor* a),*$(at::Generator* b)));}|]
```
junjihashimoto, line 175: I should test this function.

tscholak: I'm going to test it as well

tscholak (approved):
I tried it, and it seems to work. Not sure if there's a memory leak, though. If there is one, it's not a critical one.

PR #602, opened 2021-08-03
Template Haskell for slicing (Gradually Typed)

jul1u5:


tscholak (approved):
Thanks, this is great work!

PR #610, opened 2021-08-10
Fix IOHK's cache

junjihashimoto:
To use cached ghc of IOHK, update substituters of nix.

https://input-output-hk.github.io/haskell.nix/tutorials/getting-started/

Both cachix and hydra are set in the plutus-community documentation, so both are set in the same way.
https://docs.plutus-community.com/docs/setup/Ubuntu.html


tscholak:
This assumes that the user is a trusted user which may not be the case. Can we just link to the Haskell.nix documentation instead?

junjihashimoto:
Sure!

PR #616, opened 2021-08-28
Add pre-commit hooks

mcwitt:
**Note**: most of the diff comes from the final 2 commits, which apply the formatting rules to all (non-excluded) files using `pre-commit run`.

Adds https://github.com/cachix/pre-commit-hooks.nix to the build, and configures with `ormolu` and `nixpkgs-fmt` hooks.

Inside a nix shell, the hooks are automatically configured and installed (so that they run on `git commit`).

Also adds a CI workflow that returns pre-commit check status. Currently this just runs in parallel with the other workflows; eventually it might make sense for the other workflows to depend on success so that we fail fast on trivial errors.


tscholak (approved):
I scrolled through all the changes. I didn't see any obvious problems, and since the tests succeeded, I'd like to merge it! Well done!

PR #628, opened 2021-11-30
Rename Aux to Auxiliary

junjihashimoto:
In windows, some tools can not clone `Aux.hs` file.
So this pr renames `Aux.hs` to `Auxiliary.hs`.


junjihashimoto:
Not only the restrictions of windows but also vscode does not support Aux.* file.
https://github.com/microsoft/vscode/blob/e4efcb3be9bcdc46edeb02d49361d18c205c6940/src/vs/base/common/extpath.ts#L159-L196

PR #630, opened 2021-12-09
Add nms for a motivational example of Record Dot Syntax

junjihashimoto:
Is it possible to combine Record Dot Syntax and named tensor?
I make a motivational example for that.
The python code of nms below is concise enough, but I'm wondering if `Record Dot Syntax` would make it even easier for users to write codes.

```python
# https://github.com/rbgirshick/fast-rcnn/blob/master/lib/utils/nms.py#L8-L37
def nms(dets, thresh):
    x1 = dets[:, 0]
    y1 = dets[:, 1]
    x2 = dets[:, 2]
    y2 = dets[:, 3]
    scores = dets[:, 4]

    areas = (x2 - x1 + 1) * (y2 - y1 + 1)
    order = scores.argsort()[::-1]

    keep = []
    while order.size > 0:
        i = order[0]
        keep.append(i)
        xx1 = np.maximum(x1[i], x1[order[1:]])
        yy1 = np.maximum(y1[i], y1[order[1:]])
        xx2 = np.minimum(x2[i], x2[order[1:]])
        yy2 = np.minimum(y2[i], y2[order[1:]])

        w = np.maximum(0.0, xx2 - xx1 + 1)
        h = np.maximum(0.0, yy2 - yy1 + 1)
        inter = w * h
        ovr = inter / (areas[i] + areas[order[1:]] - inter)

        inds = np.where(ovr <= thresh)[0]
        order = order[inds + 1]

    return keep
```

### DIFF
--- a/hasktorch/src/Torch/Typed/Vision.hs
+++ b/hasktorch/src/Torch/Typed/Vision.hs
@@ -141,3 +141,75 @@ initMnist path = do
   testImagesBS <- decompressFile path "t10k-images-idx3-ubyte.gz"
   testLabelsBS <- decompressFile path "t10k-labels-idx1-ubyte.gz"
   return (MnistData imagesBS labelsBS, MnistData testImagesBS testLabelsBS)
+
+data Box a = Box
+  { x1 :: a,
+    y1 :: a,
+    x2 :: a,
+    y2 :: a,
+    score :: a
+  }
+  deriving (Show, Eq, Generic, Default)
+
+nms_cpu :: Num a => [Box a] -> a -> [Box a]
+nms_cpu dets = nms_cpu' (sortBy score dets)
+  where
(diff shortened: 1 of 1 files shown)


austinvhuang:
I think any ADT in the NN.hs files would be a motivating example for the overloaded field names aspect at least. Many of them should really be sharing field names (eg "bias") but end up relying on awkward prefix conventions instead.

In spite of the improvement though, I would give the compiler team a long period of bugfixing time before cutting over considering our past issues with bleeding edge GHC versions (particularly with GHC 9+).

junjihashimoto:
@austinvhuang 
When we apply `Record Dot Syntax` to NN, we will not be able to support anything less than ghc-9.2.
However, using Hasfield like genericNms of this PR will allow us to create more generic functions.

PR #635, opened 2022-03-04
Add instances of complex numbers

junjihashimoto:
Add TensorLike instances for complex numbers to read and write them.

```
ghci> :m + Data.Complex
ghci> a = [1,2,3] :: [Complex Double]
ghci> b = asTensor a
ghci> b * b
Tensor ComplexDouble [3] [ 1.0000   ,  4.0000   ,  9.0000   ]
ghci> b + b
Tensor ComplexDouble [3] [ 2.0000   ,  4.0000   ,  6.0000   ]
ghci> b / b
Tensor ComplexDouble [3] [ 1.0000   ,  1.0000   ,  1.0000   ]
ghci> b - b
Tensor ComplexDouble [3] [ 0.0000,  0.0000,  0.0000]
ghci> a = [1 :+ 1 , 2:+2,3:+3] :: [Complex Double]
ghci> b = asTensor a
ghci> b
Tensor ComplexDouble [3] [Cannot show stacktrace
Exception: value cannot be converted to type double without overflow: (1,1); type: std::runtime_error
Cannot show stacktrace
Exception: value cannot be converted to type double without overflow: (1,1); type: std::runtime_error
*** Exception: CppStdException "Exception: value cannot be converted to type double without overflow: (1,1); type: std::runtime_error"
```

### DIFF
--- a/hasktorch/hasktorch.cabal
+++ b/hasktorch/hasktorch.cabal
@@ -134,6 +134,7 @@ library
                     , vector-sized
                     , template-haskell
                     , megaparsec
+                    , half
 
  default-extensions:  Strict
                     , StrictData
@@ -200,6 +201,7 @@ test-suite spec
                     , vector-sized
                     , lens-family-core
                     , data-default-class
+                    , half
 
 test-suite doctests
   if os(darwin) || flag(disable-doctest)
--- a/hasktorch/src/Torch/DType.hs
+++ b/hasktorch/src/Torch/DType.hs
@@ -175,6 +175,24 @@ isIntegral QUInt8 = False
 isIntegral QInt32 = False
 isIntegral BFloat16 = False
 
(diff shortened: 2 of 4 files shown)


junjihashimoto:
The show instance for complex numbers was throwing the exception.
If we fix it, we'll see something like this.
```
ghci> :m + Data.Complex
ghci> v = [1:+1,2:+2] :: [Complex Double]
ghci> v
[1.0 :+ 1.0,2.0 :+ 2.0]
ghci> a = asTensor v
ghci> a
Tensor ComplexDouble [2] [ 1.0000    + i 1.0000   ,  2.0000    + i 2.0000   ]
```

junjihashimoto:
CI does not pass on macos, because python's version is changed.
I will fix it with another PR.

PR #643, opened 2022-08-30
Add a variable length tensor for typed api

junjihashimoto:
Adds a VLTensor that represents a tensor whose length is determined at runtime with existential quantification


tscholak (approved):
LGTM

junjihashimoto:
Thank you for your review!

PR #675, opened 2023-08-24
Replace haskell.nix with stacklock2nix

junjihashimoto:
Haskell.nix may not have ghc cached (of course, it may be cached.) When it is not cached, compiling ghc will run 2-3 times. It's a waste of maintenance time because every stackage upgrade has something like that.
This PR uses stacklock2nix to reduce that time, because stacklock2nix uses nixpkgs' cached ghc.
Before trying stacklock2nix, I tried using nixpkgs ghc in haskell.nix, but it didn't work.

This is the first time to support apple silicon. I tested this on apple silicon.


junjihashimoto:
ToDo：
* Merge nix-linux-*.yml of GHA into nix-linux.yml to keep maintenance costs down. (The default is CUDA version.)
* Update nix-haddock

PR #676, opened 2023-09-18
indexSelect' must use an Int64/32 tensor for indices

schnecki:
Bug: Exception: index_select(): Expected dtype int32 or int64 for index

Solution: Do not convert Tensor using `asTensor' indexList t`, as this would convert the type of `indexList` to the type of `t`. Rather use `asTensor indexList`. Another approach would be `asTensor' indexList int64_opts`, which would however always perform a transformation to `int64`.

### DIFF
--- a/hasktorch/src/Torch/Tensor.hs
+++ b/hasktorch/src/Torch/Tensor.hs
@@ -312,7 +312,7 @@ indexSelect' ::
   Tensor ->
   -- | output
   Tensor
-indexSelect' dim indexList t = unsafePerformIO $ (cast3 ATen.index_select_tlt) t dim (asTensor indexList)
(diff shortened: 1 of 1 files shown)


junjihashimoto:
@schnecki Thank you for letting me know.
When asTensor is used, the device becomes CPU. 
So we also need to use the device of input-tensor.

schnecki:
So we probably need something like this, right? 

    _toDevice (device t) (asTensor is)

That compiles locally for me, couldn't test it though.

junjihashimoto:
@schnecki 
It is correct. nllLoss' uses the same code.
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Functional.hs#L958

junjihashimoto:
@schnecki Thx!

schnecki:
Thanks to you for your great work!

PR #682, opened 2023-10-15
CI: some workflow simplifications

andreasabel:
This PR makes some cosmetic simplifications on the workflows.
- CI: setup submodules with actions/checkout
- CI: Stack is preinstalled on the GHA runners, no need to install it
- CI: GHCup is preinstalled on the GHA runners, no need to install it

I have some more ideas for CI cosmetics, like removing redundant instructions or steps. Let me know whether such work is welcome in general or whether this would conflict with some principles behind the CI setup.

So far, I commented out redundant installation steps, let me know whether you prefer that I fully delete them.

### DIFF
--- a/.github/workflows/cabal-linux.yml
+++ b/.github/workflows/cabal-linux.yml
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+      with:
+        submodules: true
     - name: Setup packages
       run: |
         sudo apt update -qq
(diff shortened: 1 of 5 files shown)


andreasabel:
CI succeeded, except for the known `cachix` problem with the nix workflows.

junjihashimoto:
@andreasabel Thx! It's fine because CI will be faster not to install stack.
In the past, stack was not installed or there were bugs, so it installs stack to use the latest one.

Issue #684, opened 2023-10-16
CI: reducing cache sizes to better cope with the 10 GB limit


andreasabel:
The current setup of caching in CI only allows 3 instances to be kept in the cache store which GitHub limits to 10 GB.
Each instance roughly has 3 GB, according to the following snapshot:
```
2023-10-16 $ gh actions-cache list
Total caches size 8.63 GB

Showing 12 of 12 cache entries in hasktorch/hasktorch

macOS-stack--        1008.87 MB  refs/heads/master    13 hours ago
macOS-cabal-ghc928-  245.53 MB   refs/heads/master    13 hours ago
Linux-cabal-         452.37 MB   refs/heads/master    13 hours ago
Linux-stack--        1.07 GB     refs/heads/master    13 hours ago
macOS-stack--        1008.92 MB  refs/pull/681/merge  2 days ago
macOS-cabal-ghc928-  245.55 MB   refs/pull/681/merge  2 days ago
Linux-stack--        1.07 GB     refs/pull/681/merge  2 days ago
Linux-cabal-         452.32 MB   refs/pull/681/merge  2 days ago
macOS-stack--        1.09 GB     refs/heads/master    2 days ago
macOS-cabal-ghc924-  281.67 MB   refs/heads/master    2 days ago
Linux-cabal-         544.02 MB   refs/heads/master    2 days ago
Linux-stack--        1.25 GB     refs/heads/master    2 days ago
```
I think caching would overall be more effective if more instances could be kept.  Consider the following scenario:
1. Cache exists for `master`, consuming instance 1.
2. Contributor A opens a PR, consuming instance 2.
3. Contributor B opens a PR, consuming instance 3.
4. Contributor C opens a PR, bumping A's cache out of the store.
5. Contributor A updates their PR, needing a complete rebuild.  Then it bumps B's cache out of the store.
6. Contributor B updates their PR, bumping C out. etc.

Of course, there won't be a gazillion of concurrent PRs, but being able to host 3 may be good.

I propose that I do some work on reducing cache sizes, keeping only parts that are effectively reused.
For instance `dist-newstyle` and `.stack-work` seem hardly reused, almost all modules are recompiled anyway.  (But I have to investigate more.) Maybe just caching the dependencies is the good strategy.

andreasabel:
For caching Stack's build products, this investigation looks relevant:
- https://github.com/commercialhaskell/stack/issues/5754#issuecomment-1696156869

[...]

PR #685, opened 2023-10-17
CI #684: revise caching for the cabal workflows

andreasabel:
This commit restructures the workflows `cabal-linux` and `cabal-macos`.

Caching
-------

1. Only cache dependencies.

   From looking into PR runs I deduced that `cabal` hardly reuses anything from `dist-newstyle`.
   Thus, this local build directory is dropped from the cache, only the global `~/.cabal/store` is now cached.
   This dramatically reduces the cache sizes (here with GHC 9.2.8):

   | cabal  | linux  | macos  |
   |--------|--------|--------|
   | before | 450 MB | 250 MB |
   | after  |  85 MB | 100 MB |

2. Save dependency cache early.

   Splitting `actions/cache` into `/restore` and `/save` step,
   we can immediately cache the dependencies after they have been build.
   So they will be preserved even if later the building of `hasktorch` or some test fails.
   This will speed up the next attempt, because currently around 30min are spent on building the dependencies.
   This is half of the build time on `cabal-macos` and short of 40% on `cabal-linux`.

3. Use more precise cache keys.

   The new cache keys feature 4 components:
   1. OS
   2. Cabal version
   3. GHC version
   4. Hash of the build plan

   The latter is created by a `cabal build all --dry-run` that determines all the versions of the dependencies.

   Caches can be restored even if the build plan changes, but not if Cabal or GHC version change.
   - Build products from other GHCs can in general not be reused.
   - Build products from other Cabal versions might be risky to reuse.

Other changes
-------------

- Skip installation of `Cabal` as it is preinstalled on the GHA runners.

- Skip installation of `hspec-discover` as it is not used in the code base (probably it was used in the past).
  Consequently, we do not need `.cabal/bin` in the `PATH`.

- Break job into smaller steps.  This has two advantages:
  1. We get timing info for each step at our fingertips.
  2. In case of breakage, individual step are easier to link to in the logs.

- Drop the `source setenv` step everywhere.
  It seems unnecessary.

- Print some information about the Haskell setup, e.g. paths.
  Useful for debugging.

### DIFF
--- a/.github/workflows/cabal-linux.yml
+++ b/.github/workflows/cabal-linux.yml
@@ -6,10 +6,14 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    env:
+      GHC: '9.2.8'
     steps:
+
     - uses: actions/checkout@v4
       with:
         submodules: true
+
     - name: Setup packages
       run: |
         sudo apt update -qq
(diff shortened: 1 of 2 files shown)


junjihashimoto:
@andreasabel Awesome!

PR #700, opened 2024-05-01
Overhaul Nix Infrastructure

collinarnett:
This PR removes the dependecy on haskell.nix and instead relies on the vanilla haskell infrastructure for nix. The primary means of exporting hasktorch will now be an overlay.

Merge Criteria:

* Support for GPU
* Support for 
    * codegen
    * hasktorch
    * libtorch-ffi
    * libtorch-ffi-helper
    * hasktorch-gradually-typed

PR #709, opened 2024-06-10
Added CLI arguments to MNIST examples

collinarnett:
* Added CLI arguments to MNIST examples
* Added nix derivation for mnist datasets

These changes were added to accommodate the changes being made in #700. Since the nix store is read only and we don't want to ship mnist with hasktorch, it makes more sense to have CLI arguments to the mnist examples.


junjihashimoto:
Thx!

PR #711, opened 2024-06-16
Added nix devshell for cpu and cuda

collinarnett:


collinarnett:
Added [magic-nix-cache](https://github.com/DeterminateSystems/magic-nix-cache) for faster builds.

junjihashimoto:
LGTM!

PR #725, opened 2025-02-15
added all systems and a pytorch hack to allow aarch64 builds

faezs:
instead of depending on libtorch-bin this patch uses python3Packages.pytorch.dev for libtorch. this allows a build for more platforms than the binary mechanism


junjihashimoto:
Thx! But we should link to libtorch since there is the ABI issue.
https://github.com/NixOS/nixpkgs/pull/367998#discussion_r1897515378

PR #726, opened 2025-02-16
Typed form of indexSelect

artemiswkearney:
Wrote this while using hasktorch for a project, and figured I should send a PR to add it.

Hopefully the code style is acceptable; I tried to make the indentation and line spacing match with the other functions in the file, although it didn't feel right to remove the kind signatures so I left those as I'd originally written them in my project.
I'm open to writing tests if needed.
Decided to put it in `Typed.Tensor` rather than `Typed.Functional` because the untyped `indexSelect` is in `Torch.Tensor`, but I could see it in either place.

There's a potential shape-safety issue in that the `Int64`s in `index` could be larger than the relevant dimension of `input`; as far as I'm aware, there isn't really any way to implement `indexSelect` that doesn't have this problem, but maybe there should be a warning in the doc comment about it or something.


junjihashimoto:
Thx!

PR #728, opened 2025-03-28
Force evaluation on all optimizer fields and tensor lists to prevent GPU space leak

kenhkan:
NOTE: This is an incomplete PR. It's opened to showcase a suggestion which may lead to more changes.

I've found a source of space leak in the Adam optimizer (potentially others too), in scenarios where:

1. Optimizer is updated in a tight training loop.
2. GPU memory is limited, relative to the number of epochs and parallel jobs.
3. Tensors (specifically a list of tensors) are used for the optimizer state record.

This space leak is only noticeable in GPU memory, not RAM. I think it's because what's accumulating on the CPU side is only the tensor wrapper, which is tiny. The space leak on the GPU side is only noticeable when I need to run a large number of parallel jobs and/or long training jobs. I've tested with 32 parallels job, each with 50+ epochs on datasets of 100k+ samples. It's more observable when training a lot of small models vs a few large ones.

My reasoning for what's happening: the momentum tensors are created and used for parameter calculation in each update, so the tensors are created on the GPU. The returned values are not strictly evaluated (https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Optim.hs#L140), so the spine is a thunk, preventing the finalizer to be run and decrementing the reference count to the tensor pointer, leading to the momentum tensors leaking on the GPU.

I haven't tested the other optimizers but I would expect the same space leak behavior, but only if the job is large or run for a long time, since the momentum tensors are small.

The change in this PR seems like a hack, so I want to ask: Is there an existing strategy on this issue that I've missed somewhere?

### DIFF
--- a/hasktorch/src/Torch/Optim.hs
+++ b/hasktorch/src/Torch/Optim.hs
@@ -101,11 +101,11 @@ instance Optimizer GDM where
 
 -- | State representation for Adam Optimizer
 data Adam = Adam
-  { beta1 :: Float, -- 1st moment forgetting factor
-    beta2 :: Float, -- 2nd moment forgetting factor
-    m1 :: [Tensor], -- 1st moment
-    m2 :: [Tensor], -- 2nd moment
-    iter :: Int -- iteration
+  { beta1 :: !Float, -- 1st moment forgetting factor
+    beta2 :: !Float, -- 2nd moment forgetting factor
+    m1 :: ![Tensor], -- 1st moment
+    m2 :: ![Tensor], -- 2nd moment
+    iter :: !Int -- iteration
   }
   deriving (Show)
 
@@ -142,6 +142,8 @@ adam lr (Gradients gradients) parameters Adam {..} = (parameters', Adam beta1 be
     -- decaying averages of 1st & 2nd moments
     f1 m1 dp = mulScalar beta1 m1 + mulScalar (1 - beta1) dp
     f2 m2 dp = mulScalar beta2 m2 + mulScalar (1 - beta2) (dp * dp)
+    !_ = forceInPlace m1'
+    !_ = forceInPlace m2'
     m1' = zipWith f1 m1 gradients
     m2' = zipWith f2 m2 gradients
     -- bias adjustment
@@ -155,6 +157,10 @@ adam lr (Gradients gradients) parameters Adam {..} = (parameters', Adam beta1 be
 
 instance Optimizer Adam where
   step = adam
+
+forceInPlace :: [a] -> [a]
+forceInPlace [] = ()
+forceInPlace (x : xs) = seq x (forceInPlace xs)
 
 --
 -- Adagrad


kenhkan:
I forgot to add: An option that I'm thinking is to make the record an instance of NFData in Control.DeepSeq so that we could call `force` on it. What do you think about that approach?

junjihashimoto:
@kenhkan 
Thank you!
Basically, the fix is good.
Since we are using the StrictData extension, we don't need the `!`, it's already applied.
I think we can use NFData to deepseq with the step function of other optimizers, but you can also use other PRs.

kenhkan:
Hi @junjihashimoto ! Thanks for the comment. I was busy with work but now have time to bring this to the finish line.

> Since we are using the StrictData extension, we don't need the !, it's already applied.

Ah! Thanks for the note. I'll remove that part.

> I think we can use NFData to deepseq with the step function of other optimizers,

I just realized that `Tensor` is in the record and it's a [ForeignPtr](https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Tensor.hs#L58). So I don't think the NFData approach works. Do you think there is another way to deal with just forcing the spine other than the initially proposed appraoch?

> but you can also use other PRs.

What do you mean by other PRs?

junjihashimoto:
I'd like to merge this pr, but the optimizer tests fail.

kenhkan:
Thanks for the guidance! I made the change. I'm just waiting for a job that I started to test it out. I'll run it for at least a few hours to see if there is any space leak over a longer span of time.

And yes, I will get the tests to pass after I've confirmed that it's working with the test job and after I update the other optimizers as well.

I'll circle back here tomorrow.

junjihashimoto:
@kenhkan 
Sorry, I forgot to implement NFData. Could you please remove the implementation to pass the CI?
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/bench/Alloc.hs#L48-L54
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/bench/Runtime.hs#L52-L58

kenhkan:
@junjihashimoto 

By "implementation" you mean remove the two highlighted files you have in `bench/`, correct?

Would https://github.com/hasktorch/hasktorch/blob/819e1dcc744c5c9755aa7cf3f1d15bed6e61af4f/hasktorch/src/Torch/Tensor.hs be the right place to put `instance NFData (ForeignPtr a)`?

Also, I've been running a job for a while now and no space leak is observed with what's currently in the PR! I'll update the other optimizers as well once the two assumptions above are confirmed.

Third question: I noticed that there is a typed version of all the optimizers, like [Adam](https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/Optim.hs#L240), that use `HList` instead of a Haskell list. I'm not familiar with `HList` and don't have a working program to test the typed version. I would assume that it has the same issue as it looks like HList uses regular lists under the hood on first glance. Can you advise on how best to patch the typed version with `HList`?

junjihashimoto:
> By "implementation" you mean remove the two highlighted files you have in bench/, correct?

Yes.

> Would https://github.com/hasktorch/hasktorch/blob/819e1dcc744c5c9755aa7cf3f1d15bed6e61af4f/hasktorch/src/Torch/Tensor.hs be the right place to put instance NFData (ForeignPtr a)?

For now, we may not need this.

> 
> Also, I've been running a job for a while now and no space leak is observed with what's currently in the PR! I'll update the other optimizers as well once the two assumptions above are confirmed.
> 

Thx!

> Third question: I noticed that there is a typed version of all the optimizers, like [Adam](https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/Optim.hs#L240), that use HList instead of a Haskell list. I'm not familiar with HList and don't have a working program to test the typed version. I would assume that it has the same issue as it looks like HList uses regular lists under the hood on first glance. Can you advise on how best to patch the typed version with HList?

HList gradient can become untyped tensors with `ATen.Castable (HList gradients) [D.ATenTensor]`.
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/Optim.hs#L66

We should pass the `tensors `to the `force` function.
https://github.com/hasktorch/hasktorch/blob/master/hasktorch/src/Torch/Typed/Optim.hs#L80

PR #729, opened 2025-03-28
Match finalizer file name

kenhkan:
Fix mismatch of FFI calls: Calls to finalizers currently do not reach the C side ([example](https://github.com/hasktorch/hasktorch/blob/master/libtorch-ffi/src/Torch/Internal/Objects.hs#L26)) and are quietly failing, leading to space leak.


junjihashimoto:
@kenhkan 
Thx!

PR #742, opened 2025-08-07
chore: Update hasktorch to setup the same torch library as libtorch

mchav:
The current configuration will have version failures. We also set the environment to the newly installed hasktorch lib.

Also some minor fixes to yaml file:
* Prefer apt-get (or scripts) vs apt (for human interactive use)
* apt-update && apt-upgrade so dependencies can build.


junjihashimoto:
@mchav Thx!

PR #744, opened 2025-08-07
Fix/remove setenv

junjihashimoto:

PR #745, opened 2025-08-07
fix: Add libtokenizers to setup.

mchav:
Also fixes nix environment variable setup.


mchav:
@junjihashimoto  Didn't work. You're right, we have to configure gradually typed separately.

mchav:
Most of the library builds. The two issues to be addressed are in #744. We can follow up with the nix build but the cabal and stack builds should be fine now.

mchav:
@junjihashimoto if you do merge this please squash it so it doesn't pollute the commit history. Thank you.

PR #752, opened 2025-08-12
Fixes for Nix setup

collinarnett:
This PR makes changes based on #744 and allows users to build the main libraries hasktorch, libtorch-ffi, codegen and examples using both Nix and cabal. It also adds `cabal-install` and `haskell-language-server` to the devshell for easier development.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -5,7 +5,7 @@ packages:
     hasktorch/*.cabal
     examples/*.cabal
     experimental/untyped-nlp/untyped-nlp.cabal
-    experimental/dataloader-cifar10/dataloader-cifar10.cabal
+--  experimental/dataloader-cifar10/dataloader-cifar10.cabal
     experimental/bounding-box/bounding-box.cabal
 
 -- source-repository-package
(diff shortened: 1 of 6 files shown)


sidkshatriya:
On second thoughts I'm wondering if switching to ghc 9.8.4 is worth it (even though I had kinda suggested it). It might be a good idea to just stick with 9.6.7 both for cabal and nix for now ? 

Much as nix is always very cool, I have recently abandoned nix as a vehicle for my hasktorch specific explorations. The libtorch-ffi automatic download linked with the environment LIBTORCH_VERSION / LIBTORCH_CUDA_VERSION is convenient and download/build times much better than for nix.

collinarnett:
One of the reasons we're updating to GHC 9.8 is because it's the latest version of GHC tracked (until today) by Stackage LTS which is also what Nixpkgs follows. 

Also as far as I know, no one is using the gradually typed package.

> The libtorch-ffi automatic download linked with the environment LIBTORCH_VERSION / LIBTORCH_CUDA_VERSION is convenient and download/build times much better than for nix.

There's still room for improvement in terms of speeding this up. Two things immediately come to mind. One is that I generally have a updated cache locally when it comes to building all the packages so I don't know what the end-user experience is in regards to how good the cachix caching is. So that could be improved. The second idea to improve build times would be enabling [incremental builds](https://nixos.org/manual/nixpkgs/unstable/#haskell-incremental-builds) which should theoretically give an experience similar to `cabal build`

PR #762, opened 2025-11-07
documentation: Callout CUDA versionining in README (fixes #761)

### LINKED ISSUE #761: cannot move tensor from "CPU:0" to "CUDA:0"
I'm following the instruction [linux+cabal+cuda11](https://github.com/hasktorch/hasktorch#linuxcabalcuda11) to install Hasktorch and run the MNIST CNN example,  but got the following:
```
static-mnist-cnn: cannot move tensor from "CPU:0" to "CUDA:0"
CallStack (from HasCallStack):
  error, called at src/Torch/Tensor.hs:301:7 in hasktorch-0.2.1.7-inplace:Torch.Tensor
```

I'm sure that CUDA drivers are installed correctly, as Pytorch works well with cudas on my server (Ubuntu 22.04 LTS).

mchav:


junjihashimoto:
Thx!

PR #780, opened 2026-03-01
Add Adam* convenience constructors

kenhkan:
@junjihashimoto Continuing the conversation on your suggestion in https://github.com/hasktorch/hasktorch/pull/777 , is this what you meant? It looks like there isn't one for the standard Adam so I added that as well.

### DIFF
--- a/hasktorch/src/Torch/Optim.hs
+++ b/hasktorch/src/Torch/Optim.hs
@@ -114,6 +114,10 @@ data Adam = Adam
 
 instance NFData Adam
 
+-- | Convenience constructor with standard defaults
+mkDefaultAdam :: [Parameter] -> AdamW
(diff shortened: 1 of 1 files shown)


junjihashimoto:
```
src/Torch/Optim.hs:119:17: error: [GHC-83865]
    • Couldn't match expected type ‘[Parameter] -> AdamW’
                  with actual type ‘Adam’
    • Possible cause: ‘mkAdam’ is applied to too many arguments
      In the expression: mkAdam 0 0.9 0.999 0.01
      In an equation for ‘mkDefaultAdam’:
          mkDefaultAdam = mkAdam 0 0.9 0.999 0.01
    |
119 | mkDefaultAdam = mkAdam 0 0.9 0.999 0.01
    |                 ^^^^^^^^^^^^^^^^^^^^^^^
```
There's a compilation error.

PR #781, opened 2026-03-01
Expose parameter group interface

kenhkan:
I need a way to access parameter group as described in [torch.optim](https://docs.pytorch.org/docs/stable/optim.html) to implement per-parameter weight decay in AdamW. I couldn't find an existing way to do that that is in line with the PyTorch convention. This is a draft to showcase an implementation that works in my case. Is there a better way to do this with Hasktorch's conventions?


junjihashimoto (approved):

junjihashimoto:
@kenhkan  Sorry for the late reply. I don't know of a better way, but I'd like you to try it out and I'll immediately incorporate any feedback you give me.

PR #782, opened 2026-04-28
fix: Remove partial signatures from Convolution.

mchav:
Partial signatures produce compile time errors where we cannot satisfy constraints. Even though it's signficantly more verbose we should specify them completely.


junjihashimoto:
Thx!

PR #784, opened 2026-06-01
Add option to set LR per parameter group

benjamin-rdm:
This change allows for setting the LR for a specific parameter group within the optimizer. Feedback appreciated.


junjihashimoto:
Thank you. It would be good to have a test.

PR #795, opened 2026-07-28
Add value-threaded PureGenerator and multinomial sampling

collinarnett:
Torch.Random's Generator mutates the underlying ATen generator in place, so two draws from the same Generator value disagree and a sequence is not reproducible from its seed alone.

PureGenerator threads the RNG as an immutable value instead: withPureGenerator clones the ATen generator before each draw and advances only the clone, so the input is never mutated and a given PureGenerator value always produces the same output. Cloning also preserves the stream position, so the returned generator continues the same stream rather than starting a reseeded one. Adds multinomial on top of it.

Torch.Typed.Random is the typed counterpart, following the one typed module per untyped module layout the rest of Torch.Typed uses. It tags PureGenerator with a device, requires floating point probabilities, and derives the output shape with MultinomialShape, which replaces the trailing dimension with the sample count.

Extracts newGeneratorPtr for the device dispatch that mkGenerator and generatorFactory each spelled out separately, so the two generator types construct their ATen handles the same way.

### DIFF
--- a/hasktorch/hasktorch.cabal
+++ b/hasktorch/hasktorch.cabal
@@ -61,6 +61,7 @@ library
                     , Torch.Typed.Auxiliary
                     , Torch.Typed.Factories
                     , Torch.Typed.Functional
+                    , Torch.Typed.Random
                     , Torch.Typed.NN
                     , Torch.Typed.NN.Convolution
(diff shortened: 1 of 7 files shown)


junjihashimoto:
Thx!

